PR #141, opened 2011-06-23
Return non-zero exit codes when there is an application failure

hello-josh:
docblox always exits 0 indicating success. This fix enables non-zero exit codes on failure so that you can properly check for failure in ant and other automated build tools


neoascetic:
Broken in development?

jaapio:
Development branch contains a very early version of v3. You cannot use this version. Or do you mean anything else?

PR #645, opened 2012-10-15
Added descriptions and answered TODOs

mvriel:
@ashnazg I have appended a lot more information onto the PSR that was present in the docs of phpDocumentor2 but not yet ported. Can you take a look at this?

Regarding the @link question: I believe that the representation part is up to the implementing body (parser/generator) and is beyond the scope of this specification.

I think we should discuss the @global usage scenario and I think we should have a chapter on option arrays to clarify how Inline PHPDoc's should be used.

### DIFF
--- a/docs/PSR.md
+++ b/docs/PSR.md
@@ -27,6 +27,7 @@ Obsoletes:
       5.3. Tags
         5.3.1. Tag Name
         5.3.2. Tag Signature
+        5.3.3. Inline PHPDoc
       5.4. Examples
     6. Inheritance
       6.1. Class Or Interface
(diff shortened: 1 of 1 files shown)


docs/PSR.md
```diff
@@ -1044,10 +1093,105 @@ This tag MUST NOT occur more than once in a "DocBlock".
...
+The following example demonstrates the use of an "Inline PHPDoc" to document
+an option array with 2 elements: 'required' and 'label'.
+
+```php
+/**
+ * Initializes this class with the given options.
+ *
+ * @param string[] $options {
+ *     @type boolean $required Whether this element is required
+ *     @type string  $label    The display name for this element
```
ephrin: What is the difference between `$required` and `$label`? How to distinguish mandatory one?

Issue #157, opened 2011-07-17
Add support for traits


mvriel:
Implement support for traits as specified in: https://wiki.php.net/rfc/horizontalreuse

PR #1245, opened 2014-05-09
#157: Classes do not inherit trait methods and props

mvriel:
Methods and properties from traits are currently not shown in the method
and property listing for a class. In this commit we treat traits equal to
parent classes and inherit their methods.


ve3:
Is this already in v3.0.0-alpha.3? because I cannot see methods from trait in the class document.

jaapio:
Could you please open a new issue for that. With some example code attached, so we can investigate the  issue.

PR #1500, opened 2014-12-16
[WIP] Provide git location to generate docs from

mvriel:
More description follows


mvriel:
Cancel these changes as a different route is planned

PR #1510, opened 2015-01-14
Fix faulty "missing argument in phpDoc block" check

sascha-egerer:
Fixes phpDocumentor/phpDocumentor2#1476


mvriel:
Please resubmit this PR against master so that we can include it in version 2.8.*

PR #1511, opened 2015-01-14
Fix faulty "missing argument in phpDoc block" check

sascha-egerer:
Fixes phpDocumentor/phpDocumentor2#1476


mvriel:
Thank you for helping us out! I am going to merge this PR (the build fails are not due to this PR) and it will be included in the next 2.8.* release

PR #1537, opened 2015-03-27
Fill in missing terms in the glossary

cincodenada:
I'm not super familiar with phpDoc, so the text is mostly taken from other places in the docs and edited a bit.  It's probably not the best summaries, but I think they're an improvement over the "TODO" block, anyway, especially when these terms are linked from the "Getting Started" page (which is where I found them)

I've generated the docs with Sphinx and fixed them up so everything looks like I was intending.

### DIFF
--- a/docs/glossary.rst
+++ b/docs/glossary.rst
@@ -166,29 +166,52 @@ Glossary
 
     Template
     Templates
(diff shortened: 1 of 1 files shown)


mvriel:
Thank you! This looks like an excellent addition

PR #1547, opened 2015-05-08
replace ProjectDescriptor with ProjectInterface

jaapio:
ProjectDescriptor is final it can be mocked so replace it with ProjectInterface. This is causing a number of test failures in PhpDocumentor2.

Warning: this is a BC break!
This PR depends on https://github.com/phpDocumentor/Reflection/pull/36


mvriel:
Thanks!

PR #1561, opened 2015-05-23
fixed 2 unittests

rgeraads:
- [x] Removed final from Parser class to make test pass
- [x] removed newline from VERSION file to make another test pass
- [x] removed pdf ServiceProvider from test since it no longer exists


mvriel:
:+1:

PR #1563, opened 2015-05-23
Introduce PHP-DI

### LINKED ISSUE #1559: Replace Cilex with PHP-DI 5
Currently the Application class works as a Dependency Injection Container based on Pimple; what we want is to replace Pimple with PHP-DI 5 as the latter offers auto-wiring and makes service definitions a lot smaller.

It is expected that this change will also reduce the number of dependencies and thus make for a smaller and simpler application.

- [x] Do not extend Application class from Cilex
- [x] Add PHP-DI 5 (currently on master) to the composer.json
- [x] Add Symfony Console to the composer.json (as it is no longer imported with Cilex)
- [x] Remove Cilex from the composer.json
- [x] Create a Service Definition file for PHP-DI (http://php-di.org/news/14-php-di-5-syntaxic-sugar.html)
- [x] Rewrite all Service Providers into that Service Definition (not all services need to be copied since PHP-DI uses autowiring)
- [x] Initialize PHP-DI from the Bootstrap and inject it into the Application class

mvriel:
> This will resolve #1559 

In this PR I have changed the way phpDocumentor works with dependency injection. Previously it made use of Cilex and Pimple to handle its needs but in an effort to simplify dependency management and reduce the number of dependencies I have replace Pimple with PHP-DI 5 RC1.

> *Important*: PHP-DI 5 is about to hit stable next week or the week thereafter and I have requested to make the Doctrine/Annotation and Doctrine/Cache dependency optional because we don't need it. Once PHP-DI 5 stable hits the shelves we'll need to update the composer.json again.

In addition to this refactoring I have removed the following components and their dependencies:

- Scrybe
- Logging

And in addition to this a minor refactoring in Plugin handling has been done because the plugins were dependent on Pimple. _This also means that phpDocumentor 2 plugins no longer work in phpDocumentor 3_.


composer.json
```diff
@@ -6,32 +6,27 @@
     "license": "MIT",
     "autoload": {
         "psr-0":{
-            "phpDocumentor": ["src/", "tests/unit/"],
-            "Cilex\\Provider": ["src/"]
+            "phpDocumentor": ["src/", "tests/unit/"]
```
jaapio: I would split this in autoload and autoload-dev

composer.json
```diff
@@ -6,32 +6,27 @@
...
-        "phpdocumentor/fileset":             "~1.0",
-        "phpdocumentor/reflection":          "dev-develop",
+        "phpdocumentor/graphviz":   "~1.0",
+        "phpdocumentor/fileset":    "~1.0",
+        "phpdocumentor/reflection": "dev-develop",
 
-        "zendframework/zend-i18n": "~2.3",
-        "zendframework/zend-servicemanager": "~2.3"
+        "zendframework/zend-i18n": "~2.4",
+        "zendframework/zend-servicemanager": "^2.4"
```
jaapio: Why are you using ^ and ~ in some cases, I would prefere to use one of them.
Beside that if we could use the lowest possible version of a package we are more flexible during and installation for the end user. 
For example I still have project that depend on zf2 ~2.3.8  those can't use phpdocumentor3 in there composer.json

src/phpDocumentor/Application.php
```diff
@@ -45,169 +33,51 @@ class Application extends Cilex
...
-            while ($monolog->popHandler()) {
-            }
-        } catch (\LogicException $e) {
-            // popHandler throws an exception when you try to pop the empty stack; to us this is not an
-            // error but an indication that the handler stack is empty.
-        }
-
-        if ($level === 'quiet') {
-            $monolog->pushHandler(new NullHandler());
+        gc_disable();
```
jaapio: was GC disabled before? And why would you disable it?

src/phpDocumentor/Command/Command.php
```diff
@@ -21,21 +21,9 @@
  * Includes additional methods to forward the output to the logging events
  * of phpDocumentor.
  */
-class Command extends \Cilex\Command\Command
+class Command extends \Symfony\Component\Console\Command\Command
```
jaapio: Please add a use statement with an alias

src/phpDocumentor/Command/Helper/LoggerHelper.php
```diff
@@ -30,19 +29,16 @@
...
-     * adds a new option `log`.
-     *
-     * @param Command $command
-     * @return void
-     */
-    public function addOptions($command)
+    /** @var Translator */
+    private $translator;
+
+    public function __construct(Dispatcher $dispatcher, Translator $translator)
```
jaapio: missing a docblock here?

src/phpDocumentor/ContainerDefinitions.php
```diff
@@ -0,0 +1,274 @@
...
+return [
+    // -- Parameters
+    'application.version' => function () {
+        return strpos('@package_version@', '@') === 0
+            ? trim(file_get_contents(__DIR__ . '/../../VERSION'))
+            : '@package_version@';
+    },
+    'cache.directory' => sys_get_temp_dir(),
+    'linker.substitutions' => [
+        'phpDocumentor\Descriptor\ProjectDescriptor' => ['files'],
```
jaapio: is is possible to use ::class here?

src/phpDocumentor/Parser/Command/Project/ParseCommand.php
```diff
@@ -47,17 +48,32 @@
     private $exampleFinder;
 
     /**
+     * Evil!
```
jaapio: please add @todo or @hack so we are able to find this later ;-)

src/phpDocumentor/Plugin/Core/ServiceProvider.php
```diff
@@ -24,102 +25,91 @@
  * phpDocumentor and, for backwards compatibility, registers the service providers for Graphs, Twig and PDF to
  * the container.
  */
-final class ServiceProvider implements ServiceProviderInterface
+final class ServiceProvider
```
jaapio: I hope is one is on the refactor wish list? Way to much properties in here,  and I don't understand what this class is doing.

src/phpDocumentor/Plugin/Graphs/ServiceProvider.php
```diff
@@ -11,22 +11,28 @@
 
 namespace phpDocumentor\Plugin\Graphs;
 
-use Cilex\Application;
-use Cilex\ServiceProviderInterface;
 use phpDocumentor\Plugin\Graphs\Writer\Graph;
 use phpDocumentor\Transformer\Writer\Collection;
 
-class ServiceProvider implements ServiceProviderInterface
+class ServiceProvider
```
jaapio: should be final.
And it looks like this is an addition on the core/serviceprovider? There must be a beter name for those classes, since they are configuring the Writer\Collection?

composer.json
mvriel: I didn't know about autoload-dev, so I will fix that

composer.json
mvriel: The ^ was added by the `composer require` command; I will fix this and indeed, a downgrade to `~2.3` is probably better

src/phpDocumentor/Application.php
mvriel: gc was disabled before; because phpDocumentor creates huge object trees while parsing we have noticed that the Garbage Collector slows down execution with larger projects while it does not offer significant benefits in terms of memory.

As such we disable gc to improve performance

src/phpDocumentor/ContainerDefinitions.php
mvriel: Yes, will fix

src/phpDocumentor/Parser/Command/Project/ParseCommand.php
mvriel: Good point

mvriel:
@jaapio Regarding the remaining service providers and plugins: the way we deal with plugins is on the refactor list as it needs to be changed (current implementation can no longer be sustained); I need to do some research on how to approach this issue because the list of plugins is determined on runtime and this can cause some conflict issues when dealing with the container.

This will be addressed in a separate PR; as such the Service Providers are maintained in the Plugins but these will change

composer.json
mvriel: I have just reviewed the composer.json and changed some of the versions; I have defaulted on the ^ operator because that will allow us to specify a lower minor (such as symfony/console ^2.3) but lock on the latest minor (symfony/console 2.6 IIRC)

mvriel:
@jaapio I have resolved all your comments except for those plugin related because that is stuff for a future PR :)

jaapio:
This one can be merged ;-)

mvriel:
And done!

PR #1570, opened 2015-05-26
Remove translation support

### LINKED ISSUE #1567: Remove translator
In an effort to simplify phpDocumentor's architecture and remove redundant features we have chosen to remove the translation support from phpDocumentor. This was used so little (if at all) that it serves no other purpose than to bloat the app.

mvriel:
> please merge #1563 first because that contains refactorings that were necessary for this issue

> this issue resolves #1567 

With phpDocumentor we are working on simplifying the current flow and removing bloat that we do not deem necessary for the app to function and which is not requested (too often). One of these things is translation support; this was intended to simplify the process of filtering messages on type but it costs too much code and adds two dependencies


mvriel:
@jaapio (and others) can you review this item after #1563 is merged?

src/phpDocumentor/Plugin/Core/Descriptor/Validator/DeprecatedValidator.php
```diff
@@ -75,7 +75,7 @@
 //            if (count($this->docblock->getTagsByName($tag)) < 1) {
 //                continue;
 //            }
-//            $this->logParserError('CRITICAL', 'PPC:ERR-50006', $this->lineNumber, array($tag, $this->entityName));
+//            $this->logParserError('CRITICAL', 'Found deprecated tag "%s" in %s', $this->lineNumber, array($tag, $this->entityName));
 //        }
 //    }
 //}
```
jaapio: can't we just remove this file?

mvriel: It's there because it needs to be restored; it went broken after a refactoring

jaapio:
can be merged, I would prefer to remove deprecated files in this major clean up

mvriel:
Agree wholeheartedly with that last comment!

PR #1575, opened 2015-05-27
fixed box.json with correct packages

rgeraads:


composer.json
```diff
@@ -24,11 +24,11 @@
 
         "phpdocumentor/graphviz":   "^1.0",
         "phpdocumentor/fileset":    "^1.0",
-        "phpdocumentor/reflection": "dev-develop"
+        "phpdocumentor/reflection": "dev-develop",
+        "herrera-io/phar-update": "^2.0"
```
mvriel: Why have you moved the phar-update package to the require section; is it necessary for a composer installation? If it is only necessary to build the phar I prefer it remains in the dev version

mvriel:
:+1:

PR #1578, opened 2015-05-28
Add Template and Action classes and tests

mvriel:


src/phpDocumentor/Renderer/ActionHandlerLocator.php
```diff
@@ -0,0 +1,48 @@
...
+
+use Interop\Container\ContainerInterface;
+
+/**
+ * Service that attempts to locate an ActionHandler object based on the class name of a given Action.
+ */
+class ActionHandlerLocator
+{
+    /**
+     * @var ContainerInterface
```
jaapio: missing a summary here

tests/unit/phpDocumentor/Renderer/ActionHandlerLocatorTest.php
```diff
@@ -0,0 +1,34 @@
...
+use Mockery as m;
+
+/**
+ * Tests the functionality for the ActionHandlerLocator class.
+ */
+class ActionHandlerLocatorTest extends \PHPUnit_Framework_TestCase
+{
+    public function testIfLocatorRetrievesAnActionWithHanderAppendedToClassName()
+    {
+        $action = m::mock('phpDocumentor\Renderer\Action');
```
jaapio: I would use the ::class constant here, Helps when you refactor the name of a class

src/phpDocumentor/Renderer/Template.php
```diff
@@ -0,0 +1,120 @@
...
+ * the Template also determines at which location to create pre-rendered output and has to copy
+ * assets such as CSS files, Javascript and images.
+ *
+ * In order to do this you can define a series of Actions with a template that prescribe which
+ * Actions to execute in the specified order.
+ *
+ * With a template you can provide a series of parameters that can be passed to every Action
+ * so that a default piece of information is provided, such as the target location.
+ */
+class Template
```
jaapio: shouldn't this be final?

mvriel:
@jaapio thanks for the review; I have fixed your feedback and added the missing test for the Template class (had forgotten to commit that in the initial commit *facepalm*)

jaapio:
looks good, Wercker should be able to pass? think you need a rebase here. But can be merged without that.

PR #1580, opened 2015-05-29
[wip] Update template tests with covers and add TemplateFactory

mvriel:


src/phpDocumentor/Renderer/TemplateFactory.php
```diff
@@ -0,0 +1,163 @@
...
+
+namespace phpDocumentor\Renderer;
+
+use Webmozart\Assert\Assert;
+use phpDocumentor\Renderer\Template\Parameter;
+
+/**
+ * Creates a new Template based on an array with a Template definition.
+ */
+class TemplateFactory
```
jaapio: Would like to see this as Final

tests/unit/phpDocumentor/Renderer/TemplateFactoryTest.php
```diff
@@ -0,0 +1,238 @@
...
+ *
+ * @copyright 2010-2015 Mike van Riel<mike@phpdoc.org>
+ * @license   http://www.opensource.org/licenses/mit-license.php MIT
+ * @link      http://phpdoc.org
+ */
+
+namespace phpDocumentor\Renderer {
+
+    use Mockery as m;
+    use phpDocumentor\Renderer\Action\Action1;
```
jaapio: Must have missed this in an earlier review, But action1 and action2 sounds not as a very descriptive name

mvriel: These are test classes needed for the tests; would you rather see something more descriptive than Action1 through Action5

jaapio: If you make clear that those classes are dummies or test assets it is fine with me. Maybe add a testassets namespace?

src/phpDocumentor/Renderer/TemplateFactory.php
```diff
@@ -0,0 +1,163 @@
...
+     * @throws \InvalidArgumentException if the class deduced from the 'name' field does not implement the Action
+     *     interface
+     * @throws \RuntimeException if the factory method of the deduced Action does not return an Action object.
+     *
+     * @return void
+     */
+    private function addActionToTemplate(Template $template, $action)
+    {
+        $actionClass = $action['name'];
+        if (strpos($actionClass, '\\') === false) {
```
jaapio: Missing a test case for this block.

mvriel: I am going to fix this in a followup PR so that we can add these changes now

tests/unit/phpDocumentor/Renderer/TemplateFactoryTest.php
mvriel: The namespace is fixed because we are testing the version of Actions that are bound to phpDocumentor; I agree with you that the names could then be more descriptive

src/phpDocumentor/Renderer/TemplateFactory.php
jaapio: :+1:

mvriel: Ok, it was a small fix so added that in the latest commit :)

PR #1582, opened 2015-05-30
Update composer file to allow php7

mvriel:
The travis tests for php7 fail because composer was configured not to allow it
and we want travis to work as an early warning system.


jaapio:
looks, good to me, can be merge :+1:

PR #1583, opened 2015-05-30
Feature/guides

jaapio:
Last part of document foundation


src/phpDocumentor/Guides/Document/Html.php
```diff
@@ -0,0 +1,50 @@
...
+        parent::__construct($path, $title, $content);
+        $this->contentType = new ContentType('text/html');
+    }
+
+    /**
+     * Returns the type of the document.
+     *
+     * @return ContentType
+     */
+    function getContentType()
```
mvriel: I am missing a visibility specifier here

tests/unit/phpDocumentor/Guides/ContentTypeTest.php
```diff
@@ -0,0 +1,30 @@
...
+ */
+class ContentTypeTest extends \PHPUnit_Framework_TestCase
+{
+    /**
+     * @covers ::__construct
+     * @covers ::__toString
+     */
+    public function testToString()
+    {
+        $contenType = new ContentType('text/html');
```
mvriel: Misspelling in the variable; should be `$contentType`

tests/unit/phpDocumentor/Guides/Document/HtmlTest.php
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+namespace phpDocumentor\Guides\Document;
+
+use phpDocumentor\Guides\ContentType;
+use phpDocumentor\Path;
+
+class HtmlTest extends \PHPUnit_Framework_TestCase
+{
+    public function testGetContentType()
```
mvriel: In the other tests you use `@covers` annotations but not in this; is this on purpose?

tests/unit/phpDocumentor/Guides/DocumentTest.php
```diff
@@ -0,0 +1,35 @@
...
+{
+    /**
+     * @covers ::__construct
+     * @covers ::getTitle
+     * @covers ::getPath
+     * @covers ::getContent
+     */
+    public function testGetters()
+    {
+        $document = new DummyDocument(new Path('my/Path'), 'SomeTitle', 'MyContent');
```
mvriel: You can also create a partial mock using mockery where you only mock the getContentType method; that will enable you to omit the DummyDocument. I don't mind either approach so if you prefer this then I am fine with it, I didn't know if you knew about the partial mock technique for testing abstract classes

mvriel:
Nothing exciting found in the code review; just a few minor style things

tests/unit/phpDocumentor/Guides/DocumentTest.php
jaapio: I know the partial mocking technique. But I think it is slower than just creating a dummy / stub. And since I'm running the test suite very often. (Partial and Full) I would like to keep things fast.

jaapio:
All done. Ready for merge

tests/unit/phpDocumentor/Guides/DocumentTest.php
mvriel: :+1:

mvriel:
Don't change a thing! Perfect :)

PR #1585, opened 2015-05-31
Add Command Bus and a Locator to interact with the DIC

mvriel:


src/phpDocumentor/Application/Commands/ContainerLocator.php
```diff
@@ -0,0 +1,65 @@
...
+ */
+
+namespace phpDocumentor\Application\Commands;
+
+use Interop\Container\ContainerInterface;
+use League\Tactician\Handler\Locator\HandlerLocator;
+use Webmozart\Assert\Assert;
+
+/**
+ * Command Handler Locator for Tactician that can locate handlers based on a provided mapping or by suffixing the FQCN
```
jaapio: These comment lines are to long. Have to scroll on github

src/phpDocumentor/Application/Commands/ContainerLocator.php
```diff
@@ -0,0 +1,65 @@
...
+
+    /**
+     * Retrieves the handler for a specified command class.
+     *
+     * @param string $commandName
+     *
+     * @throws \InvalidArgumentException if the provided command name does not match to an existing class.
+     * @throws \InvalidArgumentException if the name of the command handler does not match to an existing class.
+     *
+     * @return object
```
jaapio: Handlers do not have to implement an interface?

jaapio:
Nothing important to fix before merging.

src/phpDocumentor/Application/Commands/ContainerLocator.php
mvriel: Nope, SimpleBus requires that (or actually, until the recent version it did) but Tactician doesn't. This is a good thing imo as it decouples the Commands from the Library

src/phpDocumentor/Application/Commands/ContainerLocator.php
mvriel: 120 characters is allowed according to PSR-2, but perhaps we should limit DocBlocks to 80 for readability?

PR #1587, opened 2015-06-03
Value object for Dsn

mbed67:
Splits a DSN string into its parts

### DIFF
--- a/src/phpDocumentor/Dsn.php
+++ b/src/phpDocumentor/Dsn.php
@@ -18,7 +18,6 @@ namespace phpDocumentor;
 
 final class Dsn
 {
-
     /** @var string */
     private $scheme;
 
@@ -42,6 +41,12 @@ final class Dsn
 
     /** @var array  */
     private $parameters = [];
+
+    /** @var array  */
+    private $locationParts = [];
+
+    /** @var array  */
+    private $dsnParts = [];
 
     /**
      * @param $dsn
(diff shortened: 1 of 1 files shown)


src/phpDocumentor/Dsn.php
```diff
@@ -0,0 +1,308 @@
...
+ * @license   http://www.opensource.org/licenses/mit-license.php MIT
+ * @link      http://phpdoc.org
+ */
+
+namespace phpDocumentor;
+
+/**
+ * Value Object for DSN.
+ */
+
```
mvriel: Superfluous whiteline

src/phpDocumentor/Dsn.php
```diff
@@ -0,0 +1,308 @@
...
+    /** @var string */
+    private $user;
+
+    /** @var string */
+    private $password;
+
+    /** @var string */
+    private $path;
+
+    /** @var array  */
```
mvriel: instead of `array` it is generally recommended to use a more specific type, in this case that would be `string[]` (array of strings)

src/phpDocumentor/Dsn.php
```diff
@@ -0,0 +1,308 @@
...
+    private $query = [];
+
+    /** @var array  */
+    private $parameters = [];
```
mvriel: instead of `array` it is generally recommended to use a more specific type, in this case that would be `string[]` (array of strings)

src/phpDocumentor/Dsn.php
```diff
@@ -0,0 +1,308 @@
...
+
+    /** @var array  */
+    private $locationParts = [];
+
+    /** @var array  */
+    private $dsnParts = [];
+
+    /**
+     * @param $dsn
```
mvriel: Missing a summary and type for the dsn

src/phpDocumentor/Dsn.php
```diff
@@ -0,0 +1,308 @@
...
+     */
+    public function __construct($dsn)
+    {
+        $this->parse($dsn);
+    }
+
+    /**
+     * @return string
```
mvriel: missing a summary

src/phpDocumentor/Dsn.php
```diff
@@ -0,0 +1,308 @@
...
+     */
+    public function getParameters()
+    {
+        return $this->parameters;
+    }
+
+    /**
+     * @param $dsn
```
mvriel: missing a summary and type for the DSN

src/phpDocumentor/Dsn.php
```diff
@@ -0,0 +1,308 @@
...
+     */
+    private function parse($dsn)
+    {
+        $this->dsnParts = explode(';', $dsn);
```
mvriel: I recommend using `is_string` to test first if the received parameter is really a string and if it isn't to throw an InvalidArgumentException

src/phpDocumentor/Dsn.php
```diff
@@ -0,0 +1,308 @@
...
+        $location = $this->dsnParts[0];
+        unset($this->dsnParts[0]);
+        $this->locationParts = parse_url($location);
```
mvriel: Instead of using a property I can advice passing the locationParts variable to the `parse*` methods below; that way the input is explicit and because the locationParts are only used within this one public process (it is not used in other public methods) it is commonly a good practice to keep the variable inside the function scope and not push it into class scope

mvriel: The same advice also applies for the dsnParts

src/phpDocumentor/Dsn.php
```diff
@@ -0,0 +1,308 @@
...
+        if (array_key_exists('scheme', $this->locationParts)) {
+            $validSchemes = ['file', 'git+http', 'git+https'];
+            if (! in_array(strtolower($this->locationParts['scheme']), $validSchemes)) {
+                throw new \InvalidArgumentException(
+                    sprintf('"%s" is not a valid scheme.', $this->locationParts['scheme'])
+                );
+            }
+            $this->scheme = $this->locationParts['scheme'];
+
+        } else {
```
mvriel: You can remove the `else` (and hence reduce complexity) by starting your method with:

```
if (! array_key_exists('scheme', $this->locationParts)) {
    $this->scheme = 'file';
    return;
}
```

src/phpDocumentor/Dsn.php
```diff
@@ -0,0 +1,308 @@
...
+    private function parseScheme()
+    {
+        if (array_key_exists('scheme', $this->locationParts)) {
+            $validSchemes = ['file', 'git+http', 'git+https'];
+            if (! in_array(strtolower($this->locationParts['scheme']), $validSchemes)) {
+                throw new \InvalidArgumentException(
+                    sprintf('"%s" is not a valid scheme.', $this->locationParts['scheme'])
+                );
+            }
+            $this->scheme = $this->locationParts['scheme'];
```
mvriel: please apply the strtolower here as well, that will normalize the scheme so that consumers of this class won't have to

src/phpDocumentor/Dsn.php
```diff
@@ -0,0 +1,308 @@
...
+
+        $this->parseParameters();
+    }
+
+    /**
+     * validates and sets the scheme property
+     */
+    private function parseScheme()
+    {
+        if (array_key_exists('scheme', $this->locationParts)) {
```
mvriel: it is more common to use `isset($this->locationParts['scheme'])` but I don't know if and what the advantage of one over the other is :)

src/phpDocumentor/Dsn.php
```diff
@@ -0,0 +1,308 @@
...
+        $this->parsePassword();
+
+        $this->parseQuery();
+
+        $this->parseParameters();
+    }
+
+    /**
+     * validates and sets the scheme property
+     */
```
mvriel: Missing the `@return void` tag

src/phpDocumentor/Dsn.php
```diff
@@ -0,0 +1,308 @@
...
+            $this->scheme = 'file';
+        }
+    }
+
+    /**
+     * Validates and sets the host and path properties
+     */
+    private function parseHostAndPath()
+    {
+        if (array_key_exists('scheme', $this->locationParts) && $this->locationParts['scheme'] === 'file') {
```
mvriel: You can simplify this by changing it to `getScheme() === 'file'` because you know it is set here

src/phpDocumentor/Dsn.php
```diff
@@ -0,0 +1,308 @@
...
+            $path = array_key_exists('path', $this->locationParts) ? $this->locationParts['path'] : "";
```
mvriel: You duplicate the retrieval of a 'path' here (including existance check); you can remove this duplication by setting a variable $path at the beginning of this method

src/phpDocumentor/Dsn.php
```diff
@@ -0,0 +1,308 @@
...
+            if ($result === 0) {
+                throw new \InvalidArgumentException(
+                    sprintf('"%s" is not a valid path.', $this->locationParts['path'])
+                );
+            }
+
+            $this->path = $path;
+
+        } else {
+            if (array_key_exists('host', $this->locationParts)) {
```
mvriel: You duplicate the retrieval of a 'host' here (including existance check); you can remove this duplication by setting a variable $path at the beginning of this method

src/phpDocumentor/Dsn.php
mvriel: In your parsing methods there are a fair number of regular expressions to validate parts of your DSN. Regular expressions can generally be quite slow and in this case we can improve performance, and readability, with the following optimization:

Before parsing your parts you check if the scheme is set (thus after the parse_url call) and if so: prefix the $location with `file://`. The result of $location should then form a valid URL which you can check with http://php.net/manual/en/function.filter-var.php

After that you have validated the complete URL and you don't need to do validations of the separate components

mvriel:
@mbed67 I have finished the review; you have been more thorough I would probably have been, which is really good :) One thing I can provide is that you have some duplication in which regular expressions that you do, try to put a single preg_match in its own method with a descriptive name as you won't be able to read the reg ex later easily.

Another thing is that it is convenient with the key=value parts to first split it into a key and a value and validate those individually instead of doing both in the same regex; that will also improve the readability of your code.

src/phpDocumentor/Dsn.php
```diff
@@ -0,0 +1,232 @@
...
+    /**
+     * Parses the given DSN
+     *
+     * @param string $dsn
+     */
+    private function parse($dsn)
+    {
+        if (! is_string($dsn)) {
+            throw new \InvalidArgumentException(
+                sprintf('"%s" is not a valid DSN.', $dsn)
```
mvriel: This message won't work; since $dsn is not a string you cannot include it in a sprintf using %s; one solution would be to wrap it in a `var_export($dsn, true)` call. Using var_export with the second flag enabled will return a string, independent of the type of object

src/phpDocumentor/Dsn.php
```diff
@@ -0,0 +1,232 @@
...
+            );
+        }
+
+        $this->parseScheme($locationParts);
+
+        $this->parseHostAndPath($locationParts);
+
+        $this->port = isset($locationParts['port']) ? $locationParts['port'] : "";
```
mvriel: Port is an integer; so the default should be an integer (i.e. 0)

src/phpDocumentor/Dsn.php
```diff
@@ -0,0 +1,232 @@
...
+
+        $this->user = isset($locationParts['user']) ? $locationParts['user'] : "";
+
+        $this->password = isset($locationParts['pass']) ? $locationParts['pass'] : "";
+
+        $this->query = isset($locationParts['query']) ? explode('?', $locationParts['query']) : "";
```
mvriel: The type for the query is an array; here you set it as an empty string. The default should be an empty array

mvriel: Can you change this so that `$this->query` contains an associative array where the key is the query parameter key and the name is the value?

For example:
  
    http://github.com/phpDocumentor/phpDocumentor2?milestone=3.0,query=abc

Should result in a query array that looks like this:

```
$query = [
    'milestone' => '3.0',
    'query' => 'abc'
];
```

src/phpDocumentor/Dsn.php
```diff
@@ -0,0 +1,232 @@
...
+    /**
+     * validates and sets the parameters property
+     *
+     * @return void
+     */
+    private function parseParameters($dsnParts)
+    {
+        if (!empty($dsnParts)) {
+            foreach ($dsnParts as $part) {
+                if (! strpos($part, '=')) {
```
mvriel: small performance tip: you can do the explode before this check and instead of checking using strpos you can count the $option array; if it isn't 2 then the part is not valid

src/phpDocumentor/Dsn.php
```diff
@@ -0,0 +1,288 @@
...
+        }
+    }
+
+    /**
+     * validates and sets the parameters property
+     *
+     * @param string $dsnParts
+     * @return void
+     */
+    private function parseParameters($dsnParts)
```
mvriel: Is $dsnParts a string or array? If it is an array, please add a typehint

src/phpDocumentor/Dsn.php
```diff
@@ -0,0 +1,288 @@
...
+    {
+        if (!empty($dsnParts)) {
```
mvriel: This check is not necessary if $dsnParts is an array; if $dsnParts is empty then the foreach will never be entered

src/phpDocumentor/Dsn.php
```diff
@@ -0,0 +1,288 @@
...
+    /**
+     * Splits a key-value pair
+     *
+     * @param string $pair
+     * @return string[] $option
+     */
+    private function splitKeyValuePair($pair)
+    {
+        $option = explode('=', $pair);
+        if (count($option) != 2) {
```
mvriel: Always try to use `===` and `!==` to test values; when you use `==` or `!=` will PHP perform type juggling, which can cause odd issues such as complete strings equating to 0; see this post for how this can cause security issues when you apply this with password checking for example: https://blog.whitehatsec.com/magic-hashes/

mvriel:
This PR looks good; I would merge it but I am missing a unit test. Can you indicate if you want to add that to this PR or submit a new PR with the test?

mvriel:
As discussed on slack; there is a separate PR for the test

PR #1588, opened 2015-06-03
Value object for Uri

rgeraads:
the Uri value object for usage within the ConfigurationFactory

### DIFF
--- a/src/phpDocumentor/Uri.php
+++ b/src/phpDocumentor/Uri.php
@@ -12,6 +12,11 @@
 
 namespace phpDocumentor;
 
+/**
+ * Class Uri
+ *
+ * @package phpDocumentor
+ */
 final class Uri
 {
     /**
@@ -19,22 +24,65 @@ final class Uri
      */
     private $uri;
 
+    /**
+     * @param string $uri
+     */
     public function __construct($uri)
     {
-        $this->uri = (string)$uri;
+        $this->validateString($uri);
 
-        $this->validate($uri);
+        $uri = $this->checkScheme($uri);
+
+        $this->validateUri($uri);
+
+        $this->uri = $uri;
     }
 
+    /**
+     * @return string
+     */
     public function __toString()
     {
         return $this->uri;
     }
 
-    private function validate($uri)
+    /**
+     * Checks if $uri is of type string.
+     *
+     * @param  mixed $uri
+     * @throws \InvalidArgumentException
+     */
+    private function validateString($uri)
+    {
+        if (!is_string($uri)) {
+            throw new \InvalidArgumentException(sprintf('String required, %s given', gettype($uri)));
+        }
+    }
+
+    /**
+     * @param  string $uri
+     * @throws \InvalidArgumentException
+     */
+    private function validateUri($uri)
     {
         if (filter_var($uri, FILTER_VALIDATE_URL) === false) {
-            throw new \InvalidArgumentException('Invalid uri');
+            throw new \InvalidArgumentException(sprintf('%s is not a valid uri', $uri));
         }
+    }
+
+    /**
+     * Checks if a scheme is present
+     * If no scheme is found, it is assumed that a local path is used, and file:// is prepended.
+     *
+     * @param  string $uri
+     * @return string
+     */
+    private function checkScheme($uri)
+    {
+        if (!parse_url($uri, PHP_URL_SCHEME)) {
+            $uri = 'file://' . $uri;
+        }
+
+        return $uri;
     }
 }
--- a/src/phpDocumentor/Uri.php
+++ b/src/phpDocumentor/Uri.php
@@ -13,9 +13,8 @@
 namespace phpDocumentor;
 
 /**
- * Class Uri
- *
- * @package phpDocumentor
+ * Value object for uri.
+ * Uri can be either local or remote.
  */
 final class Uri
 {
@@ -31,7 +30,7 @@ final class Uri
     {
         $this->validateString($uri);
 
-        $uri = $this->checkScheme($uri);
+        $uri = $this->addFileSchemeWhenSchemeIsAbsent($uri);
 
         $this->validateUri($uri);
 
@@ -39,6 +38,8 @@ final class Uri
     }
 
     /**
+     * Returns a string representation of the uri.
+     *
      * @return string
      */
     public function __toString()
@@ -49,8 +50,9 @@ final class Uri
     /**
      * Checks if $uri is of type string.
      *
-     * @param  mixed $uri
-     * @throws \InvalidArgumentException
+     * @param  string $uri
+     * @throws \InvalidArgumentException if $uri is not a string
+     * @return void
      */
     private function validateString($uri)
     {
@@ -60,8 +62,11 @@ final class Uri
     }
 
     /**
+     * Checks if $uri is valid.
+     *
      * @param  string $uri
-     * @throws \InvalidArgumentException
+     * @throws \InvalidArgumentException if $uri is not a valid uri
+     * @return void
      */
     private function validateUri($uri)
     {
@@ -77,7 +82,7 @@ final class Uri
      * @param  string $uri
      * @return string
      */
-    private function checkScheme($uri)
+    private function addFileSchemeWhenSchemeIsAbsent($uri)
     {
         if (!parse_url($uri, PHP_URL_SCHEME)) {
             $uri = 'file://' . $uri;
--- a/src/phpDocumentor/Uri.php
+++ b/src/phpDocumentor/Uri.php
@@ -1,0 +1,40 @@
+<?php
+/**
+ * This file is part of phpDocumentor.
+ *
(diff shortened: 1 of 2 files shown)


src/phpDocumentor/Uri.php
```diff
@@ -0,0 +1,40 @@
...
+    /**
+     * @var string
+     */
+    private $uri;
+
+    public function __construct($uri)
+    {
+        $this->uri = (string)$uri;
+
+        $this->validate($uri);
```
mvriel: It is common to validate a value before handling it (in this case assigning it to the property); this is because you want to fail early as this will improve performance and because handling an invalid variable might give unexpected issues. In this specific instance it will not cause an issue but it is a good thing to get into your system

tests/unit/phpDocumentor/UriTest.php
```diff
@@ -0,0 +1,38 @@
...
+    {
+        $uri = new Uri('file://foo');
+
+        $this->assertInstanceOf('phpDocumentor\Uri', $uri);
+        $this->assertObjectHasAttribute('uri', $uri);
+    }
+
+    public function testItShouldDiscardAnInvalidUri()
+    {
+        $this->setExpectedException('\InvalidArgumentException', 'Invalid uri');
```
mvriel: Did you know that you can use annotations in phpunit to set this?

```
/**
 * @expectedException InvalidArgumentException
 * @expectedExceptionMessage Invalid uri
 */
```

You don't have to change this but wanted to make you aware of this in case you didn't know :)

mvriel:
@rgeraads I have the following feedback regarding this PR:

1. Can you add documentation in the form of DocBlocks to the methods and class?
2. Can you add `@covers` annotations to the test methods (see `phpDocumentor/Guides/ContentTypeTest.php` for a simple example)
3. Can you add support for schemeless URIs to re-map to local files (see next paragraphs for more information on this)

#### Schemeless URIs

When I pass a configuration location using the command line I will probably refer to a local file using the form `app/phpdoc.xml`. To support this I would need to convert it to a URL first by prefixing it with `file://` but what would be more convenient (because less conditions would be needed in the application) is if the URI class assumes that any strings passed without a scheme will have the `file://` scheme.

So for example:

> When I pass the string `app/phpdoc.xml` to the URI class it will detect that it has no scheme and translate the string into `file://app/phpdoc.xml`.

The easiest way to check for the scheme is by passing the string through the [parse_url](nl3.php.net/manual/en/function.parse-url.php) php function.

src/phpDocumentor/Uri.php
rgeraads: nice find!

tests/unit/phpDocumentor/UriTest.php
rgeraads: I explicitly changed it from annotations to a method. I don't know what is preferred, or better. It felt better this way (probably consistency) since all the tests have 'test' prepended so the exception test would be the only test that needs the docblock.

mvriel: So far the project uses the `expectedException` annotations

rgeraads: do you want me to change it to annotations? I don't mind changing it to annotations.

mvriel: Yes please :)

src/phpDocumentor/Uri.php
```diff
@@ -0,0 +1,88 @@
...
+     */
+    public function __toString()
+    {
+        return $this->uri;
+    }
+
+    /**
+     * Checks if $uri is of type string.
+     *
+     * @param  mixed $uri
```
mvriel: type is not mixed but string

src/phpDocumentor/Uri.php
```diff
@@ -0,0 +1,88 @@
...
+     * @throws \InvalidArgumentException
```
mvriel: Can you add a description with the exception that describes when this is thrown?

src/phpDocumentor/Uri.php
```diff
@@ -0,0 +1,88 @@
...
+     */
```
mvriel: Missing `@return void`

src/phpDocumentor/Uri.php
```diff
@@ -0,0 +1,88 @@
...
+ */
+final class Uri
+{
+    /**
+     * @var string
+     */
+    private $uri;
+
+    /**
+     * @param string $uri
```
mvriel: missing summary

src/phpDocumentor/Uri.php
```diff
@@ -0,0 +1,88 @@
...
+ * @license   http://www.opensource.org/licenses/mit-license.php MIT
+ * @link      http://phpdoc.org
+ */
+
+namespace phpDocumentor;
+
+/**
+ * Class Uri
+ *
+ * @package phpDocumentor
```
mvriel: Can you remove this use of `@package`? phpDocumentor does not use that

src/phpDocumentor/Uri.php
```diff
@@ -0,0 +1,88 @@
...
+
+        $uri = $this->checkScheme($uri);
+
+        $this->validateUri($uri);
+
+        $this->uri = $uri;
+    }
+
+    /**
+     * @return string
```
mvriel: missing summary

src/phpDocumentor/Uri.php
```diff
@@ -0,0 +1,88 @@
...
+     */
+    private function validateString($uri)
+    {
+        if (!is_string($uri)) {
+            throw new \InvalidArgumentException(sprintf('String required, %s given', gettype($uri)));
+        }
+    }
+
+    /**
+     * @param  string $uri
```
mvriel: missing summary

src/phpDocumentor/Uri.php
```diff
@@ -0,0 +1,88 @@
...
+     * @throws \InvalidArgumentException
```
mvriel: can you add a description that describes when the exception is thrown

src/phpDocumentor/Uri.php
```diff
@@ -0,0 +1,88 @@
...
+     */
```
mvriel: missing `@return void`

src/phpDocumentor/Uri.php
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    /**
+     * Checks if a scheme is present
+     * If no scheme is found, it is assumed that a local path is used, and file:// is prepended.
+     *
+     * @param  string $uri
+     * @return string
+     */
+    private function checkScheme($uri)
```
mvriel: can you rename this method to `addFileSchemeWhenSchemeIsAbsent($uri)`? It describes the function of the method; mainly because you don't check anything here (there is nothing validated and there are no consequences when something is wrong)

tests/unit/phpDocumentor/UriTest.php
```diff
@@ -0,0 +1,74 @@
...
+/**
+ * Test case for Uri
+ *
+ * @coversDefaultClass phpDocumentor\Uri
+ */
+final class UriTest extends \PHPUnit_Framework_TestCase
+{
+    /**
+     * @covers ::__construct
+     * @covers ::validateString
```
mvriel: Can you replace the private methods with `@covers ::<private>`? You are missing some items here and my Code Coverage will complain ;)

tests/unit/phpDocumentor/UriTest.php
```diff
@@ -0,0 +1,74 @@
...
+     * @covers ::validateUri
+     */
+    public function testItShouldBuildAUri()
+    {
+        $uri = new Uri('file://foo');
+
+        $this->assertInstanceOf(Uri::class, $uri);
```
mvriel: This check is redundant; there is no way that the $uri can be anything other than a class of type Uri

tests/unit/phpDocumentor/UriTest.php
```diff
@@ -0,0 +1,74 @@
...
+        $this->assertObjectHasAttribute('uri', $uri);
```
mvriel: This check is redundant, there is no way that the Uri doesn't have the property `uri`.

What you probably want to check is if the `__toString()` returns the Uri that you passed

mvriel:
Looking good! I only have found a few style issues and there is a test that doesn't do much.
Please do read up on the `@covers` and `@uses` tags with phpunit; I noticed a few tests where code coverage would have complained because of used but uncovered code.

More information on that topic can be found here: https://thephp.cc/news/2014/03/phpunit-4-0-code-coverage-improvements

What I do myself is run my tests from phpstorm with code coverage on (the button with a play icon and a round mesh behind it); that way I see if I forgot an `@uses` or `@covers` and if I have tested everything in the class that I test.

If you can fix these last remarks there is nothing stopping me from merging

mvriel:
Excellent!!

PR #1591, opened 2015-06-08
Several small corrections in Dsn and added a unit test for Dsn

mbed67:


tests/unit/phpDocumentor/DsnTest.php
```diff
@@ -0,0 +1,160 @@
+<?php
+/**
```
mvriel: You have used the old DocBlock here; can you replace it for the new one?

tests/unit/phpDocumentor/DsnTest.php
```diff
@@ -0,0 +1,160 @@
...
+        $this->assertEquals($parameters, $fixture->getParameters());
+    }
+
+    /**
+     * @covers ::__construct
+     * @covers ::getScheme
+     * @covers ::getHost
+     * @covers ::getPort
+     * @covers ::getPath
+     * @covers ::parse
```
mvriel: did you know that you can use `::<private>` to indicate 'any private method'? I consider private to be a free-for-all since you cannot really unintentionally cover a private; while you can with a public thing like another class

mvriel:
No problems found; if you can replace the DocBlock I am perfectly happy :)

PR #1592, opened 2015-06-10
Created FilesystemFactory with a unit test

mbed67:


src/phpDocumentor/FilesystemFactory.php
```diff
@@ -0,0 +1,89 @@
...
+
+    /**
+     * Returns a Filesystem instance based on the scheme of the provided Dsn
+     *
+     * @param Dsn $dsn
+     * @return Filesystem
+     */
+    public function create(Dsn $dsn)
+    {
+        $dsnId = spl_object_hash($dsn);
```
mvriel: Interesting approach on creating a key! It does have one downside that is a bit technical and theoretical:

The DSN is a Value Object and the concept behind Value Objects is that they are immutable and can be recreated at any time. Since spl_object_hash returns a different id if a different object is used this will potentially result in multiple cache entries to the same location.

When we look at what is identifying for a location (because each Filesystem object is unique for each location) we can state that the DSN string in the DSN object is unique because that doesn't change, even when another DSN object is used.

src/phpDocumentor/FilesystemFactory.php
```diff
@@ -0,0 +1,89 @@
...
+     * @param string $dsnId
+     * @return bool
+     */
+    private function setFilesystemWhenCached($dsnId)
+    {
+        try {
+            $this->filesystem = $this->mountManager->getFilesystem($dsnId);
+            return true;
+        } catch (LogicException $e) {
+            echo 'Caught exception: ',  $e->getMessage(), "\n";
```
mvriel: Please do not use `echo` (or otherwise directly write to the STDOUT) in a code base when you want to display debug data; most codebases have logging mechanisms to deal with this, and if they don't (like phpDocumentor 3) you can use `trigger_error`.

In this case you can wonder why an exception occurs and what happens now. If no filesystem property was present before than this class will have a fatal error once it hits line 59 because the filesystem property is empty; or worse: if a previous call to create was successful and this wasn't than the wrong file system object is returned because it will return the previous call's filesystem.

> This is actually a really good reason why you don't want to store the filesystem in a property but keep the filesystem local to the create method, if you return the property in this method instead of a boolean then you don't need the property at all.

src/phpDocumentor/FilesystemFactory.php
```diff
@@ -0,0 +1,89 @@
...
+        }
+    }
```
mvriel: Inconsistent return; there is no return false

src/phpDocumentor/FilesystemFactory.php
```diff
@@ -0,0 +1,89 @@
...
+     */
+    public function create(Dsn $dsn)
+    {
+        $dsnId = spl_object_hash($dsn);
+
+        if (! $this->setFilesystemWhenCached($dsnId)) {
+            if ($dsn->getScheme() === 'file') {
+
+                $path = $dsn->getPath();
+                $this->filesystem = new Filesystem(new Local(__DIR__. "/" . $path));
```
mvriel: The path determination can cause issues: the $path may contain an absolute path (i.e. `file:///home/mvriel/Projects/phpDocumentor`) and if that happens it will be prefixed with the directory of this FilesystemFactory (because `__DIR__` does not return the current working directory, `getcwd` does).

Please check if the Local adapter supports receiving relative paths because you might not need to have to prefix the path.

src/phpDocumentor/FilesystemFactory.php
```diff
@@ -0,0 +1,89 @@
...
+     * @return Filesystem
+     */
+    public function create(Dsn $dsn)
+    {
+        $dsnId = spl_object_hash($dsn);
+
+        if (! $this->setFilesystemWhenCached($dsnId)) {
+            if ($dsn->getScheme() === 'file') {
+
+                $path = $dsn->getPath();
```
mvriel: You can see an excellent example of why your implementation of the DSN shines here; you don't have to worry at the implementation side whether you would need to concatenate the host and path (as is the case with `file://); because the Value Object makes the translation it is only on one place instead of each caller

src/phpDocumentor/FilesystemFactory.php
```diff
@@ -0,0 +1,89 @@
...
+    private $filesystem;
+
+    /**
+     * @param MountManager $mountManager
+     * @param Container $container
+     */
+    public function __construct(MountManager $mountManager, Container $container)
+    {
+        $this->mountManager = $mountManager;
+        $this->container = $container;
```
mvriel: The container is not used and can be removed

mvriel:
@mbed67 despite the long texts there was only one real bug: because you use a filesystem property you will have issues where, when a LogicException occurs, the previous created filesystem may be returned instead of the new one.

A good rule of thumb is: if two public methods do not share a piece of data (variable) then it should not be a property. Properties are only used to communicate between scenarios across multiple public methods

src/phpDocumentor/FilesystemFactory.php
```diff
@@ -0,0 +1,64 @@
...
+     * @param Dsn $dsn
+     * @return Filesystem
+     */
+    public function create(Dsn $dsn)
+    {
+        $dsnId = hash('md5', (string)$dsn);
+
+        try {
+            $filesystem = $this->mountManager->getFilesystem($dsnId);
+        } catch (LogicException $e) {
```
mvriel: Instead of relying on exceptions (which are circumstantial evidence because you do not know for sure that a LogicException means the FS is missing) you can use the `has()` method of the mount manager to check if your intended FS already exists.

By using `has` you can rely on exceptions to be exceptions in the logic and not part of your business logic.

mvriel:
@mbed67 I am going to merge this because functionally this works as intended; I have left one comment regarding the use of exceptions; can you fix that in a new PR?

PR #1596, opened 2015-06-19
remove var_dump

bpaulin:
Hello,

I can't find a good reason for this var_dump to be here, and it makes my build outputs dirty :)

thanks for your work!


mvriel:
Thanks :)

PR #1597, opened 2015-06-22
Feature/configuration factory

rgeraads:
The ConfigurationFactory that converts phpdocumento2 / phpdocumento3 xml to an array


mvriel:
Hi @rgeraads,

You're off to a good start! During the code review I have found the following points of improvements:

- [ ] Can you please move the ConfigurationFactory to the `\phpDocumentor` namespace?
- [ ] Can you move/use the xsd to/in the `/data/xsd` folder?
- [ ] I am missing the `replaceLocation()` and `get()` method (see Class Diagram); it seems you renamed that to `convert`?
- [ ] The XML is retrieved in the constructor; can you move that to the `get()` method and in the constructor just record the URI?
    > This is because the configuration file may be overridden during execution of the application and we want to load the configuration at the latest moment, and that is in the `get`. Please keep in mind that the `get` may be called multiple times and building the configuration should not reoccur unless the URI has changed. This will save performance.
- [ ] A significant amount of information is missing from the generated array for the phpDocumentor 2  XML; the output folder and source files are not set

In addition to the above I will add a few comments inline with the source.

src/phpDocumentor/Configuration/ConfigurationFactory.php
```diff
@@ -0,0 +1,332 @@
...
+     * @param \SimpleXMLElement $version
+     *
+     * @return array
+     */
+    private function buildVersions(\SimpleXMLElement $version)
+    {
+        return [
+            'folder' => (string) $version->folder,
+            'api'    => [
+                'format'               => 'php',
```
mvriel: this should also come from the xml

src/phpDocumentor/Configuration/ConfigurationFactory.php
```diff
@@ -0,0 +1,332 @@
...
+                'source'               => [
+                    'dsn'   => 'file://.',
```
mvriel: This also comes from the DSN

src/phpDocumentor/Configuration/ConfigurationFactory.php
```diff
@@ -0,0 +1,332 @@
...
+                    'paths' => [
+                        0 => 'src'
```
mvriel: This also comes from the XML

src/phpDocumentor/Configuration/ConfigurationFactory.php
```diff
@@ -0,0 +1,332 @@
...
+                    ]
+                ],
+                'ignore'               => [
+                    'hidden'   => filter_var($version->api->ignore->attributes()->hidden, FILTER_VALIDATE_BOOLEAN),
+                    'symlinks' => filter_var($version->api->ignore->attributes()->symlinks, FILTER_VALIDATE_BOOLEAN),
+                    'paths'    => (array) $version->api->ignore->path,
+                ],
+                'extensions'           => [
```
mvriel: The extensions are also configurable from the XML

src/phpDocumentor/Configuration/ConfigurationFactory.php
```diff
@@ -0,0 +1,332 @@
...
+                    0 => 'php',
+                    1 => 'php3',
+                    2 => 'phtml'
+                ],
+                'visibility'           => 'public',
```
mvriel: This is also configurable in the XML

src/phpDocumentor/Configuration/ConfigurationFactory.php
```diff
@@ -0,0 +1,332 @@
...
+                'default-package-name' => 'Default',
```
mvriel: This is also configurable in the XML

src/phpDocumentor/Configuration/ConfigurationFactory.php
```diff
@@ -0,0 +1,332 @@
...
+                'markers'              => (array) $version->api->markers->children()->marker,
+            ],
+            'guide'  => [
```
mvriel: The XML can contain multiple guide definitions

src/phpDocumentor/Configuration/ConfigurationFactory.php
```diff
@@ -0,0 +1,332 @@
...
+     *
+     * @param \SimpleXMLElement $version
+     *
+     * @return array
+     */
+    private function buildVersions(\SimpleXMLElement $version)
+    {
+        return [
+            'folder' => (string) $version->folder,
+            'api'    => [
```
mvriel: The xml can contain multiple API definitions

src/phpDocumentor/Configuration/ConfigurationFactory.php
```diff
@@ -0,0 +1,332 @@
...
+     */
+    private function buildTemplate(\SimpleXMLElement $template)
+    {
+        if (!$template) {
+            // Use default template if none is found in the configuration
+            return [
+                0 => [
+                    'name' => 'clean'
+                ],
+                1 => [
```
mvriel: This is not necessary; this was just an example on how you can use an external location to load an XML file

src/phpDocumentor/Configuration/ConfigurationFactory.php
```diff
@@ -0,0 +1,332 @@
...
+                if (isset($phpDocumentor->parser->files->{'ignore-symlinks'})) {
+                    $ignoreSymlinks = filter_var($phpDocumentor->parser->files->{'ignore-symlinks'}, FILTER_VALIDATE_BOOLEAN);
+                }
+            }
+        }
+
+        $phpdoc2Array = [
+            'phpdocumentor' => [
+                'paths'     => [
+                    'output' => 'file://build/docs',
```
mvriel: This should come from the XML

src/phpDocumentor/Configuration/ConfigurationFactory.php
```diff
@@ -0,0 +1,332 @@
...
+                    'cache'  => '/tmp/phpdoc-doc-cache'
+                ],
+                'versions'  => [
+                    '1.0.0' => [
+                        'folder' => 'latest',
```
mvriel: Can you make this empty? The default output (sub)folder for the version 2 xml is `''` (otherwise we'd break BC a bit too much ;))

src/phpDocumentor/Configuration/ConfigurationFactory.php
```diff
@@ -0,0 +1,332 @@
...
+                        'api'    => [
+                            'format'               => 'php',
+                            'source'               => [
+                                'dsn'   => 'file://.',
+                                'paths' => [
```
mvriel: This should come from the XML

src/phpDocumentor/Configuration/ConfigurationFactory.php
```diff
@@ -0,0 +1,332 @@
...
+                                    0 => 'src'
+                                ]
+                            ],
+                            'ignore'               => [
+                                'hidden'   => $ignoreHidden,
+                                'symlinks' => $ignoreSymlinks,
+                                'paths'    => [
```
mvriel: This should come from the XML; these are the ignored paths

src/phpDocumentor/Configuration/ConfigurationFactory.php
```diff
@@ -0,0 +1,332 @@
...
+                                    0 => 'src/ServiceDefinitions.php'
+                                ]
+                            ],
+                            'extensions'           => $extensions,
+                            'visibility'           => $visibility,
+                            'default-package-name' => $defaultPackageName,
+                            'markers'              => $markers,
+                        ],
+                        'guide'  => [
```
mvriel: There is no guide entry in the original XML; this section can be removed

src/phpDocumentor/Configuration/ConfigurationFactory.php
```diff
@@ -0,0 +1,332 @@
...
+                                ]
+                            ]
+                        ]
+                    ]
+                ],
+                'templates' => [
+                    0 => [
+                        'name' => $template
+                    ],
+                    1 => [
```
mvriel: this can be removed; it was just an example

src/phpDocumentor/Configuration/ConfigurationFactory.php
```diff
@@ -0,0 +1,332 @@
...
+                            'source' => [
+                                'dsn'   => 'file://../phpDocumentor/phpDocumentor2',
+                                'paths' => [
+                                    0 => 'docs'
+                                ]
+                            ]
+                        ]
+                    ]
+                ],
+                'templates' => [
```
mvriel: There may be multiple `<template>` tags in the phpDocumentor2 XML because you can execute multiple templates in succession

mvriel:
I have updated the original review post with two more items.

@rgeraads I have completed my review; due to some extensive changes that are needed I have not reviewed the tests and will do that after the suggested changes. This is an excellent start and I am looking forward to the end result!

src/phpDocumentor/ConfigurationFactory.php
```diff
@@ -0,0 +1,354 @@
...
+    /**
+     * Replace the location of the configuration file.
+     *
+     * @param Uri $uri
+     *
+     * @return ConfigurationFactory
+     */
+    public function replaceLocation(Uri $uri)
+    {
+        return static::__construct($uri);
```
mvriel: please set the uri property here instead of calling the constructor. If someone were to use a different $schemaPath then it would revert to the default after calling this method

src/phpDocumentor/ConfigurationFactory.php
```diff
@@ -0,0 +1,354 @@
...
+    /**
+     * @var string
+     */
+    private $schemaPath;
+
+    /**
+     * @param Uri    $uri
+     * @param string $schemaPath
+     */
+    public function __construct(Uri $uri, $schemaPath = '../../data/xsd/phpdoc.xsd')
```
mvriel: a path is relative to the current working directory. always prefix with the __DIR__ constant if you intend to indicate a path relative to the current file

src/phpDocumentor/ConfigurationFactory.php
rgeraads: I changed it, but I would like to know how this could revert to default. Could you explain further?

src/phpDocumentor/ConfigurationFactory.php
rgeraads: down below i use ```$this->schemaPath``` as follows: ```$dom->schemaValidate(__DIR__ . DIRECTORY_SEPARATOR . $this->schemaPath);```
If that is not correct, I can change it in the constructor

mvriel: Please move that to the constructor; the $schemaPath may be an absolute path so you cannot prefix it with `__DIR__` by default.

src/phpDocumentor/ConfigurationFactory.php
mvriel: The second parameter ($schemaPath) of __construct is a property with a default value; you omit the second parameter here so it will assume that default value. Subsequently that default value is set on the property and will overwrite the $schemaPath that you have set when you created the object

src/phpDocumentor/Configuration/ConfigurationFactory.php
rgeraads: do you mean the entire templates section?

mvriel: No, just the second entry; the templates entry should be populated not with hardcoded entries but with all entries from the configuration file (please note: the config can refer to multiple templates)

src/phpDocumentor/ConfigurationFactory.php
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    /**
+     * Validates if the Uri contains an xml that has a root element which name is phpdocumentor.
+     *
+     * @param Uri $uri
+     *
+     * @throws \InvalidArgumentException if the root element of the xml is not phpdocumentor.
+     */
+    private function validate(Uri $uri)
```
mvriel: This method has too many responsibilities; it is both responsible for validating XML and for loading that XML from the given URI. It would be better to move the loading of the XML to another method

src/phpDocumentor/ConfigurationFactory.php
```diff
@@ -0,0 +1,138 @@
...
+
+    /**
+     * Replaces the location of the configuration file if it is different.
+     *
+     * @param Uri $uri
+     */
+    public function replaceLocation(Uri $uri)
+    {
+        if ($this->uri !== $uri) {
+            $this->validateUri = true;
```
mvriel: Instead of tracking the cache using a boolean you can also clear the xml property and force the loading method to reload from the current uri. This will reduce the number of properties and hence reduce the complexity

src/phpDocumentor/ConfigurationFactory.php
```diff
@@ -0,0 +1,138 @@
...
+    {
+        return isset($phpDocumentor->attributes()->version);
+    }
+
+    /**
+     * Validates the phpDocumentor3 xml structure against phpdoc.xsd
+     *
+     * @param $phpDocumentor
+     */
+    private function validateXmlStructure(\SimpleXMLElement $phpDocumentor)
```
mvriel: This validation routine is specific to the phpdoc3 validation, shouldn't it be inside that strategy?

src/phpDocumentor/ConfigurationFactory.php
```diff
@@ -0,0 +1,138 @@
...
+     * @return array
+     */
+    public function get()
+    {
+        $this->validate($this->uri);
+
+        $version = $this->checkIfVersionAttributeIsPresent($this->xml);
+        if ($version) {
+            $this->validateXmlStructure($this->xml);
+            $xml = new Phpdoc3XmlToArrayConverter($this->xml);
```
mvriel: You have hardcoded the strategies but it is more convenient to inject which strategies there are in the constructor and in this location loop over every registered strategy and use a `match` method in a strategy class to check if the XML can be handled by that strategy; if it can be handled you can pass the xml to the `handle` method which does the actual conversion.

There are several advantages to injecting classes:

1. You can test them individually; now you need to test each of these classes using a unit test for the factory and hence you do not test a single unit
2. You can pass dependencies specific to a strategy; because you instantiate the strategy object outside of this class you can use varying constructor arguments and thus inject separate dependencies per strategy

src/phpDocumentor/Phpdoc2XmlToArrayConverter.php
```diff
@@ -0,0 +1,154 @@
...
+{
+    /**
+     * @var \SimpleXMLElement
+     */
+    private $xml;
+
+    /**
+     * @param \SimpleXMLElement $xml
+     */
+    public function __construct(\SimpleXMLElement $xml)
```
mvriel: because you pass the xml to the constructor you can't re-use this service (for example: when replaceLocation is used) and you need to create a new service for each xml. This is not necessary as this service does not depend on this state; you can just pass the xml to the convert method and use it there

src/phpDocumentor/Phpdoc2XmlToArrayConverter.php
```diff
@@ -0,0 +1,154 @@
...
+                                'dsn'   => 'file://.',
+                                'paths' => [
+                                    0 => (string) (new Dsn($sourcePath))->getPath(),
+                                ]
+                            ],
+                            'ignore'               => [
+                                'hidden'   => $ignoreHidden,
+                                'symlinks' => $ignoreSymlinks,
+                                'paths'    => [
+                                    0 => 'src/ServiceDefinitions.php'
```
mvriel: The ignore paths should not be hardcoded but read from the XML

src/phpDocumentor/Phpdoc2XmlToArrayConverter.php
```diff
@@ -0,0 +1,154 @@
...
+                ],
+                'versions'  => [
+                    '1.0.0' => [
+                        'folder' => '',
+                        'api'    => [
+                            'format'               => 'php',
+                            'source'               => [
+                                'dsn'   => 'file://.',
+                                'paths' => [
+                                    0 => (string) (new Dsn($sourcePath))->getPath(),
```
mvriel: You can supply multiple files and folders in the XML file of phpDocumentor; this is only one

src/phpDocumentor/Phpdoc2XmlToArrayConverter.php
```diff
@@ -0,0 +1,154 @@
...
+                                'paths'    => [
+                                    0 => 'src/ServiceDefinitions.php'
+                                ]
+                            ],
+                            'extensions'           => $extensions,
+                            'visibility'           => $visibility,
+                            'default-package-name' => $defaultPackageName,
+                            'markers'              => $markers,
+                        ],
+                        'guide'  => [
```
mvriel: There is no support for guides with phpDocumentor2, this entry should be removed

src/phpDocumentor/Phpdoc2XmlToArrayConverter.php
```diff
@@ -0,0 +1,154 @@
...
+                            ]
+                        ]
+                    ]
+                ],
+                'templates' => [
+                    0 => [
+                        'name' => $template
+                    ],
+                    1 => [
+                        'location' => 'https://github.com/phpDocumentor/phpDocumentor2/tree/develop/data/templates/' . $template
```
mvriel: This is only an example in the array example made by @mbed67  and should be removed

src/phpDocumentor/Phpdoc2XmlToArrayConverter.php
```diff
@@ -0,0 +1,154 @@
...
+                                'dsn'   => 'file://../phpDocumentor/phpDocumentor2',
+                                'paths' => [
+                                    0 => 'docs'
+                                ]
+                            ]
+                        ]
+                    ]
+                ],
+                'templates' => [
+                    0 => [
```
mvriel: You can pass multiple template entries in the phpDocumentor 2 XML; these should all be added here

src/phpDocumentor/Phpdoc2XmlToArrayConverter.php
```diff
@@ -0,0 +1,154 @@
...
+                }
+
+                if (isset($phpDocumentor->parser->files->{'ignore-symlinks'})) {
+                    $ignoreSymlinks = filter_var($phpDocumentor->parser->files->{'ignore-symlinks'}, FILTER_VALIDATE_BOOLEAN);
+                }
+            }
+        }
+
+        $outputDirectory = ((string) $phpDocumentor->parser->target) ?: 'file://build/docs';
+        $sourcePath      = ((string) $phpDocumentor->parser->files->directory) ?: 'src';
```
mvriel: `$phpDocumentor->parser->files->directory` is an array of directory elements

mvriel: Missing the `file` element support in the `$phpDocumentor->parser->files` element

src/phpDocumentor/Phpdoc3XmlToArrayConverter.php
```diff
@@ -0,0 +1,138 @@
...
+{
+    /**
+     * @var \SimpleXMLElement
+     */
+    private $xml;
+
+    /**
+     * @param \SimpleXMLElement $xml
+     */
+    public function __construct(\SimpleXMLElement $xml)
```
mvriel: because you pass the xml to the constructor you can't re-use this service (for example: when replaceLocation is used) and you need to create a new service for each xml. This is not necessary as this service does not depend on this state; you can just pass the xml to the convert method and use it there

src/phpDocumentor/Phpdoc3XmlToArrayConverter.php
```diff
@@ -0,0 +1,138 @@
...
+        return [
+            'folder' => (string) $version->folder,
+            'api'    => [
+                'format'               => ((string) $version->api->attributes()->format) ?: 'php',
+                'source'               => [
+                    'dsn'   => (string) $version->api->source->attributes()->dsn,
+                    'paths' => ((array) $version->api->source->path) ?: ['src'],
+                ],
+                'ignore'               => [
+                    'hidden'   => ($ignoreHidden) ?: true,
```
mvriel: `If $ignoreHidden is true then return true, otherwise return true`?

I think you have an error in this line and the next

src/phpDocumentor/Phpdoc3XmlToArrayConverter.php
```diff
@@ -0,0 +1,138 @@
...
+                    'symlinks' => ($ignoreSymlinks) ?: true,
+                    'paths'    => ((array) $version->api->ignore->path) ?: ['src/ServiceDefinitions.php'],
```
mvriel: The default that you have added should be an empty array, that value was only provided as example

src/phpDocumentor/Phpdoc3XmlToArrayConverter.php
```diff
@@ -0,0 +1,138 @@
...
+                $extensions[] = (string) $extension;
+            }
+        }
+
+        $ignoreHidden   = filter_var($version->api->ignore->attributes()->hidden, FILTER_VALIDATE_BOOLEAN);
+        $ignoreSymlinks = filter_var($version->api->ignore->attributes()->symlinks, FILTER_VALIDATE_BOOLEAN);
+
+        return [
+            'folder' => (string) $version->folder,
+            'api'    => [
```
mvriel: A version can have `0..*` api elements; it may have none, it may have multiple

src/phpDocumentor/Phpdoc3XmlToArrayConverter.php
```diff
@@ -0,0 +1,138 @@
...
+                    'hidden'   => ($ignoreHidden) ?: true,
+                    'symlinks' => ($ignoreSymlinks) ?: true,
+                    'paths'    => ((array) $version->api->ignore->path) ?: ['src/ServiceDefinitions.php'],
+                ],
+                'extensions'           => $extensions,
+                'visibility'           => (string) $version->api->visibility,
+                'default-package-name' => ((string) $version->api->{'default-package-name'}) ?: 'Default',
+                'markers'              => (array) $version->api->markers->children()->marker,
+            ],
+            'guide'  => [
```
mvriel: A version can have `0..*` API elements, it may have none, it may have multiple

src/phpDocumentor/Phpdoc3XmlToArrayConverter.php
```diff
@@ -0,0 +1,138 @@
...
+                'format' => ((string) $version->guide->attributes()->format) ?: 'rst',
+                'source' => [
+                    'dsn'   => ((string) $version->guide->source->attributes()->dsn) ?: 'file://../phpDocumentor/phpDocumentor3',
```
mvriel: This default was only an example in the array example, if no DSN is provide then `file://.` is assumed

src/phpDocumentor/Phpdoc3XmlToArrayConverter.php
```diff
@@ -0,0 +1,138 @@
...
+    private function buildTemplate(\SimpleXMLElement $template)
+    {
+        if (!$template) {
+            // Use default template if none is found in the configuration
+            return [
+                0 => [
+                    'name' => 'clean'
+                ],
+                1 => [
+                    'location' => 'https://github.com/phpDocumentor/phpDocumentor2/tree/develop/data/templates/clean'
```
mvriel: This entry was only an example how a full path could be used and should be removed

src/phpDocumentor/Phpdoc3XmlToArrayConverter.php
```diff
@@ -0,0 +1,138 @@
...
+    /**
+     * Builds the template part of the array from the phpDocumentor3 configuration xml
+     *
+     * @param \SimpleXMLElement $template
+     *
+     * @return array
+     */
+    private function buildTemplate(\SimpleXMLElement $template)
+    {
+        if (!$template) {
```
mvriel: $template can never be `!`; you have provided a typehint and as such a valid SimpleXmlElement is always passed

mvriel:
Hey Randy, 

You have made considerable progress, that is good to see!

Due to the feedback that I have mentioned above I have not yet reviewed the unit tests. I will do that once the code is more stable.

There have been a few comments regarding how things could be a little bit better but I also found a few bugs. Please see the comments above.

Can you please move the Converter interface and concrete implementations to a namespace `ConfigurationFactory` use the following names:

`Converter` -> `ConfigurationFactory\Strategy`
`Phpdoc2XmlToArrayConverter ` -> `ConfigurationFactory\PhpDocumentor2`
`Phpdoc3XmlToArrayConverter ` -> `ConfigurationFactory\PhpDocumentor3`

In addition I would like default values (such as the default template name) to be injected into the strategies (for example using the convert method as a parameter) by the ConfigurationFactory; this way the default values are not duplicated between the two strategy classes.

src/phpDocumentor/ConfigurationFactory.php
rgeraads: I'm having trouble finding the correct strategy for doing this. One problem is that the phpdocumentor3 has a constructor argument, and the phpdocumentor2 strategy doesn't. 
I suppose having the version attribute is the case for ```match()```, but I can't seem to get the whole working.

src/phpDocumentor/Phpdoc3XmlToArrayConverter.php
rgeraads: I revised this part, but I found out that when the ignores are left out, they default to ```false```. Is that ok?

src/phpDocumentor/Phpdoc3XmlToArrayConverter.php
rgeraads: $template can result to false ```$template = {SimpleXMLElement} [0]```

One of the tests breaks when This part is removed.

src/phpDocumentor/ConfigurationFactory.php
```diff
@@ -0,0 +1,102 @@
...
+     */
+    private $xml;
+
+    /**
+     * @var string
+     */
+    private $schemaPath;
+
+    /**
+     * @param Uri    $uri
```
jaapio: Missing a summary here

src/phpDocumentor/ConfigurationFactory.php
```diff
@@ -0,0 +1,102 @@
...
+     * @var \SimpleXMLElement
+     */
+    private $xml;
+
+    /**
+     * @var string
+     */
+    private $schemaPath;
+
+    /**
```
jaapio: Missing method summary

src/phpDocumentor/ConfigurationFactory.php
```diff
@@ -0,0 +1,102 @@
...
+     */
+    private $uri;
+
+    /**
+     * @var \SimpleXMLElement
+     */
+    private $xml;
+
+    /**
+     * @var string
```
jaapio: Please add a summary to all properties.

src/phpDocumentor/ConfigurationFactory.php
```diff
@@ -0,0 +1,102 @@
+<?php
+
+namespace phpDocumentor;
+
+use phpDocumentor\ConfigurationFactory\PhpDocumentor2;
+use phpDocumentor\ConfigurationFactory\PhpDocumentor3;
+
+final class ConfigurationFactory
```
jaapio: A class needs a description

src/phpDocumentor/ConfigurationFactory.php
```diff
@@ -0,0 +1,119 @@
...
+
+        throw new \RuntimeException('No strategy found that matches the configuration xml');
+    }
+
+    /**
+     * Registers strategies that are used in the ConfigurationFactory.
+     *
+     * @return void
+     */
+    private function registerStrategies()
```
jaapio: I think it is better to add strategies as an parameter of the constructor. A strategy is an external service that is used by this class. So it needs to be injected and not created here. Just like you would do with a database connection that is passed in to a repository.

src/phpDocumentor/ConfigurationFactory.php
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    /**
+     * Validates if the xml has a root element which name is phpdocumentor.
+     *
+     * @param \SimpleXMLElement $xml
+     *
+     * @throws \InvalidArgumentException if the root element of the xml is not phpdocumentor.
+     */
+    private function validate(\SimpleXMLElement $xml)
```
jaapio: I think you should move this to the strategy to. The validation could defer per version. At the moment you are just checking the root tag. But for version 3 a xsd was created, you could use that for validation.

src/phpDocumentor/ConfigurationFactory.php
```diff
@@ -0,0 +1,119 @@
...
+
+    /**
+     * Initializes the ConfigurationFactory.
+     *
+     * @param Uri    $uri
+     * @param string $schemaPath
+     */
+    public function __construct(Uri $uri, $schemaPath = '')
+    {
+        if ($schemaPath === '') {
```
jaapio: Schema path is a phpdoc3 config thing only. So that should be injected into to the strategy. Since it may defer per version.

src/phpDocumentor/ConfigurationFactory/PhpDocumentor3.php
```diff
@@ -0,0 +1,176 @@
...
+    /**
+     * Builds the template part of the array from the phpDocumentor3 configuration xml.
+     *
+     * @param \SimpleXMLElement $template
+     *
+     * @return array
+     */
+    private function buildTemplate(\SimpleXMLElement $template)
+    {
+        if (!$template) {
```
jaapio: could you make this check more explicit? $template will never be false, I can imagine that there are situations that are not valid. But I would like to see what situation is is. Php will convert this situation to false. but that is not what it is. 

$template === false will never be true.

src/phpDocumentor/ConfigurationFactory/PhpDocumentor3.php
```diff
@@ -0,0 +1,176 @@
...
+        $dom        = new \DOMDocument();
+        $domElement = dom_import_simplexml($phpDocumentor);
+        $domElement = $dom->importNode($domElement, true);
+        $dom->appendChild($domElement);
+
+        $dom->schemaValidate($this->schemaPath);
+
+        $error = libxml_get_last_error();
+
+        if ($error) {
```
jaapio: Please make this check more explicit like $error !== false since libxml_get_last_error will return false when no error occurred.

src/phpDocumentor/ConfigurationFactory/Strategy.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+namespace phpDocumentor\ConfigurationFactory;
+
+interface Strategy
```
jaapio: please add the licence header, and missing docblocks for to this file

tests/data/phpDocumentor2ExpectedArray.php
```diff
@@ -0,0 +1,53 @@
+<?php
+
```
jaapio: Licence and docblocks are missing here. Not all methods need a docblock. But please explain why this class was needed.

tests/data/phpDocumentor3ExpectedArrays.php
```diff
@@ -0,0 +1,211 @@
+<?php
+
```
jaapio: Licence and docblocks are missing here. Not all methods need a docblock. But please explain why this class was needed.

jaapio:
@mvriel some minor changes are still needed. But in the end I think this one should be ready for merge.

src/phpDocumentor/ConfigurationFactory.php
```diff
@@ -0,0 +1,104 @@
...
+    private $strategies = [];
+
+    /**
+     * Initializes the ConfigurationFactory.
+     *
+     * @param Strategy[] $strategies
+     * @param Uri        $uri
+     * @param string     $schemaPath
+     */
+    public function __construct(array $strategies, Uri $uri, $schemaPath = '')
```
mvriel: schemaPath is an unused parameter and should be removed

src/phpDocumentor/ConfigurationFactory.php
```diff
@@ -0,0 +1,104 @@
...
+    /**
+     * Replaces the location of the configuration file if it differs from the existing one.
+     *
+     * @param Uri $uri
+     *
+     * @return void
+     */
+    public function replaceLocation(Uri $uri)
+    {
+        if ($this->uri !== $uri) {
```
mvriel: This check will only verify if the $uri is the exact same object but not whether the URLs themselves are the same (the identity of a Value Object is in its properties). 

So either

1. `!==` should be replaced with `!=`
2. the URI should be extracted from both Value Objects and compared, or 
3. a `equals()` method should be introduced in the URI class that will enable a URI to know how to compare itself to another URI.

The latter is the preferred form

src/phpDocumentor/ConfigurationFactory.php
```diff
@@ -0,0 +1,104 @@
...
+    /**
+     * Converts the phpDocumentor configuration xml to an array.
+     *
+     * @return array
+     *
+     * @throws \RuntimeException if no matching strategy can be found.
+     */
+    public function get()
+    {
+        foreach ($this->strategies as $strategy) {
```
mvriel: Protip (doesn't necessarily need to be changed at this stage): the `get()` method could be called multiple times during the life of the application but the parsing of the XML is a one-time thing. An optimalisation here would not be to cache the retrieved XML but to cache the converted array

src/phpDocumentor/ConfigurationFactory/PhpDocumentor2.php
```diff
@@ -0,0 +1,181 @@
...
+        $ignoreHidden       = true;
+        $ignoreSymlinks     = true;
+
+        if (isset($phpDocumentor->parser)) {
+            $extensions = $this->buildExtensionsPart($phpDocumentor->parser);
+            $markers    = $this->buildMarkersPart($phpDocumentor->parser);
+
+            $visibility         = ((string) $phpDocumentor->parser->visibility) ?: $visibility;
+            $defaultPackageName = ((string) $phpDocumentor->parser->{'default-package-name'}) ?: $defaultPackageName;
+            $template           = ((string) $phpDocumentor->transformations->template->attributes()->name) ?: $template;
```
mvriel: This is inside the isset for the `$phpDocumentor->parser`; meaning that if the user has specified a template but not a parser section that the template that he/she has selected is not used

src/phpDocumentor/ConfigurationFactory/PhpDocumentor2.php
```diff
@@ -0,0 +1,181 @@
...
+                if (isset($phpDocumentor->parser->files->{'ignore-symlinks'})) {
+                    $ignoreSymlinks = filter_var(
+                        $phpDocumentor->parser->files->{'ignore-symlinks'},
+                        FILTER_VALIDATE_BOOLEAN
+                    );
+                }
+            }
+        }
+
+        $outputDirectory = ((string) $phpDocumentor->parser->target) ?: 'file://build/docs';
```
mvriel: These two lines are outside the isset for the `$phpDocument->parser`; won't that give issues?

src/phpDocumentor/ConfigurationFactory/PhpDocumentor2.php
```diff
@@ -0,0 +1,181 @@
...
+        $directories     = ((array) $phpDocumentor->parser->files->directory) ?: ['src'];
```
mvriel: The default should not be 'src' but an empty folder (current working directory); this would confuse a use since some projects do not have an src folder

src/phpDocumentor/ConfigurationFactory/PhpDocumentor2.php
```diff
@@ -0,0 +1,181 @@
...
+                        'api'    => [
+                            'format'               => 'php',
+                            'source'               => [
+                                'dsn'   => 'file://.',
+                                'paths' => $sourcePaths,
+                            ],
+                            'ignore'               => [
+                                'hidden'   => $ignoreHidden,
+                                'symlinks' => $ignoreSymlinks,
+                                'paths'    => ['src/ServiceDefinitions.php'],
```
mvriel: This is not extracted from the XML file while it is present; also: the value `src/ServiceDefinitions.php` was an example and should not be used. The paths should contain a list of files and folders that may be ignored by the parser

src/phpDocumentor/ConfigurationFactory/PhpDocumentor2.php
mvriel: The config file for phpDocumentor 2 may feature multiple template tags to indicate that multiple templates should be executed in sequence; this only checks for one template reference

src/phpDocumentor/ConfigurationFactory/PhpDocumentor2.php
```diff
@@ -0,0 +1,181 @@
...
+     *
+     * @param \SimpleXMLElement $phpDocumentor
+     *
+     * @return array
+     */
+    public function convert(\SimpleXMLElement $phpDocumentor)
+    {
+        $this->validate($phpDocumentor);
+
+        $extensions         = [];
```
mvriel: the default for extensions is php, php3 and phtml; by omitting content here the parser won't pick up on any file

src/phpDocumentor/ConfigurationFactory/PhpDocumentor2.php
```diff
@@ -0,0 +1,181 @@
...
+        $markers            = [];
```
mvriel: the default for markers is 'TODO' and 'FIXME'; without this no markers will be interpreted

src/phpDocumentor/ConfigurationFactory/PhpDocumentor3.php
```diff
@@ -0,0 +1,197 @@
...
+
+    /**
+     * Initializes the PhpDocumentor3 strategy.
+     *
+     * @param string $schemaPath
+     */
+    public function __construct($schemaPath)
+    {
+        if ($schemaPath === '') {
+            $schemaPath = __DIR__ . '../../../../data/xsd/phpdoc.xsd';
```
mvriel: there should be a '/' at the start of the relative path string; without it it will concatenate the directory name with .. and there will be no slash in between

src/phpDocumentor/ConfigurationFactory/PhpDocumentor3.php
```diff
@@ -0,0 +1,197 @@
...
+                    break;
+                case 'template':
+                    $template = $this->buildTemplate($value);
+                    break;
+                default:
+                    break;
+            }
+        }
+
+        $outputDirectory = ((string) $phpDocumentor->parser->target) ?: 'file://build/docs';
```
mvriel: As can be seen in https://github.com/phpDocumentor/phpDocumentor2/issues/1522; the path in the XML is not $phpDocumentor->parser->target but $phpDocumentor->paths->output

src/phpDocumentor/ConfigurationFactory/PhpDocumentor3.php
```diff
@@ -0,0 +1,197 @@
...
+        $cacheDirectory  = ((string) $phpDocumentor->parser->cache) ?: '/tmp/phpdoc-doc-cache';
```
mvriel: As can be seen in https://github.com/phpDocumentor/phpDocumentor2/issues/1522; the path in the XML is not $phpDocumentor->parser->cache but $phpDocumentor->paths->cache

src/phpDocumentor/ConfigurationFactory/PhpDocumentor3.php
```diff
@@ -0,0 +1,197 @@
...
+    /**
+     * Checks if the provided configuration xml is for phpDocumentor3.
+     *
+     * @param $phpDocumentor
+     *
+     * @return bool
+     */
+    public function match(\SimpleXMLElement $phpDocumentor)
+    {
+        return (int) $phpDocumentor->attributes()->version === 3;
```
mvriel: Please do not convert it into an int but do a string comparison; if we want to introduce another parser for 3.1 then the adapter for 3.1 is not selected but this on instead (`(int)"3.1" === 3`)

src/phpDocumentor/ConfigurationFactory/PhpDocumentor3.php
```diff
@@ -0,0 +1,197 @@
...
+    {
+        $extensions = [];
+        foreach ($version->api->extensions->children() as $extension) {
+            if ((string) $extension !== '') {
+                $extensions[] = (string) $extension;
+            }
+        }
+
+        $ignoreHidden   = filter_var($version->api->ignore->attributes()->hidden, FILTER_VALIDATE_BOOLEAN);
+        $ignoreSymlinks = filter_var($version->api->ignore->attributes()->symlinks, FILTER_VALIDATE_BOOLEAN);
```
mvriel: ignore-symlinks may be removed from the entire phpDocumentor 3 spec; with the introduction of FlySystem we won't be supporting symlinks anymore. (you couldn't have known this; this came up during implementation of FlySystem)

src/phpDocumentor/ConfigurationFactory/PhpDocumentor3.php
```diff
@@ -0,0 +1,197 @@
...
+
+        return [
+            'folder' => (string) $version->folder,
+            'api'    => [
+                'format'               => ((string) $version->api->attributes()->format) ?: 'php',
+                'source'               => [
+                    'dsn'   => (string) $version->api->source->attributes()->dsn,
+                    'paths' => ((array) $version->api->source->path) ?: ['src'],
```
mvriel: please do not use 'src' as default but an empty entry; see the previous file for more information why

src/phpDocumentor/ConfigurationFactory/PhpDocumentor3.php
```diff
@@ -0,0 +1,197 @@
...
+
+        $ignoreHidden   = filter_var($version->api->ignore->attributes()->hidden, FILTER_VALIDATE_BOOLEAN);
+        $ignoreSymlinks = filter_var($version->api->ignore->attributes()->symlinks, FILTER_VALIDATE_BOOLEAN);
+
+        return [
+            'folder' => (string) $version->folder,
+            'api'    => [
+                'format'               => ((string) $version->api->attributes()->format) ?: 'php',
+                'source'               => [
+                    'dsn'   => (string) $version->api->source->attributes()->dsn,
```
mvriel: I am missing the default value here

src/phpDocumentor/ConfigurationFactory/PhpDocumentor3.php
```diff
@@ -0,0 +1,197 @@
...
+                    'paths' => ((array) $version->api->source->path) ?: ['src'],
+                ],
+                'ignore'               => [
+                    'hidden'   => $ignoreHidden,
+                    'symlinks' => $ignoreSymlinks,
+                    'paths'    => ((array) $version->api->ignore->path) ?: [],
+                ],
+                'extensions'           => $extensions,
+                'visibility'           => (string) $version->api->visibility,
```
mvriel: I think I am missing a default here

src/phpDocumentor/ConfigurationFactory/PhpDocumentor3.php
```diff
@@ -0,0 +1,197 @@
...
+                'default-package-name' => ((string) $version->api->{'default-package-name'}) ?: 'Default',
+                'markers'              => (array) $version->api->markers->children()->marker,
```
mvriel: I am missing a default here (in case the markers entry is omitted)

src/phpDocumentor/ConfigurationFactory/PhpDocumentor3.php
```diff
@@ -0,0 +1,197 @@
...
+                $extensions[] = (string) $extension;
+            }
+        }
+
+        $ignoreHidden   = filter_var($version->api->ignore->attributes()->hidden, FILTER_VALIDATE_BOOLEAN);
+        $ignoreSymlinks = filter_var($version->api->ignore->attributes()->symlinks, FILTER_VALIDATE_BOOLEAN);
+
+        return [
+            'folder' => (string) $version->folder,
+            'api'    => [
```
mvriel: A 'version' element can have zero or more 'api' elements; please omit this entry if no 'api' is provided and iterate over them if multiple are provided

src/phpDocumentor/ConfigurationFactory/PhpDocumentor3.php
```diff
@@ -0,0 +1,197 @@
...
+                    'hidden'   => $ignoreHidden,
+                    'symlinks' => $ignoreSymlinks,
+                    'paths'    => ((array) $version->api->ignore->path) ?: [],
+                ],
+                'extensions'           => $extensions,
+                'visibility'           => (string) $version->api->visibility,
+                'default-package-name' => ((string) $version->api->{'default-package-name'}) ?: 'Default',
+                'markers'              => (array) $version->api->markers->children()->marker,
+            ],
+            'guide'  => [
```
mvriel: A 'version' element can have zero or more 'guide' elements; please omit this entry if no 'guide' is provided and iterate over them if multiple are provided

src/phpDocumentor/ConfigurationFactory/PhpDocumentor3.php
```diff
@@ -0,0 +1,197 @@
...
+                'format' => ((string) $version->guide->attributes()->format) ?: 'rst',
+                'source' => [
+                    'dsn'   => ((string) $version->guide->source->attributes()->dsn) ?: 'file://.',
+                    'paths' => (array) $version->guide->source->path,
```
mvriel: I am missing a default (`['']`)

mvriel:
My apologies for the number of points; I noticed some things that would quite possibly lead to issues later on.

src/phpDocumentor/ConfigurationFactory/PhpDocumentor2.php
rgeraads: do you want me to change it one of the following?
- `[]`
- `['']`
- `['.']`
- `[cwd()]`

src/phpDocumentor/ConfigurationFactory/PhpDocumentor3.php
rgeraads: default is an empty array

src/phpDocumentor/ConfigurationFactory/PhpDocumentor3.php
rgeraads: api does not have a maxOccurs in the xsd. do you want me to add it as  maxOccurs="unbounded"?

src/phpDocumentor/ConfigurationFactory/PhpDocumentor3.php
rgeraads: guide does not have a maxOccurs in the xsd. do you want me to add it as maxOccurs="unbounded"?

src/phpDocumentor/ConfigurationFactory/PhpDocumentor3.php
rgeraads: visbility must be present and declared, that's why there is no default

PR #1603, opened 2015-07-02
Refactor RunCommand to split it into logical Commands

mvriel:


src/phpDocumentor/Application/Commands/CacheProject.php
```diff
@@ -0,0 +1,54 @@
...
+     *
+     * @param string $target
+     */
+    public function __construct($target)
+    {
+        $filesystem = new Filesystem();
+        if (! $filesystem->isAbsolutePath($target)) {
+            $target = getcwd() . DIRECTORY_SEPARATOR . $target;
+        }
+        if (! @file_exists($target) && ! @mkdir($target, 0777, true)) {
```
jaapio: is 0777 required here? I think you want to check if you could write to that directory and read from it.

mvriel: 0777 is the default of PHP (mind you, it is a mask and not a permission set) and required to set the third parameter to true (which is: recursive creation)

Issue #1339, opened 2014-07-10
Only one @package tag is allowed...  but only one package tag is present


rscarson:
Getting this for every single file since yesterday's update...

freshfelicio:
Got the same bug.

arunu:
Me too. Any fix available ?

olivierpons:
Still present on my debian/ phpdocumentor 2.8.2
Annoying to see same 64 errors on 64 files...
Any fix available yet?

markbijl:
Hi everyone, I am also still experiencing this issue, any outlook on when it might be fixed..?

rscarson:
Is this project abandoned or something? It's been a year, this is getting absurd

mvriel:
@rscarson no, the project is far from abandoned. We are currently focussing most efforts on creating the next version of phpDocumentor. This message does not break the generation of the documentation but is a nuisance that I have yet to discover the source of.

rscarson:
Fair enough! Sorry if I sounded rude :)

markbijl:
Good to hear it is still being worked on! We monitor our 'documentation issues' via Jenkins and generating a checkstyle report, so I was a bit shocked when we went from a few issues to a couple of thousand...

@rscarson Would be great if we can get rid of this issue :smile:

lsces:
Now that I've got control of things, this is the first error I need to clear. Every code base I'm working with is giving this error and obviously there is little point trying to tidy up until it has been fixed? I'm seeing other problems which are hopefully not a result of this, but @subpackage does not seem to be working as expected either.

tarcinil:
I just wanted add that I am noticing this as well, but only when I include @package on the page-level, if this is applied at the class-level, I don't notice an error.

Another strange behavior, If you place @package at the page level, it will get mapped to the correct package in generated documents, but if you have @package at the class-level, it will get mapped to the /Default package.

I am using 2.8.4.

NicolasCARPi:
:+1: for this issue :)

lsces:
Can anybody indicate a previous version that pre-dates the problem. I need to get some new documentation finished, but with several thousand files I don't want to have to start restyling everything.

rscarson:
Not to my knowledge. This is a program breaking bug the devs seem intent on ignoring.

I've switched to alternatives, and I'd recommend you do the same.

olivierpons:
May I ask you to give the alternatives ?

Etdashou:
[...]

PR #1613, opened 2015-07-28
Update FileAssembler.php

rscarson:
TO fix issue #1339


mvriel:
Unfortunately I am unable to merge this pull request because it is based on an older branch and updates code that is no longer in existence. But thanks to your analysis in #1339 I have been able to write a hotfix (that I dedicated to you in the commit message) and that is pushed and tagged. Thank you, this helped immensely

PR #1617, opened 2015-07-28
[WIP] Replace Transformer

mvriel:
In the new architecture we have replaced the Transformer for a TransformCommand that
will follow a different path to generate the documentation. This new architecture
should make the application more flexible and remove complexity from the codebase.


src/phpDocumentor/Renderer/Action/AppendFileHandler.php
```diff
@@ -0,0 +1,103 @@
...
+    {
+        $source      = $this->getSourceLocation($action);
+        $destination = $this->getDestination($action);
+
+        Assert::fileExists($source);
+        Assert::writable(dirname($destination));
+
+        $destinationFS = $action->getRenderPass()->getFilesystem();
+
+        if (is_file($source)) {
```
jaapio: shouldn't have been this FlySystem?

src/phpDocumentor/Renderer/Action/AppendFileHandler.php
```diff
@@ -0,0 +1,103 @@
...
+
+    /**
+     * @param AppendFile $action
+     *
+     * @return string
+     */
+    private function getSourceLocation(Action $action)
+    {
+        $source = (string)$action->getSource();
+        if (! $this->filesystem->isAbsolutePath($source)) {
```
jaapio: I don't understand this method. Maybe Puli could help us here?

mvriel: I have looked into Puli and didn't see any convincing arguments to use it; if anything it would make things more complex as I see it. Perhaps I am wrong; I haven't actually worked with Puli yet but I wasn't convinced for this use-case

PR #1626, opened 2015-08-09
Feature/api reference factory

jaapio:


src/phpDocumentor/ApiReference/DocumentGroupDefinition.php
```diff
@@ -0,0 +1,68 @@
...
+     *
+     * @return DocumentGroupFormat
+     */
+    public function getFormat()
+    {
+        return $this->format;
+    }
+
+    /**
+     * Returns the interal filesystem
```
mvriel: s/interal/internal

src/phpDocumentor/ApiReference/DocumentGroupDefinition.php
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    /**
+     * Returns an array of paths to files to process.
+     *
+     * @return string[]
+     */
+    public function getFiles()
+    {
+        return [];
```
mvriel: Is this a WIP thing? This class is final so that means that the getFiles will always return an empty array

src/phpDocumentor/ApiReference/Factory.php
```diff
@@ -0,0 +1,74 @@
...
+     * Creates Document group using the provided definition.
+     *
+     * @param DocumentGroupDefinitionInterface $definition
+     * @return DocumentGroup
+     */
+    public function create(DocumentGroupDefinitionInterface $definition)
+    {
+        if ($definition instanceof DocumentGroupDefinition) {
+
+            $projectFactory = new ProjectFactory(
```
mvriel: It would be better to inject the ProjectFactory in a constructor

src/phpDocumentor/ApiReference/Factory.php
```diff
@@ -0,0 +1,74 @@
...
+{
+    /**
+     * Creates Document group using the provided definition.
+     *
+     * @param DocumentGroupDefinitionInterface $definition
+     * @return DocumentGroup
+     */
+    public function create(DocumentGroupDefinitionInterface $definition)
+    {
+        if ($definition instanceof DocumentGroupDefinition) {
```
mvriel: Why do you conditionally match here; the 'match' method prescribes that this must be a DocumentGroupDefinition; so I would expect an exception if anything else is returned. Also: this if can re-use the match method; can it not?

mvriel:
Review completed; I had a few questions on specific topics. Also: I notice that the generic ApiReference Factory creates a specific PHP API object; is this something that should be extracted to a strategy?

src/phpDocumentor/ApiReference/Factory.php
jaapio: this was the point where I needed your opinion. Like you can see on line 42 a new strategy is created using the current handled definition. Since the strategy needs its file system. 

But since I can only addl strategies via the constructor this is a problem. We could leave it like this or adapt the Project factory to add extra strategies. via a method. The latter means that we also need a remove strategy method. because the factory could be reused in a second iteration and the file system strategy is only needed for this specific definition...

src/phpDocumentor/ApiReference/DocumentGroupDefinition.php
jaapio: Yes it is, I don't know yet how this will work. Most likely I will use the flyfinder plugin. But I didn't figure out yet how that works. And how it should be configured.

composer.json
```diff
@@ -32,7 +32,8 @@
         "league/flysystem": "~1.0",
         "php-di/php-di": "^5.0",
         "league/event": "^2.1",
-        "symfony/validator": "^2.7"
+        "symfony/validator": "^2.7",
+        "phpdocumentor/flyfinder": "dev-master@dev"
```
mvriel: I have just merged an update to flyfinder, when you merge develop into this PR: can you also update this dependency?

src/phpDocumentor/ApiReference/Factory.php
```diff
@@ -0,0 +1,77 @@
...
+     * @return DocumentGroup
+     */
+    public function create(DocumentGroupDefinitionInterface $definition)
+    {
+        /** @var DocumentGroupDefinition $definition */
+        if (!$this->matches($definition)) {
+            throw new InvalidArgumentException('Definition must be an instance of ' . DocumentGroupDefinition::class);
+        }
+
+        $projectFactory = new ProjectFactory(
```
mvriel: Can we inject this factory into this one? I would prefer it if every instantiation here is handled by the Dependency Injection Container. In the ContainerDefinitions file you can define the array with dependencies that you want to inject in the ProjectFactory's first argument

src/phpDocumentor/ApiReference/Factory.php
```diff
@@ -0,0 +1,77 @@
...
+                ),
+                new ProjectFactoryStrategy\Function_(),
+                new ProjectFactoryStrategy\Interface_(),
+                new ProjectFactoryStrategy\Method(),
+                new ProjectFactoryStrategy\Property(new PrettyPrinter()),
+                new ProjectFactoryStrategy\Trait_(),
+            ]
+        );
+
+        $project = $projectFactory->create('My Project', $definition->getFiles());
```
mvriel: This is currently specific to PHP; what we would want is that the format determines how data is handled. I would not recommend fixing that in this PR to limit the scope but fix this in a followup PR

src/phpDocumentor/FlyFinder/SpecificationFactory.php
```diff
@@ -0,0 +1,91 @@
...
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ *
+ * @copyright 2010-2015 Mike van Riel<mike@phpdoc.org>
+ * @license   http://www.opensource.org/licenses/mit-license.php MIT
+ * @link      http://phpdoc.org
+ */
+
+namespace phpDocumentor\FlyFinder;
```
mvriel: Perhaps we should move this to `phpDocumentor\Infrastructure\FlyFinder`; this is not part of our Domain Model

src/phpDocumentor/FlyFinder/SpecificationFactory.php
```diff
@@ -0,0 +1,91 @@
...
+
+        $ignoreSpec = null;
+        if (isset($ignore['paths'])) {
+            foreach ($ignore['paths'] as $path) {
+                $ignoreSpec = $this->orSpec($this->inPath($path), $ignoreSpec);
+            }
+        }
+
+        if (isset($ignore['hidden']) && $ignore['hidden'] === true) {
+            $ignoreSpec = $this->orSpec(new IsHidden(), $ignoreSpec);
```
mvriel: Shouldn't this be an andSpec?

src/phpDocumentor/FlyFinder/SpecificationFactory.php
```diff
@@ -0,0 +1,91 @@
...
+    {
+        $pathSpec = null;
+        foreach ($paths as $path) {
+            $pathSpec = $this->orSpec($this->inPath($path), $pathSpec);
+        }
+
+        $ignoreSpec = null;
+        if (isset($ignore['paths'])) {
+            foreach ($ignore['paths'] as $path) {
+                $ignoreSpec = $this->orSpec($this->inPath($path), $ignoreSpec);
```
mvriel: I am missing the inversion using notSpec; this statement says that the ignored paths should be included

mvriel: Also: shouldn't that then be andSpec?

src/phpDocumentor/FlySystemFactory.php
```diff
@@ -12,6 +12,7 @@
 
 namespace phpDocumentor;
```
mvriel: We want to move this to the `phpDocumentor\Infrastructure` namespace since this is not part of our domain model (the interface is; because that is specific in our domain but this concrete implementation is specific to Flysystem). I could understand if the distinction is not entirely clear, please indicate if that is the case then I will try toelaborate

src/phpDocumentor/Reflection/Php/Factory/File/FlySystemAdapter.php
```diff
@@ -0,0 +1,80 @@
...
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ *
+ * @copyright 2010-2015 Mike van Riel<mike@phpdoc.org>
+ * @license   http://www.opensource.org/licenses/mit-license.php MIT
+ * @link      http://phpdoc.org
+ */
+
+namespace phpDocumentor\Reflection\Php\Factory\File;
```
mvriel: We want to move this to the `phpDocumentor\Infrastructure\*` namespace since this is not part of our domain model (the interface is; because that is specific in our domain but this concrete implementation is specific to Flysystem). I could understand if the distinction is not entirely clear, please indicate if that is the case then I will try to elaborate

mvriel:
Just a few remarks about the location of classes but no visible bugs or anything: :+1:

jaapio:
@mvriel fixed a number of comments. Think this one can be merged. That will allow me to continue the integration.

PR #1629, opened 2015-08-16
Feature/api ref document group

jaapio:
This PR contains the factory and changes needed to create a Api Ref documentation Group Definition.


src/phpDocumentor/FlyFinder/SpecificationFactory.php
```diff
@@ -0,0 +1,91 @@
...
+{
+    /**
+     * Creates a SpecificationInterface object based on the ignore and extension parameters.
+     *
+     * @param array $paths
+     * @param array $ignore
+     * @param array $extensions
+     * @return SpecificationInterface
+     */
+    public function create(array $paths, array $ignore, array $extensions)
```
jaapio: @mbed67 could you have a look at this factory. To see if I used FlyFinder correctly

mbed67: @jaapio each specification has the ability to chain to another specification like so: 
return $pathSpec->andSpecification($extensions)->andSpecification($ignoreSpec->notSpecification());
This is because they all extend CompositeSpecification.

PR #1042, opened 2013-09-09
WIP: Adds partials to templates

siad007:
Related to #974

Can be invoked by:

```
<?xml version="1.0" encoding="UTF-8" ?>
<phpdocumentor>
    <partials>
        <partial name="introduction" href="docs/introduction.md"></partial>
        <partial name="anotherIntroduction" type="markdown">
            <content><![CDATA[
                # Another introduction
                ...
            ]]></content>
        </partial>
    </partials>
    <translator>
        <locale>de</locale>
    </translator>
</phpdocumentor>
```

Paths are always relative to the project.

PR #1642, opened 2015-08-25
a modified version of Clean template

web-elements:
I have created this modified version of Clean template for myself. Maybe could be useful to other.
Differences from Clean:
- Add index page w/ full packages / namespace tree and _maybe_ a partial "readme" (I followed #1042 but never worked, don't know if I have made some mistake or is a bug of phpDocumentor)
- Add Search for Constants, Functions, Traits, Interfaces, Classes and Methods
- Add an alternate navigation for packages
- Add full indexes for Constants, Functions, Traits, Interfaces and Classes
- Add the line number where a function or a class are in the file
- Fix source-view in namespace, now correctly show the code for all the functions
- Fix navigation menu, now autoselect where you are
- Fix version and since in classes, now show the number
- Fix the meta-title for the reports (deprecated, errors, markers)
- Update Font Awesome to 4.4.0


mvriel:
Hi @web-elements,

Thank you for showing us your modified version of Clean. We'll compare it to the existing Clean template and use all the good bits in honour of your efforts. It may take some time however as develop is currently not stable and will remain in that state for somewhat longer while we are refactoring for phpDocumentor3. But don't worry; I'll leave this PR open as a reminder so we won't forget.

Kind regards,

Mike

cbarina:
Is there some way to try this out in an install of phpdoc prior to being merged into the main codebase? I'm trying to resolve some issues I'm having with the clean template (in particular, missing references to files when documenting procedural files rather than class files) and wonder if a fix of the template could help.

jaapio:
@cbarina you could clone this repository. check out the latest tag and run `composer install` then you should be ready to go.

cbarina:
@jaapio Do you mean clone the latest phpdoc repo (v2.90)? I've been installing via the PEAR channel in the past, and it looks like the latest version there is 2.8.5. Would I need to first remove the PEAR-based install, then setup and use composer?

jaapio:
You can have them both. If you clone the repo and run composer install you will get an separate installation. Which you need to call directly since it will not be found by your os. 

If you need any help let me know.

cbarina:
@web-elements and @jaapio I finally managed to pull this and merge it into the composer-based install - it's pretty slick! I'm going to clean it up a bit more, then put in another pull request. I think the actual changed files are only in data/templates/clean-packages. Should I somehow put it here, or create a new pull request based on the latest develop branch, for ease?

jaapio:
I would recommend you to create a new pr based on master. I will back-port it to develop since develop contains a new phpdocumentor version.

mvriel:
Since we are currently in the process of migrating to trunk-based development (and thus want to delete develop), this PR's branch, unfortunately, diverges significantly from develop and we need to pick what we need I am going to close this PR for now.

PR #1659, opened 2015-10-16
fixes #1653

### LINKED ISSUE #1653: Phpdocumentor2 configuration strategy is missing defaults
In phpdoc2 a file phpdoc.tpl.xml was merged with the user's configuration. Either this step was missing in the new strategy or it doesn't work as expected.ld 

The strategy should not read this template file but it should contain the defaults. So those can be added when the strategy was executed.

rgeraads:


src/phpDocumentor/ConfigurationFactory/PhpDocumentor2.php
```diff
@@ -69,23 +51,23 @@ public function convert(\SimpleXMLElement $phpDocumentor)
...
                                 'dsn'   => 'file://.',
-                                'paths' => $sourcePaths,
+                                'paths' => $this->buildSourcePaths($phpDocumentor),
                             ],
                             'ignore'               => [
-                                'hidden'   => $ignoreHidden,
-                                'symlinks' => $ignoreSymlinks,
+                                'hidden'   => $this->buildIgnoreHidden($phpDocumentor),
+                                'symlinks' => $this->buildIgnoreSymlinks($phpDocumentor),
                                 'paths'    => ['src/ServiceDefinitions.php'],
```
jaapio: Why is this path here? looks like an unintended default?

src/phpDocumentor/ConfigurationFactory/PhpDocumentor2.php
```diff
@@ -26,40 +36,12 @@ public function convert(\SimpleXMLElement $phpDocumentor)
...
-            $sourcePaths[] = (string) (new Dsn($directory))->getPath();
-        }
+        $outputDirectory = $this->buildOutputDirectory($phpDocumentor);
 
         $phpdoc2Array = [
             'phpdocumentor' => [
                 'paths'     => [
-                    'output' => (string) (new Dsn($outputDirectory))->getPath(),
+                    'output' => (new Dsn($outputDirectory))->getPath(),
                     'cache'  => '/tmp/phpdoc-doc-cache',
```
jaapio: cant I configure the cache path?

rgeraads: as far as i know it is indeed not possible.

src/phpDocumentor/ConfigurationFactory/PhpDocumentor2.php
rgeraads: you seem correct, I updated the pr with added functionality for ignorePaths

PR #1621, opened 2015-08-05
Fixed @see tag links for classes #1620

jxmallett:
A fix for issue #1620. @see links in class docblocks now link correctly.

You no longer have to specify link text for @see tags in class or
method blocks.

Issue #1620, opened 2015-08-05
Links for @see tags in class DocBlocks don't go anywhere


jxmallett:
I've just started using phpDocumentor for the first time and it's going well.

Unfortunately, I've found that if i put a `@see` tag in a class docblock it doesn't get linked like other `@see` tags do. eg:

```
/**
 * @see SecondClass SecondClass // A link will be displayed but the href will be empty.
 */
class FirstClass {
    /**
     * @see SecondClass SecondClass // This will link to SecondClass, as expected.
     */
    public function MyFunction() {
    }
}

class SecondClass {
}
```

I've tried using fully resolved namespaces, like `\MyCompany\SecondClass` and also linking to methods. I've tried linking to everything I can think of but I can't get anything to work.

Perhaps related: @see links will only display if I specify link text. `@see SecondClass` won't display anything, regardless of where the tag is.

I'm using phpdoc 2.8.5 with the 'clean' template. I'm using PHP 5.5.27, on Ubuntu 14.04.

Is this a bug or am I doing something wrong? Any help would be much appreciated!

jxmallett:
Wow that was hard to debug! Tracked it down to a problem with the template though. I'll file a pull request soon.

mvriel:
Thank you for your pull request; the fix will become available with the next release of phpDocumentor or on the master branch of phpDocumentor

PR #1664, opened 2015-10-27
Fixing @see tag links in clean template

PhrozenByte:
This is a follow-up to #1620 and #1621 leaving some cases out


mvriel:
Thank you!

mvriel:
Also: I want to point out that you have issued this PR against develop, which will become the next major version of phpDocumentor (phpDocumentor 3). If your intention is to have this included in the next minor release of phpDocumentor 2, can you re-issue this PR against master?

Thanks in advance

PR #1668, opened 2015-10-30
Fixing getLatestVersion line feed

oswaldderiemaecker:
Fixing the version line feed for the feature "when I download the latest phar from github to phpDocumentor.phar"


mvriel:
Thanks!

PR #1671, opened 2015-11-02
restore functionality of template:list

mbed67:


src/phpDocumentor/Renderer/Template/PathsRepositoryInterface.php
```diff
@@ -0,0 +1,43 @@
...
+     */
+    public function findByTemplateAndPath(Template $template, Path $view);
+
+    /**
+     * Lists all available templates
+     *
+     * @return string[]
+     */
+    public function listTemplates();
+}
```
jaapio: missing a new line here

jaapio:
@mvriel, except for the missing newline  this looks good. Did you think about using flysystem to fetch templates? This allows us to use templates without actually downloading them?

@mbed67, this is looking good!

mvriel:
Awesome addition! Consider it merged.

Regarding Flysystem: that is an interesting option and something we could look into after v3 (to prevent scope creep)

PR #1673, opened 2015-11-02
Feature/api factory cache

jaapio:
This pr adds cache middleware to the project factory. Which allows partial processing


src/phpDocumentor/ContainerDefinitions.php
```diff
@@ -160,6 +171,27 @@ class_alias(Extension::class, 'phpDocumentor\\Plugin\\Core\\Xslt\\Extension');
     DocumentationFactory::class => \DI\object()
         ->method('addDocumentGroupFactory', \DI\get(phpDocumentor\ApiReference\Factory::class)),
 
+    //ApiReference
+    phpDocumentor\ApiReference\Factory::class => function (ContainerInterface $c) {
+        $strategies = [
+            new Argument(new PrettyPrinter()),
```
mvriel: An improvement here could be to retrieve the service from the DIC using `\DI\get(Argument::class)` so that it will autoresolve the dependency and ensure that it can be defined in the DIC

mvriel:
This item looks good!

src/phpDocumentor/ApiReference/Factory.php
```diff
@@ -16,21 +16,36 @@
...
+    /**
+     * @var array
+     */
+    private $middleware;
+
+    /**
+     * @param ProjectFactoryStrategy[] $stategies
+     * @param Middleware[] $cache
+     */
+    public function __construct(Emitter $emitter, array $stategies = array(), array $middleware = array())
```
mvriel: typo: `stategies` should be `strategies` (also in other places in this file)

mvriel: Please use short array notation instead of `array()`

src/phpDocumentor/ApiReference/Factory.php
```diff
@@ -47,25 +62,15 @@ public function create(DocumentGroupDefinitionInterface $definition)
             throw new InvalidArgumentException('Definition must be an instance of ' . DocumentGroupDefinition::class);
         }
 
-        // TODO: move this to a custom factory so that we can inject the factory in this class and then have the
-        // middleware as dependency of that factory.
+        $strategies = $this->stategies;
+        $strategies[] = new File(
```
mvriel: I would prefer that the File factory doesn't instantiate the necessary adapters at constructor time but when a `get` or `create` method is invoked (which then accepts a FlySystemAdapter or destination uri). That way we can move the instantiation of these objects out of  this class into the DIC.

src/phpDocumentor/ApiReference/Factory.php
mvriel: shouldn't it also be `$middlewares`? (plural)

src/phpDocumentor/ApiReference/Factory.php
mvriel: Using such a factory will also remove the need for injecting the middleware into this class; and that will improve the SRP part of this class

mvriel: It also makes testing easier; now the behaviour is untested because it is hard to test the hardcoded strategy and middlewares without 'polluting' the test

src/phpDocumentor/Reflection/Php/Factory/File/CacheMiddleware.php
```diff
@@ -0,0 +1,90 @@
...
+    /**
+     * Executes this middle ware class.
+     * A middle ware class MUST return a File object or call the $next callable.
+     *
+     * @param CreateCommand $command
+     * @param callable $next
+     *
+     * @return File
+     */
+    public function execute($command, callable $next)
```
mvriel: the `--force` command line parameter enforces that the cache is always overwritten; how can we reintroduce that here?

mvriel: The cache path is also configurable from the command line; can we re-introduce that here as well?

mvriel:
This seems to be an excellent starting point; I do wonder whether we should have the middleware invoke a command and then have the command perform the caching. As this is a business logic affair it would seem logical but I am interested in your opinion.

src/phpDocumentor/ApiReference/Factory.php
jaapio: I would like tho implement this. But I will have to change the ProjectFactory class for this. It should be able to expose it's strategies. (Which I don't like). Or it should accept a FileStrategyAdapter on it's create method. Which I don't like either. Since the FileStrategyAdapter is very specific to the FileStrategy. 

The actual problem lies in the requirement of the strategy to get the filesystem from the definition. Which is only available when the Factory::create method is called....

src/phpDocumentor/Reflection/Php/Factory/File/CacheMiddleware.php
jaapio: If we can inject the commandline parameters some way here that should be possible.

jaapio:
I updated this pr, major changes are done by removing the middleware and strategies dependencies from the Factory. This was posible due the introduction of File interface.

src/phpDocumentor/Reflection/File/FlySystemFile.php
```diff
@@ -0,0 +1,72 @@
...
+ * Wrapper for FlySystem's FilesystemInterface.
+ */
+final class FlySystemFile implements File
+{
+    /**
+     * @var FilesystemInterface
+     */
+    private $fileSystem;
+    /**
+     * @var
```
mvriel: missing type

src/phpDocumentor/Reflection/File/FlySystemFile.php
```diff
@@ -0,0 +1,72 @@
...
+     */
+    private $fileName;
+
+    /**
+     * FlySystemFile constructor.
+     * @param FilesystemInterface $fileSystem
+     * @param $fileName
```
mvriel: Missing type

tests/unit/phpDocumentor/Reflection/Middleware/LoggingMiddlewareTest.php
```diff
@@ -0,0 +1,38 @@
...
+namespace phpDocumentor\Reflection\Middleware;
+
+use League\Event\Emitter;
+use Mockery as m;
+use phpDocumentor\Reflection\File\LocalFile;
+use phpDocumentor\Reflection\Php\Factory\File\CreateCommand;
+use phpDocumentor\Reflection\Php\ProjectFactoryStrategies;
+use PHPUnit_Framework_TestCase;
+
+class LoggingMiddlewareTest extends PHPUnit_Framework_TestCase
```
mvriel: Missing `@coversDefaultClass`

tests/unit/phpDocumentor/Reflection/Middleware/LoggingMiddlewareTest.php
```diff
@@ -0,0 +1,38 @@
...
+{
+    public function testExecute()
```
mvriel: missing `@covers`

tests/unit/phpDocumentor/Reflection/Php/Factory/File/CacheMiddlewareTest.php
```diff
@@ -0,0 +1,95 @@
...
+
+namespace phpDocumentor\Reflection\Php\Factory\File;
+
+use Mockery as m;
+use phpDocumentor\Reflection\File as SourceFile;
+use phpDocumentor\Reflection\Php\File;
+use phpDocumentor\Reflection\Php\StrategyContainer;
+use Stash\Pool;
+
+class CacheMiddlewareTest extends \PHPUnit_Framework_TestCase
```
mvriel: Missing `@coversDefaultClass`

tests/unit/phpDocumentor/Reflection/Php/Factory/File/CacheMiddlewareTest.php
```diff
@@ -0,0 +1,95 @@
...
+{
+    public function testCacheIsUsed()
```
mvriel: missing `@covers`

tests/unit/phpDocumentor/Reflection/Php/Factory/File/CacheMiddlewareTest.php
```diff
@@ -0,0 +1,95 @@
...
+        $fixture = new CacheMiddleware($poolMock);
+
+        $result = $fixture->execute($command, function () use ($file) {
+            return $file;
+        });
+
+        $this->assertSame($file, $result);
+    }
+
+    public function testChecksHash()
```
mvriel: missing `@covers`

src/phpDocumentor/Reflection/Php/Factory/File/CacheMiddleware.php
```diff
@@ -0,0 +1,90 @@
...
+use phpDocumentor\Reflection\Php\File;
+use Stash\Item;
+use Stash\Pool;
+
+final class CacheMiddleware implements Middleware
+{
+    /**
+     * Cache namespace used for this repository.
+     */
+    const CACHE_NAMESPACE = 'Documentation\\Api\\Php';
```
mvriel: How will this work with multiple versions? Will versions mix their caches? (perhaps something to address in a next PR)

tests/unit/phpDocumentor/Reflection/Middleware/LoggingMiddlewareTest.php
mvriel: Also missing `final`

tests/unit/phpDocumentor/Reflection/Php/Factory/File/CacheMiddlewareTest.php
mvriel: also missing `final`

mvriel:
Only very minor style things were found during review and one possible clash in the caching key which should be addressed in a subsequent PR if this is indeed a problem.

Aside from that :+1:

Issue #1677, opened 2015-11-21
Wrong param name in dockblock does not trig an error (and a solution)


marcimat:
In phpdocumentor 2.8.5, no error is send if a `@param` tag has a wrong name like in this example : 

    /** Summary
     * @param string $wrong_name
     */
    function demo($other_name) {}

However, a Validator is there to do that (DoesArgumentNameMatchParamValidator.php), but a test return always true : `empty($value->parameter)` [line 36](https://github.com/phpDocumentor/phpDocumentor2/blob/master/src/phpDocumentor/Plugin/Core/Descriptor/Validator/Constraints/Functions/DoesArgumentNameMatchParamValidator.php#L36). 

In fact, the class [ValidationValueObject](https://github.com/phpDocumentor/phpDocumentor2/blob/master/src/phpDocumentor/Plugin/Core/Descriptor/Validator/ValidationValueObject.php) need an `__isset()` method like : 

    public function __isset($name)
    {
        return array_key_exists($name, $this->data);
    }

marcimat:
Also, an additional parameter does not throw the notice, for example in : 

    /** Summary
     * @param string $not_there
     */
    function demo() {}

The validator is there (DoesParamsExistsValidator.php), but a test return always false : `is_array($parameters)` [line 39](https://github.com/phpDocumentor/phpDocumentor2/blob/master/src/phpDocumentor/Plugin/Core/Descriptor/Validator/Constraints/Functions/DoesParamsExistsValidator.php#L39). This `$parameters` is a Collection, witch interface `\ArrayAccess`. 

Possible solution is to replace the line by :

    if (count($arguments) > 0 && $parameters instanceof \ArrayAccess) {

marcimat:
Thanks :)

PR #1678, opened 2015-11-21
Fix for #1677 : correct validation tests on `@param` tags

marcimat:
As seen in #1677, some errors and notices wasn't found for the `@param` tag in dockblock.


mvriel:
:+1:

Issue #1599, opened 2015-06-25
Documentation Standards in phpDocumentor 3


mvriel:
> _Important:_ this item should be developed against the develop branch and not the master branch; the develop branch contains all parts needed for phpDocumentor 3 where master does not.

The basic idea behind this feature is that we want to be able to check a codebase that has been parsed by the phpDocumentor Reflection package against a Documentation Standard, and receive a list of violations of that Standard so that developers can ensure consistency in their project and make sure their documentation is in order.

The Reflection package of phpDocumentor will analyse the source code of an entire project and create an object graph representing all elements of that project including their associated Documentation (using DocBlocks).

This object graph starts with a phpDocumentor\Reflection\Project object, which in turn consists of a series of objects representing Files and in those files is a tree of elements in those files.

This feature should accept any element in that tree (project, file, class, etc) and recursively check those elements against a Documentation Standard. A Documentation Standard will contains a series of rules and settings for those rules; if an element is applicable for a rule, such as a check on `@return` on a Method, but fails the rules business logic then a Violation should be emitted.

After one or more elements are checked should we have a list of violations of which we know at least:
1. Which element the violation happened on, identified by their FQSEN
2. What the severity of the violation is
3. A code identifying the rule
4. A (configurable) message 
## Acceptance Criteria
- [ ] It should be possible to define Documentation Standards that consist of a series of preconfigured rules
- [ ] It should be possible to extend or import Documentation Standards in a Documentation Standard and disable one or more rules, or adjust one or more parameters of rules
- [ ] Each rule should represent a single violation of the Documentation Standards, see below for an initial list of rules
- [ ] Every rule may have a severity associated with it, this represents if the rule is an error, warning or notice. 
- [ ] The default severity of a Rule may be overridden
- [ ] Every rule has a default message associated with it
- [ ] The default message of a rule may be overridden
- [ ] A rule can be limited to a specific set of Elements. For example: checks on the `@return` tag are only allowed for functions and methods
- [ ] After the verification is completed a list of violations should be returned with at least 
- [ ] Test covered
## Technical Guidelines
- The description of this item states that we want to check Reflected information; this is true for this specific item but we will want to re-use this component to verify all documentation interpreted by phpDocumentor; this might be a design consideration.
[...]

PR #1680, opened 2015-11-30
[WIP] Feature/validation

jaapio:
Validation setup.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -680,18 +680,18 @@
             "source": {
                 "type": "git",
                 "url": "https://github.com/phpDocumentor/ReflectionDocBlock.git",
-                "reference": "1f637ffba2af2867bda3b4f682487442f190ea14"
+                "reference": "82f9aba5a508878923f6a2f88dfb698ec26bbb87"
             },
             "dist": {
                 "type": "zip",
                 "url": "https://api.github.com/repos/phpDocumentor/ReflectionDocBlock/zipball/82f9aba5a508878923f6a2f88dfb698ec26bbb87",
-                "reference": "1f637ffba2af2867bda3b4f682487442f190ea14",
+                "reference": "82f9aba5a508878923f6a2f88dfb698ec26bbb87",
                 "shasum": ""
             },
             "require": {
                 "php": ">=5.5",
                 "phpdocumentor/reflection-common": "^1.0@dev",
-                "phpdocumentor/type-resolver": "^0.1.2",
+                "phpdocumentor/type-resolver": "^0.1.5",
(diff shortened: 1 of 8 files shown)


jaapio:
@mvriel I create this pr as a discussion about the implementation of #1599.
Since I based the code on an other pr you didn't merge yet it is a bit messy. But all interesting code is in commit  9543579.

I suggest that we introduce an interface for file in phpdocumentor/Reflection. So we can create a decorator like plugin structure. Each middleware can add it's own result to the file object as a wrapper around this file. Which allows other classes to process the results. E.g. a view can get all messages from a file and render those. Same would apply to a console logger which logs the validation messages to the console.

A down side is that the classes using the specialized decorators will have to check if the specialized methods are available. Since this might change when middlewares are disabled.

I would like to hear what you are thinking about this setup.

mvriel:
@jaapio I foresee maintenance issues with using specialized versions of the File class since you would need to specifically check in a view which sort of File you have. This also only works when you have a single decorator because nesting this trick several times breaks aforementioned checks.

I suggest creating a new Collection object (i.e. `Violations`) that is instantiated in the DIC and provided to the validator. This Violations collected can then be populated with the discovered validation violations and a view that wants to read the violations can then have the Violations Collection object injected and use that to process all issues.

This have a few advantages:

- The File Value Object is consistent and thus every class using this Value Object knows what it can expect
- We can use the Violations collection also to record violations in other files, such as the menu.xml Menu Definition (Markdown parsing) or Markdown/ReST files. 
- There is less coupling between the Reflection library and the Validation library

jaapio:
Lets see how that will work out. It feels a bit strange to inject an object and modify that with other classes.  It would have been nice if we could make the violations collection immutable. But at the moment I don't have a better solution. I think this is more or less the issue of doing these kind of things in middlewares. 

Even when we move the validation to a step after the parsing process the voilations have to be stored in the DI-container. Otherwise the view doesn't have access to it. Or we should find a way to add those violations to the project. But that will result in the same issues as you are seeing with the file class.

mvriel:
An alternative may be to have a ViolationsRepository that is responsible for persisting and retrieving violations (for example in cache or memory). That may be a more stable approach as that is an immutable object?

jaapio:
I will try the approach with the collection first. Basicly that will do the same as a repository would do. Using the cache for temporary storage doesn't feel right either. Cache should be disposable at any time, from my point of view. In this case that will cause a datalost.

mvriel:
Closing this PR as it is superseded by https://github.com/phpDocumentor/phpDocumentor2/pull/1721

Issue #1589, opened 2015-06-04
Deprecated Symfony validator


DoBi:
When running phpdoc I get a lot of ```PHP Deprecated``` messages. 

For example for ```phpdoc --version```:
```
PHP Deprecated:  The Symfony\Component\Validator\Validator class is deprecated since version 2.5 and will be removed in 3.0. Use the Symfony\Component\Validator\Validator\RecursiveValidator class instead. in /home/dominik/Entwicklung/dobi/core/router/vendor/symfony/validator/Validator.php on line 14
PHP Deprecated:  The Symfony\Component\Validator\Mapping\ClassMetadataFactory class is deprecated since version 2.5 and will be removed in 3.0. Use the Symfony\Component\Validator\Mapping\Factory\LazyLoadingMetadataFactory class instead. in /home/dominik/Entwicklung/dobi/core/router/vendor/symfony/validator/Mapping/ClassMetadataFactory.php on line 14
PHP Deprecated:  The class Symfony\Component\Validator\DefaultTranslator is deprecated since version 2.7 and will be removed in 3.0. Use Symfony\Component\Translation\IdentityTranslator instead. in /home/dominik/Entwicklung/dobi/core/router/vendor/symfony/validator/DefaultTranslator.php on line 14
phpDocumentor version 2.8.2
```

The functionality is not restricted. I only want to report this warnings.
This must be considered for the future.

skoop:
I've added a fix for these messages in https://github.com/phpDocumentor/phpDocumentor2/pull/1690

PR #1690, opened 2015-12-04
#1589: Replace deprecated Symfony2 classes with the new counterparts

skoop:
These changes should fix the deprecated messages for Symfony components.


jaapio:
tnx @skoop!

Issue #1598, opened 2015-06-23
date.timezone settings not valid for PHP 5.6


bgaillard:
Hi, I just upgraded the PHP version in use on our CI server and encounter the following warnings. 

```
string(134) "/home/ec2-user/.jenkins/jobs/fatcap-php-services-testing/workspace/target/reports/phpdocumentor/files/FatCap.Model.IVisualContent.html"
PHP Warning:  date_default_timezone_get(): It is not safe to rely on the system's timezone settings. You are *required* to use the date.timezone setting or the date_default_timezone_set() function. In case you used any of those methods and you are still getting this warning, you most likely misspelled the timezone identifier. We selected the timezone 'UTC' for now, but please set date.timezone to select your timezone. in /usr/share/pear/phpDocumentor/vendor/twig/twig/lib/Twig/Extension/Core.php on line 89
PHP Stack trace:
PHP   1. {main}() /usr/bin/phpdoc:0
PHP   2. phpDocumentor\Application->run() /usr/bin/phpdoc:24
PHP   3. Symfony\Component\Console\Application->run() /usr/share/pear/phpDocumentor/src/phpDocumentor/Application.php:183
PHP   4. Symfony\Component\Console\Application->doRun() /usr/share/pear/phpDocumentor/vendor/symfony/console/Symfony/Component/Console/Application.php:126
PHP   5. Symfony\Component\Console\Application->doRunCommand() /usr/share/pear/phpDocumentor/vendor/symfony/console/Symfony/Component/Console/Application.php:195
PHP   6. Symfony\Component\Console\Command\Command->run() /usr/share/pear/phpDocumentor/vendor/symfony/console/Symfony/Component/Console/Application.php:874
PHP   7. phpDocumentor\Command\Project\RunCommand->execute() /usr/share/pear/phpDocumentor/vendor/symfony/console/Symfony/Component/Console/Command/Command.php:253
PHP   8. Symfony\Component\Console\Command\Command->run() /usr/share/pear/phpDocumentor/src/phpDocumentor/Command/Project/RunCommand.php:275
PHP   9. phpDocumentor\Transformer\Command\Project\TransformCommand->execute() /usr/share/pear/phpDocumentor/vendor/symfony/console/Symfony/Component/Console/Command/Command.php:253
[...]

PR #1694, opened 2015-12-10
#1598 Reduced specificity of if statement

CosetteN:
Line 98, Per comment thread on issue, changed 
    "if (false === ini_get('date.timezone')" 
to 
    "if (false == ini_get('date.timezone')"

First pull request on github.  Thank you for linking through up-for-grabs.net website!

### DIFF
--- a/src/phpDocumentor/Application.php
+++ b/src/phpDocumentor/Application.php
@@ -95,7 +95,7 @@ final class Application
      */
     private function setTimezone()
     {
-        if (false === ini_get('date.timezone')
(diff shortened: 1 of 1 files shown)


jaapio:
Thanks for your contribution. I will apply this manually to the master branch, since you created a patch for develop, which contains the unreleased 3.0 version of phpdocumentor.
#1598

CosetteN:
It really was too much to hope a my first dive in would be completely
right, wasn't it?  thank you for being understanding.  Have a great weekend!

PR #1697, opened 2015-12-12
Hotfix/fix tests

jaapio:
I solved 2 issues in unittests thtat where not fixed during a refactor.
And checked the composer dependencies for php 5.3.3. Since the lock file is only used when installing phpdocumentor as a solo project this will not give any errors when people use composer to install phpdocumentor as dependency


jaapio:
@mvriel there are so many behat tests failing that I would like to suggest that we disable them again.

PR #1711, opened 2015-12-27
Adds command to reconfigure cache pool.

jaapio:
This pr as a new command to the run-cli-command which reconfigures the cache pool.
Reintroduce usage of --force to disable cache usage
cache can be disabled from config by setting "use-cache" to false

### DIFF
--- a/src/phpDocumentor/Application/Commands/ConfigureCacheHandler.php
+++ b/src/phpDocumentor/Application/Commands/ConfigureCacheHandler.php
@@ -1,0 +1,52 @@
+<?php
+/**
+ * This file is part of phpDocumentor.
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ *
+ * @copyright 2010-2015 Mike van Riel<mike@phpdoc.org>
+ * @license   http://www.opensource.org/licenses/mit-license.php MIT
+ * @link      http://phpdoc.org
+ */
+
+namespace phpDocumentor\Application\Commands;
+
+use phpDocumentor\Application\Configuration\ConfigurationFactory;
+use Stash\Pool;
+
+final class ConfigureCacheHandler
+{
+    /**
+     * @var Pool
+     */
+    private $pool;
+
+    /**
+     * @var ConfigurationFactory
+     */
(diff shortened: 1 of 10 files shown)


tests/unit/phpDocumentor/Application/Commands/ConfigureCacheHandlerTest.php
```diff
@@ -0,0 +1,99 @@
...
+
+    /**
+     * @covers ::__invoke
+     */
+    public function testLeaveCache()
+    {
+        $this->invokeTest(true);
+    }
+
+    private function invokeTest($useCache)
```
mvriel: Wouldn't it be better to make this the test and have a data provider insert the true and false?

src/phpDocumentor/Application/Commands/ConfigureCacheHandler.php
```diff
@@ -0,0 +1,52 @@
...
+     * @var ConfigurationFactory
+     */
+    private $configurationFactory;
+
+    /**
+     * ConfigureCacheHandler constructor.
+     */
+    public function __construct(Pool $pool, ConfigurationFactory $configurationFactory)
+    {
+
```
mvriel: Redundant whiteline

tests/unit/phpDocumentor/Application/Commands/ConfigureCacheHandlerTest.php
```diff
@@ -0,0 +1,99 @@
...
+     * @var ConfigureCacheHandler
+     */
+    private $fixture;
+
+    /**
+     * @var m\MockInterface
+     */
+    private $configStrategy;
+
+    /**
```
mvriel: Empty docblock

tests/unit/phpDocumentor/Application/Commands/ConfigureCacheHandlerTest.php
jaapio: I tried that, but for some reason it didn't work.
On the other side this is clearer since the test name describes what i expect.

mvriel:
A few style things but I will fix those now; nothing that stops the merge

PR #1721, opened 2016-01-18
Adds validating middleware to file construction.

jaapio:
This pr adds an other concept of validation without using an external validator library.
This way we don't have to convert your object stucture to an array. Which makes it easier
to validate new kind of rules.

The middleware implementation is just a way how things could work. Validation can easely added
as a post parse step using a command in the Cli run command class.


jaapio:
@mvriel I think this implementation is better than the implementation in #1680. It doesn't require the conversion to an array, which saves memory. And it doesn't require an external library.

mvriel:
I like this, I like this a lot :)

I can see how we can also extend this with XML based rulesets where you can mix and match your own rules. The only comment I currently have is the missing tests but this seems like a solid architecture. I like how you added the event to the middleware so that concerns stay separate

mvriel:
@jaapio can we extract the changes you made from this PR into a new PR? Or do you think we can better close this PR and open a new one as soon as we implement this?

jaapio:
Think we both agree on this approach. But it will not fit in the current implementation. We can close this. And we will reimplement it when we are at the point where we were when I implemented this.

PR #1732, opened 2016-02-02
add test locations to PHPCS config file

ppaulis:
Hey @mvriel,

we made a few upgrades on our PHPCS activity. Instead of specifying the test locations in the continuousphp pipeline, you can now simply add it to the phpcs.xml file. To not break any future builds, the old way is still supported.
If you decide to accept this PR, simply replace the 3 Paths in the pipeline config by one single path "phpcs.xml".

Check out my latest build with the new PHPCS config :)
https://continuousphp.com/git-hub/ppaulis/phpDocumentor2/build/0b2a996f-f3a2-4854-9f9c-c1f3e7913be4/output

### DIFF
--- a/phpcs.xml
+++ b/phpcs.xml
@@ -1,8 +1,8 @@
 <?xml version="1.0"?>
-<ruleset name="continuousphp-phing-tasks">
+<ruleset name="phpDocumentor">
     <file>src</file>
-    <file>tests/unit</file>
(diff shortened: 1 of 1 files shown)


mvriel:
Hey @ppaulis,

Thanks for this! I am now upgrading our build process and this is very convenient (didn't know it existed).

PR #1740, opened 2016-02-20
Refactored the CopyFile Action

mvriel:
I have changed the CopyFile Action to make use of the new Artifact and
Asset abstractions; I had to make a little bit of extra effort to ensure
that it is compatible with previous editions of phpDocumentor. The CopyFile
Action will do a recursive copy of the files reported by the Assets
abstraction.

This also meant that I have expanded the Assets abstraction to include the
concept of a Folder; which means that you will return a collection of Path
objects that you can subsequently retrieve using the Assets abstraction.

I have explicitly chosen to return Path objects and not complete Asset
objects because that might grow a bit large (being a recursive action
through the folder and all subfolders) and because that may cause I/O that
is never used because the user may not want to retrieve all assets. In
addition, memory would increase significantly because all files would be
stored in memory instead of retrieved on a per Asset basis.


mvriel:
Issued against wrong branch ..

PR #1741, opened 2016-02-20
Refactor copy action to v3

mvriel:
I have changed the CopyFile Action to make use of the new Artifact and
Asset abstractions; I had to make a little bit of extra effort to ensure
that it is compatible with previous editions of phpDocumentor. The
CopyFile Action will do a recursive copy of the files reported by
the Assets abstraction.

This also meant that I have expanded the Assets abstraction to include
the concept of a Folder; which means that you will return a collection
of Path objects that you can subsequently retrieve using the Assets
abstraction.

I have explicitly chosen to return Path objects and not complete
Asset objects because that might grow a bit large (being a recursive
action through the folder and all subfolders) and because that may
cause I/O that is never used because the user may not want to
retrieve all assets. In addition, memory would increase significantly
because all files would be stored in memory instead of retrieved
on a per Asset basis.

### DIFF
--- a/src/DomainModel/Renderer/Asset/Folder.php
+++ b/src/DomainModel/Renderer/Asset/Folder.php
@@ -1,0 +1,55 @@
+<?php
+/**
+ * This file is part of phpDocumentor.
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ *
+ * @copyright 2010-2016 Mike van Riel<mike@phpdoc.org>
+ * @license   http://www.opensource.org/licenses/mit-license.php MIT
+ * @link      http://phpdoc.org
+ */
+
+namespace phpDocumentor\DomainModel\Renderer\Asset;
+
+use phpDocumentor\DomainModel\Path;
+use Webmozart\Assert\Assert;
+
+class Folder extends \ArrayObject
+{
+    /**
+     * @var Path
+     */
+    private $path;
+
+    /**
+     * @param Path $path
+     * @param Path[] $locations
+     */
+    public function __construct(Path $path, array $locations)
+    {
+        Assert::allIsInstanceOf($locations, Path::class);
+
+        $this->path = $path;
+        parent::__construct($locations);
+    }
+
+    /**
+     * @return Path
+     */
+    public function path()
+    {
+        return $this->path;
+    }
+
+    /**
+     * @param int|string $index
+     * @param Path $newval
+     */
+    public function offsetSet($index, $newval)
+    {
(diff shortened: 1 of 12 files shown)


src/DomainModel/Renderer/Asset/Folder.php
```diff
@@ -0,0 +1,55 @@
...
+ * @license   http://www.opensource.org/licenses/mit-license.php MIT
+ * @link      http://phpdoc.org
+ */
+
+namespace phpDocumentor\DomainModel\Renderer\Asset;
+
+use phpDocumentor\DomainModel\Path;
+use Webmozart\Assert\Assert;
+
+class Folder extends \ArrayObject
```
jaapio: Missing use statement

src/Infrastructure/Renderer/FlySystemAssets.php
```diff
@@ -42,7 +44,25 @@ public function get(Path $location)
...
         }
 
-        return new Asset($this->filesystem->get((string)$location));
+        /** @var File $flyLocation */
+        $flyLocation = $this->filesystem->get((string)$location);
+        if ($flyLocation->isDir()) {
+            $listing = $this->filesystem->listContents((string)$location, true);
+            $locations = array_map(
+                function ($value) {
+                    if ($value['type'] === 'dir') {
```
jaapio: A folder cannot have any child folders?

src/Infrastructure/Renderer/FlySystemAssets.php
```diff
@@ -42,7 +44,25 @@ public function get(Path $location)
...
+
+                    return new Path($value['path']);
+                },
+                $listing
+            );
+
+            return new Asset\Folder($location, array_values(array_filter($locations)));
+        }
+
+        return new Asset($flyLocation->read());
```
jaapio: Is this a potential memory issue? You are reading the full file content here

jaapio:
Instead of creating an asset object for folders and assets we could consider to just create an interface for both. And create an implementation in the infrastructure layer that proxies to flysystem. I can imaging that you want to do this in a future pr.

src/DomainModel/Renderer/Assets.php
```diff
@@ -38,13 +38,14 @@ public function includeTemplateAssets(Template $template);
     public function has(Path $location);
 
     /**
-     * Retrieves the asset with the given name.
+     * Retrieves an array with the requested asset or if the name matches multiple assets (such as with a folder)
+     * then it returns all of them.
      *
```
mbed67: Doesn't it currently return either an Asset (not array of Assets) or a Folder?

PR #1742, opened 2016-02-20
Improve QA pipeline

mvriel:
In this Pull Request I am working on improving our QA and Build pipeline; in order to improve our process I have made the following changes:

- Removed old tests that were not executed anyway and where the components will be refactored
- Marked a few folders to be excluded from code coverage because they will be refactored
- Added a test scenario to a test
- Removed various unused packages from the composer.json
- Removed phpunit and phpcs from the composer.json as dependencies
- Added scripts to the composer.json that will download the phar files for Box, PHPUnit and PHPCS
- Changed the phpunit and phpcs binary in the bin folder to use these phar files
- Added custom scripts to the composer.json so that you can easily run tests using the following commands:
  - `composer test` - for unit tests using phpunit
  - `composer coverage` - for unit tests using phpunit, and generating code coverage in `build/coverage`
  - `composer style` - for checking the coding standards using phpcs


bin/phpdoc.php
```diff
@@ -1,14 +0,0 @@
-#!/usr/bin/env php
```
mvriel: `./bin/phpdoc.php` has been deprecated in favor of `./bin/phpdoc` since forever; I have now removed it

bin/phpunit
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
```
mvriel: We now install phpunit as a phar file in the build directory as part of a composer install; this re-introduces the `./bin/phpunit` binary

bin/phpcs
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
```
mvriel: We now install phpcs as a phar file in the build directory as part of a composer install; this re-introduces the `./bin/phpcs` binary

jaapio:
Looks good, but wondering why you removed all dev applications like phpunit & php code sniffer. But leave behat as dependency?

mvriel:
@jaapio because I didn't convert that one (yet) as it doesn't play a big role (yet) :)

PR #1751, opened 2016-03-16
[WIP] First work on setting up reducers for the project mapper

mbed67:


jaapio:
I think this looks great in general. Using the interpreter and interpret classes from the reflection component doesn't feel right. We might consider to move this to a separate package?

I would try to split the file reducer in to more small reducers. The initial state could contain all file properties. But a new reducer in the same chain would add classes, functions, constants to this state. This will make it easier to test how all parts are handled.  Reducers should be very small. Add a single thing to the state.

Keep up the good work!

mbed67:
@jaapio Is this what you mean by splitting the reducer?
And @jaapio and @mvriel where shall we put the interpreter and interpret classes? It is very small... isn't a separate package overkill? Can't we pull it into the main application?
And also: if this is the way to go, there will be a lot of files to create in order to finish the project mapper. How shall I split up the PR's so that they will not be too large?

jaapio:
Yes this is what I mean. 
I agree with you that is is very small, but that makes it very reusable. It is not an option to put this in the main application since the reflection component has to be a standalone library. So we should either put it in a separate package or in de reflection component. 

An other option is to use http://pipeline.thephpleague.com/ which is almost the same thing?

mbed67:
Hi @jaapio, you are right. I forgot that it is also being used in the reflection component.
The 'pipeline' is indeed very similar and we could use it. However with that solution you would have to finish the whole pipeline once you enter it, and with Mikes solution we can step out of the pipeline if we are done. I don't know how much of an advantage/disadvantage that is...?
If we stick with the interpret/interpreter solution, I think it might be best to put it in a separate package.
I'm also curious about @mvriel his opinion on this :-)

mvriel:
I have a backlog when it comes to reading up on stuff; I have received a few new projects to work on which are taking a fair amount of time. I will respond as soon as I can

PR #1757, opened 2016-05-15
Adds missing xsl namespace

### LINKED ISSUE #1756: Zend template empty
Hi, i have a trouble when i use zend template i got and empty page, but when i use the responsive-twig template i can view the class methods and other info.

Also, in the responsive-twig template the tag @link and @see doesn't display the links to interactive navigation when the target is an external link like that:

@link http://silex.sensiolabs.org/doc/providers.html Silex provider documentation

I doing something bad?

responsive-twig
zend

jaapio:
Namespace php was not defined in docblock xsl. This caused critical errors
during processing the template. Resulting in an empty api-doc. Namespace is required
to be able to use the php: prefix.

fixes #1756


jaapio:
@mbed67 this solves our issue

Issue #1724, opened 2016-01-21
Redundant documentation in interface+implementation, or errors


fi5er:
With an interface defining a method and a class implementing it, both must have nearly identical doc blocks. If the implementation has no doc block on the method, or it's missing the summary or `@param` tags, phpDocumentor2 sees that as an error (in the CLI, with a nasty looking red background, and on the "Errors" report page).

Errors notwithstanding, the doc block from the interface does show up in the generated documentation for the implementation, as it should. As far as I can see, there is no way to silence this type of error.

Is my understanding correct that I have to choose between maintaining two largely identical copies of a method doc block, or living with lots of error messages? When there are lots of non-critical warnings, they tend to get ignored, and any legitimate warning will probably go unnoticed.

```php
/** The IThing interface summary. */
interface IThing
{
    /**
     * Method summary (in interface).
     *
     * Method description (in interface).
     *
     * @param  string  $foo  a parameter
     * @return  string  some string
     */
    public function methodOne($foo);
}

/** The Thing class summary. */
final class Thing implements IThing
{
    // need at least summary + @param here
    public function methodOne($foo) { /*...*/ }
}
```

The same basic question also applies to inherited properties and methods.

jaapio:
With `@inheritdoc` you can import the summary. But there is no way to copy all tags. We will have a look if and how we can achieve this.

waltertamboer:
I am struggling with the same problem. Since all methods should have a docblock, it would be nice to simply have an "@inherit" annotation that inherits the docblock of the parent. I'm not sure if this is considered in PSR-5?

In PSR-5 there is a [part about inheritance](https://github.com/phpDocumentor/fig-standards/blob/b994d853b8ba33bd4bfb00e53cc268d57a5d1e7b/proposed/phpdoc.md#61-making-inheritance-explicit-using-the-inheritdoc-tag) but it seems a bit too unclear to me. I'm not sure if `@inheritDoc` inherits only the summary (because `{@inheritDoc}` does) or if inherits the other tags as well?

https://github.com/php-fig/fig-standards/pull/169/files

jaapio:
I checked that for you. But I can't find anything in de code of phpdocumentor about the `@inheritDoc` tag. So I'm not sure how this works. `{@inheritDoc}`  is implemented. 

[...]

PR #1760, opened 2016-05-16
Restore {@inheritDoc} behavior

### LINKED ISSUE #1505: {@inheritdoc} not working with interfaces
Hi, {@inheritdoc} not working with interfaces. I am using the version 2.8.1. It is not automatically inheriting @param, @return etc... at method level.

jaapio:
{@inheritDoc} now takes the summary and description from the interface that
was implemented.

This parially fixes #1505. I think `{@inheritDoc}` now works as before. But not fully as requested in #1505 

Refs #1724

### DIFF
--- a/src/phpDocumentor/Descriptor/MethodDescriptor.php
+++ b/src/phpDocumentor/Descriptor/MethodDescriptor.php
@@ -250,11 +250,7 @@ class MethodDescriptor extends DescriptorAbstract implements Interfaces\MethodIn
 
         /** @var ClassDescriptor|InterfaceDescriptor $parentClass|null */
         $parentClass = $associatedClass->getParent();
-        if ($parentClass) {
-            if (!$parentClass instanceof ClassDescriptor && !$parentClass instanceof Collection) {
-                return null;
-            }
-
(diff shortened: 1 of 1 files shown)


jaapio:
@mvriel I don't know if this was the original functionality of this tag. But I think is was like this?

mvriel:
The 'original' functionality of this tag was that it could only be used in a description (not in the summary) since summaries don't support the inline tag functionality (marked with the braces).

What it did then was import the parent description into the location of the `{@inheritDoc}` tag so that you could combine the two descriptions into one.

When you omit the summary and description then these are automatically inherited from the parent element (default inheritance behaviour).

Symfony and Doctrine later introduced a usage of `{@inheritDoc}` that it would mean 'inherit all information from the parent element' when used on its own in a DocBlock.

So:

```
/**
 * {@inheritDoc}
 */
```

is equal to omitting the DocBlock as far as phpDocumentor is concerned.

phpDocumentor adopted this behaviour to be compatible with the expectations from Doctrine and Symfony users but only under the condition that is was the only tag in a DocBlock.

jaapio:
I modified this fix. So it works as expected. Somehow the original functionality was broken for interfaces.

Issue #1763, opened 2016-05-23
phpDocumentor 2.9.0 reports itself as 2.8.5


PhrozenByte:
The [`VERSION` file](https://github.com/phpDocumentor/phpDocumentor2/blob/v2.9.0/VERSION) needs to be updated.

```
$ grep -A 1 '"name": "phpdocumentor/phpdocumentor"' composer.lock 
            "name": "phpdocumentor/phpdocumentor",
            "version": "v2.9.0",

$ vendor/bin/phpdoc --version
phpDocumentor version 2.8.5
```

mvriel:
This issue will resolve itself after v2.9.1 is released; I prefer to focus on fixing bugs and releasing that version than trying to re-tag and re-release v2.9.0. Our apologies for the inconvenience

PR #1766, opened 2016-05-25
update version in version file

mbed67:
This PR fixes issue #1763


VERSION
```diff
@@ -1 +1 @@
-2.8.5
\ No newline at end of file
+2.9.0
```
jaapio: should be 2.9.1 since that will be the next release I think

PR #1775, opened 2016-07-02
[wip] use xslt to transform phpdoc2 to phpdoc3

rgeraads:
Instead of supporting multiple strategies, we should only support the latest, and upcast the previous version to the last strategy. XSL can do this for us.
The idea is that if it doesn't match the phpdoc3 strategy, it tries to upcast it to phpdoc3 and retry that strategy.

I finished the first version of the style.xsl (possibly needs renaming) which upcasts phpdoc2 to phpdoc3 config.

Aside from the ConfigurationFactory flow, this can be used to let users convert their old configuration to the new one, e.g. by typing `phpdoc --convert phpdoc2.xml phpdoc3.xml`

What I don't know yet is how to put this in the ConfigurationFactory flow.


jaapio:
Sorry for the delayed response. 

I would create a chain of converters. So you have a 2to3 converter which can be wrapped in a 3to4 converter. Your convert method could be something like this.

```php
  public function convert($config)
  {
      if (!$this->matchesVersion($config) {
          $config = $this->wrappedConverter->convert($config);
      }

      return $this->innerConvert($config);
  }
```
results in:
config => 2to3 => 3to4 => config

This way you only need to modify the factory and add an extra step when needed. Existing classes can be left as is. It even allows people to add there own steps to convert config from a different format.

mvriel:
At this point in time we do not need this functionality due to our re-alignment. I am going to close the PR but if we need a solution like this again we can refer to this closed PR

PR #1781, opened 2016-07-29
Updated readme.md file with PHAR-based installation instructions

cbarina:
After figuring out how to install via PHAR, documented here for others to use.


jaapio:
Thanks!

PR #1797, opened 2016-10-31
Include tree menu navigation memorization cookie

lpotherat:
Quick and dirty draft to remember nav tree position across page changes.

This is my "not perfect" patch that I apply on my own docs, maybe it can bring ideas to implement this behaviour. 

Cookies does not work in local file systems, maybe HTML5 storage can be a solution ?


mvriel:
My apologies that we have not responded sooner to your PR. We have had serious time, life and staffing issues. At this point, we are changing the way templates work for v3; due to this state of flux I will be closing the PR at this moment but I appreciate your input.

Again, my sincere apologies

PR #1815, opened 2016-12-23
fix(method assembler): argument type should not be assigned after name

Fishdrowned:
Wrong result for doc block:
@method foo($param = 'default')
It should return type = mixed and default = 'default', but actually, it return type = \= and default = false instead


jaapio:
Thanks, picked your issue fix. The other fix was solved by updating the tests so it matches the implementation in the application itself.

PR #1817, opened 2017-01-02
Expanded --ignore docs

jaredmorgs:
If applied, this PR implements the following

- Expands the basic *--ignore* examples for new users.
- Provides examples for single, and multiple directories.

## Justification
I was trying to configure phpDocumentor to work with multiple top-level directories and found that the lack of basic *--ignore* syntax was a bit confusing. I had to look at StackOverflow to get help, which usually means that the core project documentation is missing some information. 

Please work with me to correct any project-specific style issues with my proposed changes. I felt it better to propose a starting point, from which we could collaborate to get it ready for inclusion.

### DIFF
--- a/docs/guides/running-phpdocumentor.rst
+++ b/docs/guides/running-phpdocumentor.rst
@@ -82,13 +82,20 @@ For more information on the options and format supported by the configuration it
 Influencing the List of Project Files
 -------------------------------------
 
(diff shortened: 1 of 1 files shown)


jaredmorgs:
Noticed some builds were failing, so if this is because of a code standard I've breached please let me know and I'll fix.

jaapio:
The build is failing for a while. I'm sure that your contribution didn't change this. 

Thanks for your contribution!

PR #1830, opened 2017-03-02
Added `@​see` inline example

sloanlance:
The existing documentation doesn't indicate that `@​see` may be used inline.


mvriel:
Thanks for the addition! I have merged it and will backport it to master as well

PR #1836, opened 2017-04-12
Correct link for Markdown

fridde:
I have no idea why *Markdown* linked to some weird site with no clear connection to the language. Did the content change or has someone hijacked the site?

### DIFF
--- a/docs/getting-started/your-first-set-of-documentation.rst
+++ b/docs/getting-started/your-first-set-of-documentation.rst
@@ -216,5 +216,5 @@ phpDocumentor; see the guide on :doc:`../guides/running-phpdocumentor` for more 
 .. _Class constant: http://php.net/manual/en/language.oop5.constants.php
 .. _Property:       http://php.net/manual/en/language.oop5.properties.php
 .. _Method:         http://php.net/manual/en/language.oop5.basic.php
-.. _Markdown:       http://daringfireball.com
+.. _Markdown:       https://help.github.com/articles/basic-writing-and-formatting-syntax/
 .. _templates:      http://phpdoc.org/templates


jaapio:
Thanks!

ashnazg:
`daringfireball.net` has been the longest standing Markdown reference link I have, and I'm fairly sure I got it from @mvriel to begin with.

Mike, it's legit for phpDocumentor to point to it as the canonical reference, right?

ashnazg:
Also, I _think_ that Github's flavor of markdown differs at least a little bit, and thus if @mvriel 's original Markdown reference for phpDocumentor's implementation was indeed `daringfireball.net`, then it's possible the Github reference will lead phpDoc writers astray.

fridde:
@ashnazg Sorry for not getting back earlier.

After I looked around on Daring Fireball for a while, I actually found what you where referring to. Maybe a link directly to https://daringfireball.net/projects/markdown/basics would have made it more clear that the link had anything to do with markdown. 

Simply linking to daringfireball.net was like linking *Cat* to *https://www.wikipedia.org/* and expecting  people like me to understand that the article about cats is hidden *somewhere*. (I'm exaggerating to make a point, sorry)

Anyway, I got my information from [Line 102](https://github.com/phpDocumentor/phpDocumentor2/blob/68cf85e30aab12ce27bb8889f765f14705b2f7b9/docs/guides/docblocks.rst#L102) of the guide on Docblocks, so that's why I assumed that the Github flavor was the one to link to. But I might be wrong.

Do you know where in the code the markdown parser is invoked? Maybe that would clear things up.

ashnazg:
Ah, that Docblocks guide text does indeed say to follow the Github flavor... that's definitive enough for me.  +1 on this change, then.

PR #1845, opened 2017-05-16
depend on newer fileset for symfony 3.x support

chrisjohnson00:
when running a symfony 3.x project, you can't use phpdoc because fileset depends on a symfony 2.x version of symfony/finder.  This, paired with an update to phpdocumentor/fileset allows symfony 3.x projects to use phpdoc.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -30,7 +30,7 @@
         "symfony/console":                   "^2.8|~3.0",
         "symfony/validator":                 "^2.8|~3.0",
         "phpdocumentor/graphviz":            "^1.0",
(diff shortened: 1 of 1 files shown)


jaapio:
Thanks for your interest in phpdocumentor. I won't merge your pr since we are currently looking at our priorities. 

Beside that, we advise our users to use the phar or pear install method. We simply can't keep up with the other projects and be compatible with all configurations. In the future we might even drop support for composer.

chrisjohnson00:
ok, thanks.

Issue #1767, opened 2016-05-25
Invalid @see class references in clean template


PhrozenByte:
When using the `@see` block tag to reference a class in the root namespace (e.g. `Pico`), phpDocumentor 2.9.0 fails to interpret the namespace and shows a reference to `\\Pico` in the clean template (the link target is a empty string).

``` php
/**
 * …
 *
 * @see Pico
 */
```

![bildschirmfoto von 2016-05-25 22 29 50](https://cloud.githubusercontent.com/assets/920356/15555278/5171b7bc-22c8-11e6-90ce-2c24da842286.png)

Also see http://picocms.org/phpDoc/master/classes/PicoPluginInterface.html#method_getPico

mbed67:
This issue has been resolved with the following PR: https://github.com/phpDocumentor/phpDocumentor2/pull/1848

PR #1848, opened 2017-05-21
Fix bug #1767: Also find elements that are in the global namespace when using @see tag

mbed67:
The linker treated elements that were referenced using a @see tag but were in the global namespace as undocumented because it was only looking at the namespace of the container class.
A check for the global namespace is added

The unit test has been extended to make sure code coverage is 100%. Also the incomplete tests are fixed.

The createMock... methods have been moved under the public methods (sorry, this messes up the diff a bit in github)


jaapio:
:+1: looks good to me.

PR #1852, opened 2017-05-30
Correction in property fqsen

mbed67:
One of the issues that was discovered regarding the @see tag (see investigation described in https://github.com/phpDocumentor/phpDocumentor2/issues/1079) was a missing $ sign in the property fqsen displayed from a regular @see tag.

The displayed fqsen would be something like this: \Namespace\Classname::property instead of \Namespace\Classname::$property.

When inspecting the ProjectDescriptor, I noticed that the fqsens in the PropertyDescriptors were missing the $ sign as well. However, the property elements in the ProjectDescriptor - Indexes - Elements did have a $ sign. That $ sign was added in the ElementsIndexBuilder, but that didn't fix the fqsen in the PropertyDescriptor which is being used to display the property name in a @see or @link tag. With this commit the $ sign is added in the correct place.

Also the test coverage of PropertyDescriptor is completed to 100%

Using the test script described in issue 1079 I checked if this change only affected the displayed fqsen in the regular @see tags and if all other texts and links stayed the same (still containing the same bugs as well). This is the case.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,6 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
         "This file is @generated automatically"
     ],
-    "hash": "b4c4a21e61177c5c41ca830b97440296",
     "content-hash": "a130907616ec35df4ea63f3eb004f545",
     "packages": [
         {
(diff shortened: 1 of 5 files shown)


mbed67:
@jaapio I tested it through the browser with the clean template and the responsive template and that seems to work. These templates don't need to be adjusted. If you like I could test more templates.

jaapio:
Good enough for me now. But we have to improve those tests as discussed. :-)

Issue #1726, opened 2016-01-22
Internal links created by `@see` are broken


fi5er:
Using phpDocumentor2 version 2.8.5, most of the internal links created by `@see` or `{@see}` are unusable (absolute, missing, or contain backslashes etc):

``` php
namespace TheNS;

/** The class docu. */
class TheClass {

    /** @var int  A property description. */
    public $aProperty = 42;

    /**
     * The method summary.
     *
     * - {@see \TheNS\TheClass}
     * - {@see \TheNS\TheClass::$aProperty}
     * - {@see \TheNS\TheClass::aMethod()}
     * - {@see \TheNS\TheClass              the class}
     * - {@see \TheNS\TheClass::$aProperty  the property}
     * - {@see \TheNS\TheClass::aMethod()   the method}
     *
     * @see \TheNS\TheClass
     * @see \TheNS\TheClass::$aProperty
     * @see \TheNS\TheClass::aMethod()
     *
     * @see \TheNS\TheClass              the class
     * @see \TheNS\TheClass::$aProperty  the property
     * @see \TheNS\TheClass::aMethod()   the method
     */
    public function aMethod () {}
}
```

All six inline `{@see}` tags create links with absolute URLs: `href="/classes/TheNS.TheClass.html"`, tested with the "clean" and "responsive" templates.

The next three `@see` links are missing from the docu, see issue #1079.

The last three `@see` links are OK in the "clean" template, but in the "responsive" template, the href attributes are `%5CTheNS%5CTheClass`, `%5CTheNS%5CTheClass::aProperty`, and `%5CTheNS%5CTheClass::aMethod()`, respectively. `%5C` is a backslash.

This goes for the documentation of non-namespaced files as well. It doesn't make a difference whether FQSENs are used or not.

frickenate:
Confirming on my part in 2.8.5 using the default clean template. If there is no text attached to the tag, then the link tag is generated by without any link text, making it impossible to see or follow the declarations to their definition.

PhrozenByte:
dup of #1672

@frickenate, please don't report problems in random issues, always create a new issue for a new problem (otherwise it's nearly impossible for the phpDoc devs to track and fix the issues). Your additional remarks have nothing to do with @fi5er's issue. Just fyi, the issue with missing text links is a dup of #1670; I don't know anything about your second paragraph, please make sure it hasn't been reported yet using GitHub's search engine and open a new issue when necessary.

[...]

Issue #1079, opened 2013-11-23
@see without description not working


smcjones:
If I use @see className, nothing shows up.. I need to type @see className className for it to be effective.

fi5er:
> I believe this issue is resolved in version 2.6; I want to test this again to be sure

@mvriel: Just checked, it's still there in 2.8.5 (tested "clean" and "responsive" templates).

mbed67:
I'm investigating the issue using this test file and the phpDocumentor2 code base. The phpDocumentor2 version is the latest master but using version 2.0.5 of phpdocumentor/reflection-docblock
[TestSeeTagIssue.php.zip](https://github.com/phpDocumentor/phpDocumentor2/files/1035792/TestSeeTagIssue.php.zip)

mbed67:
The results are as follows for templates responsive and clean:
```
namespace phpDocumentor\Descriptor;

use \phpDocumentor\Descriptor\Builder\AssemblerAbstract;

/**
 * Class to check if see tags are working correctly
 *
 * Inline relative class {@see AssemblerAbstract testClass relative} : responsive + clean ok
 * Inline relative class {@see AssemblerAbstract} : responsive + clean ok
 *
 * Inline absolute class {@see \phpDocumentor\Descriptor\Builder\AssemblerAbstract testClass absolute} : responsive + clean ok
 * Inline absolute class {@see \phpDocumentor\Descriptor\Builder\AssemblerAbstract} : responsive + clean ok
 *
 * Inline relative method {@see AssemblerAbstract::getBuilder() testMethod relative} : responsive + clean ok
 * Inline relative method {@see AssemblerAbstract::getBuilder()} : responsive + clean ok
 *
 * Inline absolute method {@see \phpDocumentor\Descriptor\Builder\AssemblerAbstract::getBuilder() testMethod absolute} : responsive + clean ok
 * Inline absolute method {@see \phpDocumentor\Descriptor\Builder\AssemblerAbstract::getBuilder()} : responsive + clean ok
 *
 * Inline relative property {@see AssemblerAbstract::$builder testProperty relative} : responsive + clean ok
 * Inline relative property {@see AssemblerAbstract::$builder} : responsive + clean ok
 *
 * Inline absolute property {@see \phpDocumentor\Descriptor\Builder\AssemblerAbstract::$builder testProperty absolute} : responsive + clean ok
[...]

PR #1854, opened 2017-06-02
Issue 1726 internal see links broken

mbed67:
The inline relative see tags in docblocks of properties and methods could not be resolved for elements in the same namespace as the container class because the property and method descriptors were missing their namespace. This way the context was not complete for the see tags and the relevant element could not be found. With this change the property and method assemblers will set the namespace in the descriptors.

This PR fixes one of the issues that were discovered in the test of issue 1079. This leaves two more bugs to fix.

The code coverage of the two classes is completed to 100%

The result is visually tested with the clean and responsive templates. The links work again and no negative side effects have been noticed.

refs #1726, #1079


jaapio:
I don't have any comments on this. Excellent job!

Issue #822, opened 2013-05-16
Add support for magic static methods


mvriel:
The current `@method` does not describe static methods; support for this is in PHP so it should be added as an option

jaapio:
Fixex, will be included in v2.9.1

PR #1855, opened 2017-06-04
Support `@method static int foo()` syntax for magic methods

TysonAndre:
After this change, phpdoc API documentation generated from `@method` tags will indicate that the method is static the phpdoc tag was of the form `@method static Type foo()`

Do this by reading isStatic() from ReflectionDocBlock

- This was added to ReflectionDocBlock before the docblock 2.0.0 release,
  so it should be safe to assume that it's available (excluding beta versions of ReflectionDocBlock)

This was tested locally with the following snippet 
(verified generated HTML API documentation was correct and static):

    /** 
     * @method static int fooStatic() implicitly a void
     */  
    class MyClass{}

Add a draft of updating the docs for static method support.
- Not sure how to indicate the first version where phpdocumentor supports static method
- Feel free to suggest the exact wording and layout, or fix it after this PR

For issue #822

### DIFF
--- a/tests/unit/phpDocumentor/Descriptor/ClassDescriptorTest.php
+++ b/tests/unit/phpDocumentor/Descriptor/ClassDescriptorTest.php
@@ -361,8 +361,10 @@ class ClassDescriptorTest extends \PHPUnit_Framework_TestCase
 
     /**
      * @covers phpDocumentor\Descriptor\ClassDescriptor::getMagicMethods
+     * @dataProvider provideMagicMethodProperties
+     * @param bool $isStatic
      */
-    public function testGetMagicMethods()
+    public function testGetMagicMethods($isStatic)
     {
         $methodName  = 'methodName';
         $description = 'description';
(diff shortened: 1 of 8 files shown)


tests/unit/phpDocumentor/Descriptor/ClassDescriptorTest.php
```diff
@@ -377,6 +377,7 @@ public function testGetMagicMethods()
         $methodMock->shouldReceive('getDescription')->andReturn($description);
         $methodMock->shouldReceive('getResponse')->andReturn($response);
         $methodMock->shouldReceive('getArguments')->andReturn($arguments);
+        $methodMock->shouldReceive('isStatic')->andReturn(true);
```
jaapio: Please do add this to a new test. It is unclear to me now if the original behavior was not changed unintended.
It looks like is't ok, but I want to be sure.

tests/unit/phpDocumentor/Descriptor/TraitDescriptorTest.php
```diff
@@ -101,6 +101,7 @@ public function testMagicMethodsReturnsExpectedCollectionWithTags()
     {
         $mockMethodDescriptor = m::mock('phpDocumentor\Descriptor\Tag\MethodDescriptor');
         $mockMethodDescriptor->shouldReceive('getMethodName')->andReturn('Sample');
+        $mockMethodDescriptor->shouldReceive('isStatic')->andReturn(false);
```
jaapio: Same applies here, please add a new testcase.

jaapio:
Thanks for the quick response.

PR #1856, opened 2017-06-07
Issue-1079 and issue-1026 Fix @see links in responsive template

mbed67:
Regular @see links were not visible as text and not working as links in the responsive template.
The links were treated the same way as the license, link and author tags, but because they may reference structural elements, this doesn't always work. With this PR the @see and @uses tags are dealt with in a separate template. And when there is no 'http' in the link, they will be processed using the already existing xslt extension: **phpDocumentor\Plugin\Core\Xslt\Extension::path**

This fix has been tested in responsive template with regular @see tags in docblocks of class, method, property and constant. @see tags referencing structural elements and @see tags referencing websites were used in testing. @uses don't work well, but that is because they don't end up in the structure.xml correctly. This is another bug for which I will make an issue.

Regular @see tags referencing a class using a namespace alias don't work either. That was the last bug I found in my investigation in issue-1079. For that I will make a separate PR.

### DIFF
--- a/data/templates/responsive/elements/common.xsl
+++ b/data/templates/responsive/elements/common.xsl
@@ -167,11 +167,14 @@
     <xsl:template match="tag[@name = 'see' or @name = 'uses']" mode="tabular">
         <xsl:variable name="url">
             <xsl:choose>
-                <xsl:when test="not(contains(@link, 'http'))">
+                <xsl:when test="php:function('phpDocumentor\Plugin\Core\Xslt\Extension::typeOfElement', string(@link)) = 'documented'">
                     <xsl:value-of select="concat($root, php:function('phpDocumentor\Plugin\Core\Xslt\Extension::path', string(@link)))" />
                 </xsl:when>
-                <xsl:otherwise>
+                <xsl:when test="php:function('phpDocumentor\Plugin\Core\Xslt\Extension::typeOfElement', string(@link)) = 'url'">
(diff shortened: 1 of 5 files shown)


data/templates/responsive/elements/common.xsl
```diff
@@ -164,10 +164,39 @@
         </tr>
     </xsl:template>
 
-    <xsl:template match="tag[@name = 'license' or @name = 'link' or @name = 'see' or @name = 'uses' or @name = 'author']" mode="tabular">
+    <xsl:template match="tag[@name = 'see' or @name = 'uses']" mode="tabular">
+        <xsl:variable name="url">
+            <xsl:choose>
+                <xsl:when test="not(contains(@link, 'http'))">
```
jaapio: I expect this to fail when an referenced element contains 'http', I think you need to check if the link is a valid path for the extension used below. You might need to add an extra method to that class.

mbed67: @jaapio when an element contains 'http' it will take the 'otherwise' branch on line 173

mbed67: hmm you mean when part of the documentation contains http?

jaapio: Exactly, an class call HttpClient is not uncommon.

mbed67: Ah, I see what you mean. Good one :-)

PR #1860, opened 2017-06-24
Issue 1079 regular relative see tags to other class

mbed67:
Relative classes that were represented in a namespace alias were treated as ambiguous type and passed on to the Linker for further investigation.
This way they got the wrong namespace, namely the namespace of the container class, and they couldn't be found in the element list.
With this commit a check is added to see if an element is represented by a namespace alias.

Also urls with scheme is ftp:// are left unchanged.

Unit tests are added.

### DIFF
--- a/src/phpDocumentor/Descriptor/Builder/Reflector/Tags/SeeAssembler.php
+++ b/src/phpDocumentor/Descriptor/Builder/Reflector/Tags/SeeAssembler.php
@@ -38,23 +38,22 @@ class SeeAssembler extends AssemblerAbstract
         $descriptor->setDescription($data->getDescription());
 
         $reference = $data->getReference();
+        $context = $data->getDocBlock() ? $data->getDocBlock()->getContext() : null;
 
         if (substr($reference, 0, 7) !== 'http://'
             && substr($reference, 0, 8) !== 'https://'
+            && substr($reference, 0, 6) !== 'ftp://'
             && $reference !== 'self'
             && $reference !== '$this'
+            && $reference[0] !== '\\'
         ) {
             // TODO: move this to the ReflectionDocBlock component
             // Expand FQCN part of the FQSEN
             $referenceParts = explode('::', $reference);
-            if (count($referenceParts) > 1 && $reference[0] != '\\') {
-                $type = current($referenceParts);
-                $type = new Collection(
-                    array($type),
-                    $data->getDocBlock() ? $data->getDocBlock()->getContext() : null
-                );
-                $referenceParts[0] = $type;
-            } elseif (isset($reference[0]) && $reference[0] != '\\') {
+
+            if (count($referenceParts) > 1 || $this->referenceIsNamespaceAlias($reference, $context)) {
+                $referenceParts = $this->setFirstReferencePartAsType($context, $referenceParts);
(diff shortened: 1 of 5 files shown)


src/phpDocumentor/Descriptor/Builder/Reflector/Tags/SeeAssembler.php
```diff
@@ -38,23 +38,22 @@ public function create($data)
...
+        $context = $data->getDocBlock() ? $data->getDocBlock()->getContext() : null;
 
         if (substr($reference, 0, 7) !== 'http://'
             && substr($reference, 0, 8) !== 'https://'
+            && substr($reference, 0, 6) !== 'ftp://'
             && $reference !== 'self'
             && $reference !== '$this'
+            && $reference[0] !== '\\'
         ) {
             // TODO: move this to the ReflectionDocBlock component
```
jaapio: Think this todo is not needed anymore since the code below has been changed? Or it should be moved to another line?

src/phpDocumentor/Descriptor/Builder/Reflector/Tags/SeeAssembler.php
```diff
@@ -65,4 +64,37 @@ public function create($data)
 
         return $descriptor;
     }
+
+    /**
+     * @param $context
```
jaapio: Please add type hints here

src/phpDocumentor/Descriptor/Builder/Reflector/Tags/SeeAssembler.php
```diff
@@ -65,4 +64,37 @@ public function create($data)
...
+     * @param $referenceParts
+     * @return mixed
```
jaapio: Looks like this return type is incorrect

src/phpDocumentor/Descriptor/Builder/Reflector/Tags/SeeAssembler.php
```diff
@@ -65,4 +64,37 @@ public function create($data)
...
+     *
+     * @param $reference
+     * @param $context
+     * @return bool
+     */
+    private function referenceIsNamespaceAlias($reference, $context)
+    {
+        /** @var \phpDocumentor\Reflection\DocBlock\Context $context*/
+        foreach ($context->getNamespaceAliases() as $alias) {
+            if (substr($alias, -strlen($reference), strlen($reference)) === $reference) {
```
jaapio: substr with a negative value will already return the string from the end. So I think you can remove the third param here.

tests/unit/phpDocumentor/Transformer/Router/RendererTest.php
```diff
@@ -0,0 +1,253 @@
...
+
+    public function setUp()
+    {
+        $this->originalQueue = m::mock('phpDocumentor\Transformer\Router\Queue');
+        $this->renderer = new Renderer($this->originalQueue);
+    }
+
+
+    /**
+     * @covers \phpDocumentor\Transformer\Router\Renderer::__construct
```
jaapio: Please use the `@coversDefaultClass` param to remove the redundant fqcn

PR #1879, opened 2017-08-17
Replace selfupdate command

### LINKED ISSUE #1820: Abandoned package "herrera-io/phar-update" leads to warnings
Looks like the abovementioned package has recently changed it's status to "abandoned" (together with some other sub-dependencies). When phpdocumentor is included in a project as composer dependency, this leads to the following warnings on `composer update`:

```bash
Package kherge/version is abandoned, you should avoid using it. No replacement was suggested.
Package herrera-io/json is abandoned, you should avoid using it. Use kherge/json instead.
Package herrera-io/phar-update is abandoned, you should avoid using it. No replacement was suggested.
```

The mentioned packages should be replaced.

jaapio:
Update the selfupdate command to be compatible with github releases.
Will now update default to the latest stable version. With the --pre option
a unsable release can be updated.

Fixes #1820

### DIFF
--- a/src/phpDocumentor/Command/Phar/UpdateCommand.php
+++ b/src/phpDocumentor/Command/Phar/UpdateCommand.php
@@ -30,6 +30,7 @@ use Symfony\Component\Console\Output\OutputInterface;
  */
 class UpdateCommand extends Command
 {
+    const PHAR_URL = 'https://github.com/phpDocumentor/phpDocumentor2/releases/latest';
 
     /**
      * Initializes this command and sets the name, description, options and arguments.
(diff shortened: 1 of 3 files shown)


src/phpDocumentor/Command/Phar/UpdateCommand.php
```diff
@@ -62,63 +62,56 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
 
-        if (strpos($version, 'v') === 0) {
-            // strip v tag from phar version for self-update
-            $version = str_replace('v', '', $version);
+        if (PHP_VERSION_ID < 50600) {
+            $message = 'Self updating is not available in PHP versions under 5.6.' . "\n";
+            $message .= 'The latest version can be found at ' . self::PHAR_URL;
+            $output->writeln(sprintf('<error>%s</error>', $message));
+            return 1;
+        } elseif (Application::VERSION === ('@' . 'package_version' . '@')) {
```
mbed67: Could this be written in one string '@package_version@' or is it split up for a reason?

PR #1882, opened 2017-08-21
Issue 920 Make uses show up in templates clean and responsive

mbed67:
With this PR the @uses tags show up in the clean and responsive templates. 
The @throws tags and the @used-by tags have not yet been fixed.

A behat scenario is included to test the correct working for tags referring to the same file. However, it has been manually tested that referrals to other files also work.


jaapio:
I don't see any issues in this. Thanks again for this great job!

PR #1894, opened 2017-09-27
Make sure Fqsen's and Url's are handled correctly in StandardRouter

mbed67:
Due to refactoring the reference in the see tags now points to an fqsen object instead of a descriptor. These could not be handled yet in the StandardRouter which caused the urls to go wrong in the output.

Regular @see tags work again, except for tags pointing to methods and properties (and probably constants) in a class that is in a namespace alias. It doesn't have the correct namespace in the fqsen currently. I will make a separate task for that.

### DIFF
--- a/tests/unit/phpDocumentor/Transformer/Router/UrlGenerator/Standard/FqsenDescriptorTest.php
+++ b/tests/unit/phpDocumentor/Transformer/Router/UrlGenerator/Standard/FqsenDescriptorTest.php
@@ -1,0 +1,67 @@
+<?php
+/**
+ * phpDocumentor
+ *
+ * PHP Version 5.3
+ *
+ * @copyright 2010-2014 Mike van Riel / Naenius (http://www.naenius.com)
+ * @license   http://www.opensource.org/licenses/mit-license.php MIT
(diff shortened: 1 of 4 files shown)


tests/unit/phpDocumentor/Transformer/Router/UrlGenerator/Standard/FqsenDescriptorTest.php
```diff
@@ -0,0 +1,67 @@
...
+        $result = $fixture($fqsen);
+
+        // Assert
+        $this->assertSame(false, $result);
+    }
+
+    public function provideFqsens()
+    {
+        return array(
+            array('\\My\\Space\\Class', '/classes/My.Space.Class.html'),
```
jaapio: Looks like you are missing a namespace conversion here. Not sure how to detect the difference between a namespace and class because the fqsen are equal. And can even overlap. So I think this is a short comming to the full implementation stack. 
In other parts of phpdocumentor namespaces are prefixed with a `~`. But that won't work here since that's not an allowed FQSEN format. e.g `~\\My\\Space` 

Think we should document that somewhere?

PR #1897, opened 2017-10-04
Correct see fqsens

mbed67:
When running phpdoc against it's own code base, it crashed on a @see tag of src/phpDocumentor/Plugin/Scrybe/Template/Twig.php because it was not a valid FQSEN. While correcting that tag, some other @see tags in the same file were corrected and the unit test was extended


jaapio:
No comments on this one. Thanks!

PR #1903, opened 2017-10-12
Fix menu always hidden

UziTech:
The first argument to .filter is the index.
The second argument is the element.


jaapio:
Thanks!

PR #1911, opened 2017-10-31
fix argument default = 0

UziTech:
when an argument default is `0` then it would not show

### DIFF
--- a/data/templates/responsive-twig/class.html.twig
+++ b/data/templates/responsive-twig/class.html.twig
@@ -61,7 +61,7 @@
                         <a id="method_{{ method.name }}"></a>
                         <div class="element clickable method {{ method.visibility }} {{ method.deprecated ? 'deprecated' }} method_{{ method.name }}{{ method.parent.name != node.name ? ' inherited' : '' }}" data-toggle="collapse" data-target=".method_{{ method.name }} .collapse">
                             <h2>{{ method.summary ?: method.name }}</h2>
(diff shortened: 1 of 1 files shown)


data/templates/responsive-twig/class.html.twig
```diff
@@ -61,7 +61,7 @@
                         <a id="method_{{ method.name }}"></a>
                         <div class="element clickable method {{ method.visibility }} {{ method.deprecated ? 'deprecated' }} method_{{ method.name }}{{ method.parent.name != node.name ? ' inherited' : '' }}" data-toggle="collapse" data-target=".method_{{ method.name }} .collapse">
                             <h2>{{ method.summary ?: method.name }}</h2>
-                            <pre>{{ method.name }}({% for argument in method.arguments %}{{ argument.types ? argument.types|join('|')~' ' }}{{ argument.byReference ? '&' }}{{ argument.name }}{{ argument.default ? ' = '~argument.default }}{% if not loop.last %}, {% endif %}{% endfor %}) {{ method.response.types ? ': '~method.response.types|join('|') }}</pre>
+                            <pre>{{ method.name }}({% for argument in method.arguments %}{{ argument.types ? argument.types|join('|')~' ' }}{{ argument.byReference ? '&' }}{{ argument.name }}{{ argument.default != '' ? ' = '~argument.default }}{% if not loop.last %}, {% endif %}{% endfor %}) {{ method.response.types ? ': '~method.response.types|join('|') }}</pre>
```
jaapio: It would be better to check for a null value with !== this does not cast the type which your current check does.

If the default value is not null but an empty string we should fix that also

UziTech: twig doesn't have a `!==` but i changed it to `is not null`

jaapio:
Thanks for your quick response

PR #1919, opened 2017-12-11
Use Null Coalesce Operator

carusogabriel:


jaapio:
Thanks for this small code improvement!

Issue #694, opened 2012-12-02
scalar type support


ro0NL:
Hi,

`@param scalar $foo` seems to be an invalid type. I use this quite often in documentation because my code also uses `is_scalar` to check for it. Basicly an equivalent for `@param  float|integer|string|boolean $foo`.

It currently renders as `Namespace\Class\scalar` :(

PR #1923, opened 2017-12-22
[Docs] Add scalar to supported types

icanhazstring:
With PR #694 the `scalar` type was added as a valid type but the docs weren't updated.
Now many parsing relying on the current state of the docs don't recognize `scalar` as a valid type.


jaapio:
Thanks!
Also merged this to our develop branch

icanhazstring:
Thanks for merging. Merry Xmas ;)

PR #1932, opened 2018-01-01
WIP: Feature/v3 configuration

jaapio:
This pr is a follow up on pr #1940. 
It replaces the configuration of phpdocumentor with the new v3 config setup. A lot of refactoring was needed to accomplish this. I moved the commands to stages of a pipeline, the commands are refactored to pipeline only.


jaapio:
This pr is blocked by some internal discussions and tests of a new plugin structure.

ashnazg:
If there's a `.env.dist`, does that imply that the `.env` file should not be committed in?

src/phpDocumentor/Parser/Middleware/EmittingMiddleware.php
```diff
@@ -26,6 +26,7 @@
      */
     public function execute($command, callable $next)
     {
+        var_dump($command->getFile()->path());
```
ashnazg: `var_dump()` intentional?

jaapio: Saw that one to. Fixing it currently :smiley:

PR #1934, opened 2018-01-03
Feature/remove cilex

jaapio:
Cilex is blocking the updates of symfony components. This
pr removes a lot of magic from the commands which were fetching dependencies
form the application.
The application is updated with some logic to access the inner container to support
the old behavior of the application like cilex provides. This needs to be removed
in the future.

### DIFF
--- a/box.json
+++ b/box.json
@@ -19,7 +19,6 @@
                 "tests"
             ],
             "in": [
-                "vendor/cilex/cilex/src",
                 "vendor/composer",
                 "vendor/container-interop",
                 "vendor/doctrine/annotations/lib",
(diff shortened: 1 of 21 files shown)


mvriel:
Consider merging https://github.com/phpDocumentor/phpDocumentor2/pull/1940 instead.

This PR breaks BC because all userland ServiceProviders depend on the Cilex Application. In the PR above I have re-implemented a basic "Cilex application" class to keep BC. We should throw DEPRECATED warnings when people still use this and remove the BC layer in phpDocumentor 4

jaapio:
Replaced by #1940

Issue #1930, opened 2017-12-31
phpDocumentor alpha 3 not generating any document


NestorAcevedo:
phpDocumentor Alpha 3 is not generating any documentation. Only generates page skeleton.
## Expected Behavior

Normal PHP Documentor generated. Commonly it takes some time to finish.

## Actual Behavior
Generating only page skeleton. Just like the issue https://github.com/phpDocumentor/phpDocumentor2/issues/1874

Commonly when I run phpDocumentor on my project, it takes around 1 minute to achieve the documentation. But in this case only takes less than 1 second

## Steps to Reproduce the Problem

  1. Run phpdocumentor from command line 

## Your environment

PHP7 + MariaDB. Yii Framework 2.0.13

* Version used: 3.0.0-alpha1
* Install method Phar file
* php version 7.1.12
* Operating system and version (e.g. Ubuntu 16.04, Windows 7): openSUSE Tumbleweed 
* Link to your project: Local

ashnazg:
In case it helps... I originally had the same "fast execution, empty docs" behavior yesterday.  My problem turned out to just be running the executable from inside the ./bin dir instead of from the root dir that contained the phpdoc.dist.xml.

Bad result:  `cd phpDocumentor2/bin; ./phpdoc -c ../phpdoc.dist.xml`
Good result:  `cd phpDocumentor2; ./bin/phpdoc -c phpdoc.dist.xml`

Granted, you're probably _not_ executing from a phpDocumentor2 clone, but I did want to highlight the _possibility_ that some unexpected issue with _relative pathing_ might be in play here.

ashnazg:
I'm duplicating this behavior right now, when using any `-d` value except just `.`.  It looks like the underlying file finder logic _does_ see the files in these dirs, _but_ they are not "satisfying the specification" when each file is considered.

jaapio:
I do not completely understand what command you are executing right now.

I found an issue with single file patterns using `-f.` could you add the actual commands you tried. So we can write some tests to check the behavior.

ashnazg:
`./bin/phpdoc -d src -t build` from the root of a phpDocumentor2 clone.

NestorAcevedo:
In my case, I run inside my local bin directory "/home/myuser/bin"
`php phpDocumentor.phar -i "vendor/,console/migrations/" -d "/home/myuser/public_html/myproject" -t "/home/myuser/Documentos/myproject/" --title="My project"`

[...]

PR #1936, opened 2018-01-05
specs need empty path initialization when using local filesystem Dsn;

ashnazg:
Fixes Issue #1930


mvriel:
Let's merge this as soon as Travis agrees

PR #1939, opened 2018-01-06
Add custom phpstan image

jaapio:
since phpdocumentor needs some extra packages we need a custom build
docker image.


jaapio:
@mvriel I had to switch to another base image due some issue in the original phpstan image. 
Will merge this when travis build is passing

PR #1940, opened 2018-01-07
[WIP]: Initial effort to replace Cilex with Symfony4

mvriel:
In this change I have replaced Cilex with Symfony 4. The basic idea behind this change is that it would be easier to build the application on a complete framework, and the bundles could work as a plugin system instead of our hand-rolled one.

As it stands, Symfony 4 is light weight enough to work as a microframework for all intents and purposes.

This change is not complete and merge-able, the following things need to be done:

- [x] Move the CompatibilityLayer class into a Bundle and split it into the right classes
- [x] See if the ServiceProvider can use the ContainerBuilder instead of the Container
- [ ] Undo the changes in the ServiceProvider classes to ensure that BC is completely kept
- [ ] Add E_USER_DEPRECATED messages when people use stuff from the Compatibility Layer so that we can remove it in phpDocumentor 4

When this is done we can subsequently begin migrating the services from our own Service Providers


bin/console
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/env php
```
jaapio: Think this file should be named phpdoc?

bin/php-parse
```diff
@@ -0,0 +1 @@
+../vendor/nikic/php-parser/bin/php-parse
```
jaapio: I think all other files in bin should be ignored.

jaapio:
closed by #1932

PR #1943, opened 2018-01-12
Use Null Coalesce Operator

carusogabriel:


jaapio:
Thanks for your contribution. But I don't think this makes it more readable. So I don't want to merge this PR.

PR #1957, opened 2018-02-11
Fix/1925 ignore

jaapio:
I added a number of tests to show that the recent fixes done by @ashnazg are solving the issue of ignored files.


ashnazg:
Need to bump php versions in appveyor file.

PR #1963, opened 2018-03-18
Adds marker tests

### LINKED ISSUE #1956: Markers are not detected anymore
## Expected Behavior
I have a file with a comment, exactly copied from the phpdocumentor configuration manual (http://docs.phpdoc.org/references/configuration.html). I would expect to see it in the generated output when I click on "Reports"->"markers"

## Actual Behavior
No marker is parsed or transformed into the final documentation.

  1. Make any php-file with a TODO or FIXME marker
  2. Run the phpdoc to generate the documentation
  3. No marker is detected

## Your environment

Include as many relevant details about the environment you experienced the bug in and how to reproduce it.
Since many issue might only occur on a specific code base it would be very helpfull if you link to your project
or add any code that is involved in to the issue.

* Version used: 3.0.0-alpha1
* Install method: composer
* PHP 7.1.10 (cli) (built: Oct  6 2017 01:08:19) ( NTS )
Copyright (c) 1997-2017 The PHP Group
Zend Engine v3.1.0, Copyright (c) 1998-2017 Zend Technologies
* Operating system and version (e.g. Ubuntu 16.04, Windows 7): OSX 10.13.2

jaapio:
Our previous reflection component was able to find markers in the source code.
We removed this behaviour during refactoring to focus on reflecting the source tree and
not it's contents.
After some benchmarking I found that this would be the most suitable way of implementing the
marker feature in phpdocumentor it self.

fixes #1956


jaapio:
Had some issues with the current implementation because the source of files is not always included. The --sourcecode flag is not working currently.

PR #1989, opened 2018-06-15
recognize windows filesystem

ashnazg:
This gets the `testInvokeOverridesConfig` test passing on Windows.

### DIFF
--- a/src/phpDocumentor/DomainModel/Uri.php
+++ b/src/phpDocumentor/DomainModel/Uri.php
@@ -98,10 +98,12 @@ final class Uri
      */
     private function addFileSchemeWhenSchemeIsAbsent($uri): string
     {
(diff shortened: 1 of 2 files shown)


src/phpDocumentor/DomainModel/Uri.php
```diff
@@ -98,7 +98,10 @@ private function validateUri($uri)
      */
     private function addFileSchemeWhenSchemeIsAbsent($uri): string
     {
-        if (!parse_url($uri, PHP_URL_SCHEME)) {
+        $results = parse_url($uri, PHP_URL_SCHEME);
+        if ($results === 'C') {
```
jaapio: Think this is to limited. C is the default drive on windows. But it can be any letter. Think the check should be something like  `preg_match($results, '[a-z]')`

PR #1994, opened 2018-06-20
Update to alpha dependencies

ashnazg:


.travis.yml
```diff
@@ -1,48 +1,58 @@
...
 
 jobs:
   include:
-    - stage: test
+    - stage: dogfood
+      php: 7.1
       script:
-      - bin/phpunit --no-coverage
-      - bin/behat -p default
+        - chmod u+x ./bin/phpdoc
```
jaapio: Think we should change these permissions in the repo. not here. Git holds the file permissions

.travis.yml
```diff
@@ -1,48 +1,58 @@
...
+        - ./bin/phpdoc run
 
-    - stage: test
+    - stage: phar
```
jaapio: moving this to a separate stage doesn't really make any sense I think. build will take longer with this since it cannot start a new stage before all prior stages and jobs are done. Failing tests on a certain phpversion will hide broking things in the phar.

composer.json
```diff
@@ -30,9 +30,9 @@
         "league/pipeline": "^0.3.0",
         "monolog/monolog": "~1.6",
         "padraic/phar-updater": "^1.0",
-        "phpdocumentor/flyfinder": "@beta",
-        "phpdocumentor/graphviz": "~1.0",
-        "phpdocumentor/reflection": "dev-develop@dev",
+        "phpdocumentor/flyfinder": "1.0.0-beta2",
```
jaapio: you want to suffix this with `@beta` So you can use `minimum-stabiltity: stable` this prevents us from using other packages from any alpha or betha phase.

.travis.yml
```diff
@@ -1,48 +1,58 @@
...
+        - travis_retry wget --no-verbose https://scrutinizer-ci.com/ocular.phar && php ocular.phar code-coverage:upload --format=php-clover build/logs/clover.xml
 
     - stage: lint
       php: 7.1
       services:
         - docker
+      before_script:
+        - travis_retry php phive.phar --no-progress install --trust-gpg-keys 8E730BA25823D8B5 phpstan
       script:
+        - ./tools/phpstan analyse src --level max --configuration phpstan.neon
```
jaapio: We are running phpstan at level 2 at the moment. please set that here, and create a new PR to upgrade to max.

.travis.yml
ashnazg: I hadn't realized that separate configs could still be labeled as the same stage.

composer.json
ashnazg: These explicit alphas are temporary, just testing stage stepping stones.  Once phpdoc2 itself proves out, I'll do beta release tags, set them to beta in here, and ultimately we can have them all as final releases.

PR #2004, opened 2018-07-28
Add support of module zend-servicemanager v3

ofry:
Composer package "doctrine/doctrine-orm-module" (Zend Framework Module that provides Doctrine ORM functionality),
 starting with version 2.1.0, depends on zend-servicemanager ^3.3

Signed-off-by: ofry <tim4job@bmail.ru>

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -58,7 +58,7 @@
         "zendframework/zend-i18n": "~2.1",
         "zendframework/zend-serializer": "~2.1",
         "zendframework/zend-servicemanager": "~2.1 || ~3.1",
-        "zendframework/zend-stdlib": "~2.1",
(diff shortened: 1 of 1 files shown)


jaapio:
This PR cannot be merged since the lock file of phpdocumentor isn't changed. Beside that I'm not sure if we use zend-servicemanager somewhere. I think it shouldn't be a dependency of this project at all. 

We should find out why it is a dependency at the beginning first. Phpdocumentor is not fully tested so simply removing it without any tests to check if a removal won't break phpdocumentor isn't safe.

jaapio:
Closing this because I removed the servicemanager from phpdocumentor in #2103

PR #2007, opened 2018-08-10
Add documentation for Twig templating

rmccue:
Hi! I've been working on a custom template for phpDocumentor, and the lack of documentation around the templating system has been killer. It's taken me a while to get up to speed on how all the pieces fit together for the docs, but I figured now that I have, may as well write it down somewhere.

This PR has a start towards filling out the documentation, starting with just the Twig page for now. My reStructuredText skills are a bit rusty, but I checked the syntax here and it should be good.

Happy to contribute more documentation here if this is valuable. I figure the "Element Structure" page was intended for documentation about the various `Descriptor\*Descriptor` classes? If so, I can fill that out, and link it across from here too.


jaapio:
Thanks this is a good starting point for this part of the documentation. If you are able to add some more that will be more than welcome.

Issue #1520, opened 2015-02-28
Move Partials into Reflection Component


mvriel:
In order to Clean up the API for phpDocumentor and make it more intuitive and convenient for contributors should the `Partials` classes be moved into the Descriptor namespace and hence into the Reflection v2 component

mvriel:
No longer relevant

PR #2009, opened 2018-08-10
Store partials internally as full objects

### LINKED ISSUE #2008: Allow generating files for partials
I'd like to be able to generate files for partials, allowing something like:
```xml
<partials>
        <entry>
            <name>contributing</name>
            <link>./CONTRIBUTING.md</link>
        </entry>
        <entry>
            <name>some-other-file</name>
            <link>./some-other-file.md</link>
        </entry>
</partials>
```

Combined with a template with:
```
<transformation
	writer="twig"
	query="partials"
	source="templates/my-template/partial.twig"
	artifact="partials/{{name}}.html"
/>
```

This is almost possible, but `$project->partials` is a Collection of strings, rather than a collection of Partial objects.

## Expected Behavior

`query="partials"` should pass `phpDocumentor\Partials\Partial` objects as the node to the templates.

## Actual Behavior

The query passes `string`s as the node. This then causes the dynamic artifact name of `artifact="partials/{{name}}.html"` to fail, as it attempts to find the `string` property of a string.

1. Add partial to configuration
2. Add transformation with `query="partials"` and dynamic `artifact="partials/{{name}}.html"`

----

This could be solved by moving the content generation step from `Partials\ServiceProvider->register()` into a new `Partials\Partial->toString()`, and storing the actual Partial objects in the collection.

rmccue:
Fixes #2008 by storing partials as full objects internally rather than rendered strings.

In the process, removes the need for the custom PartialCollection. This brings partials more in line with regular Descriptors (see #1520).

Having partials render the Markdown themselves feels a bit weird; given 3.x can break APIs, maybe this should be changed to return the raw content instead? Templates could then use `{{ project.partials.introduction|markdown }}` instead.


jaapio:
Thanks for your contribution.
Could you please rebase your branch to the latest version in develop so we can get this build passing?

About the bc break, we want to avoid a bc break in templates for now. Because we want to move the users of phpdocumentor as soon as possible to v3.

rmccue:
This branch is already up-to-date with the latest `develop`; https://github.com/phpDocumentor/phpDocumentor2/commit/c74af84cedcbf94a7684023e53432661d89032e1 itself is failing.

Makes sense re: BC break.

PR #2014, opened 2018-08-29
Responsive template update

anmccall85:
This is in response to the following issue:
https://github.com/phpDocumentor/phpDocumentor2/issues/578

Please let me know if I am missing anything or if I have made any mistakes.


jaapio:
I'm sorry that it took so long to respond on your PR. Due to a limited time of the development team we decided to drop te support of phpdocumentor v2. v2 is very outdated. And couldn't support the new features of php for example. 

Due to the same time limitations we had to cut in the number of supported templates. Since the xml template were quite intensive to maintain we have decided to drop the support of xml templates in v3. For this PR this means that it will not be merged. 

The new default template in phpdoc v3 - RC is an improved template including all kind of new fancy features to improve you developer experience. We think that the new template will be more convenient for people to change to there own standards. 

Thanks for taking the time to contribute to phpDocumentor. But as hard as it can be PR's can be rejected. I do hope that you will keep using phpDocumentor to improve the docs of your projects.

PR #2015, opened 2018-08-29
Support nested vendor-dir, fixing #2002

### LINKED ISSUE #2002: Bootstrap.php cannot find path to composer.json to get autoload.php
Bootstrap.php cannot find path to composer.json caused on line :138 in Bootstrap.php

## Expected Behavior

Generating documentation via lib/vendor/bin/phpdoc -p -d crud_core/ -t doc/

## Actual Behavior

> lib/vendor/bin/phpdoc -p -d crud_core/ -t doc/
PHP Fatal error:  Uncaught RuntimeException: phpDocumentor expected to find an autoloader at "/home/mr/devel/git/crud_core/lib/vendor/phpdocumentor/phpdocumentor/src/phpDocumentor/../../vendor/autoload.php" but it was not there. Usually this is because the "composer install" command has not been ran yet. If this is not the case, please open an issue at http://github.com/phpDocumentor/phpDocumentor2 detailing what installation method you used, which path is mentioned in this error message and any other relevant information. in /home/mr/devel/git/crud_core/lib/vendor/phpdocumentor/phpdocumentor/src/phpDocumentor/Bootstrap.php:101
Stack trace:
#0 /home/mr/devel/git/crud_core/lib/vendor/phpdocumentor/phpdocumentor/src/phpDocumentor/Bootstrap.php(60): phpDocumentor\Bootstrap->createAutoloader('/home/mr/devel/...')
#1 /home/mr/devel/git/crud_core/lib/vendor/phpdocumentor/phpdocumentor/bin/phpdoc(23): phpDocumentor\Bootstrap->initialize()
#2 {main}
  thrown in /home/mr/devel/git/crud_core/lib/vendor/phpdocumentor/phpdocumentor/src/phpDocumentor/Bootstrap.php on line 101
Script lib/vendor/bin/phpdoc -p -d crud_core/ -t doc/ handling the cdoc event returned with error code 255

  1. change vendor-dir in composer configuration to lib/vendor for example
  2. composer install
  3. php lib/vendor/bin/phpdoc OR make script in composer.json and run composer phpdoc

## Your environment

Rather than doing something like on line :131 use for example getenv('DOCUMENT_ROOT').
Everytime user run script in location where composer.json is this cause that all will work fine and you can have vendor everywhere. Please try it and apologize me and my English. This is my first bug report.

project_folder ->
  [here i am running the script via composer cdoc]
  - composer.json
  - public_html
  - lib
    - vendor

Include as many relevant details about the environment you experienced the bug in and how to reproduce it.
Since many issue might only occur on a specific code base it would be very helpfull if you link to your project
or add any code that is involved in to the issue.

* Version used: 2.9.0
* Install method: composer
* php version 7.0.1
* Operating system and version (e.g. Ubuntu 16.04, Windows 7): Debian Stretch (actual)
* Link to your project:
* ...
* ...

asmundstavdahl:
Fix #2002 by improving Bootstrap::findVendorPath to handle nested vendor-dir in composer.json.

### DIFF
--- a/src/phpDocumentor/Bootstrap.php
+++ b/src/phpDocumentor/Bootstrap.php
@@ -154,7 +154,7 @@ class Bootstrap
                     . " (vendor-dir should be at {$standardRootDir}/{$relativeVendorDir})"
                 );
             }
-        } else if (is_dir($phpDocumentorVendorDir)) {
+        } elseif (is_dir($phpDocumentorVendorDir)) {
             // e.g. /path/to/clone/of/phpDocumentor2/vendor
             return $phpDocumentorVendorDir;
         } else { // Look for a composer.json in shallower paths
--- a/src/phpDocumentor/Bootstrap.php
+++ b/src/phpDocumentor/Bootstrap.php
@@ -170,7 +170,9 @@ class Bootstrap
 
                     return $rootCandidate . '/' . $relativeVendorDir;
                 }
-            } while (is_dir($rootCandidate));
+            } while (is_dir($rootCandidate)
+                && substr_count($rootCandidate, '/..') <= substr_count($rootCandidate, '/') / 2
+            );
         }
 
         return null;
--- a/src/phpDocumentor/Bootstrap.php
+++ b/src/phpDocumentor/Bootstrap.php
@@ -113,36 +113,70 @@ class Bootstrap
     /**
      * Attempts to find the location of the vendor folder.
      *
-     * This method tries to check for a composer.json in a directory 5 levels below the folder of this Bootstrap file.
+     * This method tries to check for a composer.json in a directory 5 levels shallower than the folder of this file.
      * This is the expected location if phpDocumentor is installed using composer because the current directory for
      * this file is expected to be 'vendor/phpdocumentor/phpdocumentor/src/phpDocumentor'.
+     *
+     * If a composer.json is not found in the aforementioned directory, we search in gradually shallower directories
+     * until we find a composer.json or reach the filesystem root.
      *
      * If a composer.json is found we will try to extract the vendor folder name using the 'vendor-dir' configuration
      * option of composer or assume it is vendor if that option is not set.
      *
+     * If no custom composer.json can be found, then we assume that the vendor folder is that of phpDocumentor itself.
      *
-     * If no custom composer.json can be found, then we assume that the vendor folder is that of phpDocumentor itself,
-     * which is `../../vendor` starting from this folder.
-     *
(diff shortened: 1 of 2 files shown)


jaapio:
Could your please add testcases to cover this new features.

asmundstavdahl:
Hold on more testing incoming

src/phpDocumentor/Bootstrap.php
```diff
@@ -170,7 +170,9 @@ public function findVendorPath($baseDir = __DIR__)
 
                     return $rootCandidate . '/' . $relativeVendorDir;
                 }
-            } while (is_dir($rootCandidate));
+            } while (is_dir($rootCandidate)
+                && substr_count($rootCandidate, '/..') <= substr_count($rootCandidate, '/') / 2
```
asmundstavdahl: This check is needed because, at least during Bootstrap::testFindVendorPathComposerInstalledWithNestedVendorDirCanFail, is_dir returns true for paths that are shallower than the root; e.g. is_dir("vfs://root/../..") == true

asmundstavdahl:
IMHO I think it's good to go but I would appreciate any kind of feedback.

jclaveau:
I didn't see your PR when a prepared a fix and realized we have two different approaches: https://github.com/phpDocumentor/phpDocumentor/pull/2153 .

IMHO, looking for `composer.json` is unnecessary and unreliable: when PHPDoc is installed in `/path/to/lib/vendor/phpdocumentor/phpdocumentor` so `AutoloaderLocator.php` is in `/path/to/lib/vendor/phpDocumentor/phpDocumentor/src/phpDocumentor` the only thing we are sure of is the presence of an autoload.php file in the vendor dir.

The vendor dir can have a different name due to `vendor-dir` in composer.json or COMPOSER_VENDOR_DIR environment var, the composer.json can also be named differently due to the COMPOSER environment var (https://getcomposer.org/doc/03-cli.md#composer) and this is un detectable (it changes the composer.lock file accordingly).

So the only sure thing is that `autoload.php` MUST be "4 folders above" if PHPDoc is installed as a dependency.

If PHPDoc not is installed as a dependency (for dev purpose or as a standalone tool), the folder is `/path/to/phdoc/vendor/` and we have the control on it so no change is required.

mvriel:
I agree with @jclaveau that his approach gives a more simple approach to solving the issue by checking the autoload.php instead of composer.json. 

I always find it hard to reject a PR because I appreciate the effort of the contributor so much; so my apologies to @asmundstavdahl. If you believe the other approach to not address a use-case that you foresee, please do let us know

PR #2016, opened 2018-09-04
Use cache folder in config for Twig

mvriel:
Prior to this change the Twig cache files were stored in the system's temp
directory. Although this works in many cases, it may lead to issues when
different projects have different twig files.

This change includes two things:

1. Change pipelines to read stages using tags
2. Introduce a new stage that sets the Twig cache folder to use a subfolder
   of the configured cache folder.

By changing the pipelines we can dynamically introduce a new stage in the
pipeline with a specific priority. This new stage reads the configuration
and preconfigures the twig service.

### DIFF
--- a/src/phpDocumentor/Plugin/Twig/Stage/ConfigureTwig.php
+++ b/src/phpDocumentor/Plugin/Twig/Stage/ConfigureTwig.php
@@ -1,0 +1,27 @@
+<?php
+
+declare(strict_types=1);
+
+namespace phpDocumentor\Plugin\Twig\Stage;
+
+use Twig\Environment;
+
+final class ConfigureTwig
+{
+    /** @var Environment */
+    private $twig;
+
+    public function __construct(Environment $twig)
+    {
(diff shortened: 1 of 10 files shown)


mvriel:
Fixes https://github.com/phpDocumentor/phpDocumentor2/issues/1605

src/phpDocumentor/Plugin/Twig/Stage/ConfigureTwig.php
```diff
@@ -0,0 +1,27 @@
+<?php
+
+declare(strict_types=1);
+
+namespace phpDocumentor\Plugin\Twig\Stage;
+
+use Twig\Environment;
```
jaapio: Is this the right type? Since the twig writer expects a `twig_environment`?

mvriel: Twig_Environment and Twig\Environment are the same classes :) One is an alias

PR #2027, opened 2018-10-23
Fix the functions/methods signatures

### LINKED ISSUE #2020: Unable to parse scalar return type declarations
`v3.0.0-alpha.2-nightly-gdff5545/phpDocumentor.phar` is unable to parse scalar return type declarations in interfaces and traits, I suspect the behaviour is the same in classes but haven't had the time to test it.

`interface IMyInterface {
  public function MyFunction( ) : bool;
}`

`trait TMyTrait {
  public function MyFunction( ) : IMyInterface {
    return $this;
  }
}`

The parse error will say smth amongst the lines of `unexpected : expected ;`

Not much more to be said..

letrunghieu:
After version v3.0.0-alpha.3 was release, there is still a bug on displaying scalar return type. Not nullable types (`int`, `bool`, ...) is missing from the output HTML while the nullable types (`?int`, `?bool`, ...) is displayed incorrectly as `int`, `bool`, ...

For example, this is a test input file:
```php
<?php

class Foo {
	public function getInt(): int {return 1;}
	public function getOptionalInt(): ?int {return null;}
}
```

With the v3.0.0-alpha.3 release, the output document is:
screenshot_2018-10-23 my docs

With this PR, the output document is
screenshot_2018-10-23 my docs 1

This PR also fixes #2020

### DIFF
--- a/data/templates/clean/elements/method.html.twig
+++ b/data/templates/clean/elements/method.html.twig
@@ -5,7 +5,7 @@
             <article class="method">
                 <h3 class="{{ method.visibility }} {% if method.deprecated %}deprecated{% endif %}">{{ method.name }}()</h3>
                 <a href="#source-view" role="button" class="pull-right btn" data-toggle="modal" style="font-size: 1.1em; padding: 9px 14px"><i class="icon-code"></i></a>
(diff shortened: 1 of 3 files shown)


jaapio:
Thanks for this PR. Before I can merge this I need to update the other templates and add a note for other template developers that the need to update their templates.

jaapio:
I checked the code that should provided the types. But I think we discovered a bc break here. 
Thanks for pointing us on this break. We fixed it in the code so templates don't need changes right now. 

The current version in develop should not have this issue any more.

PR #2038, opened 2018-12-01
Fix crash with ini_set when opcache is enabled on CLI

TysonAndre:
The error was:
FatalThrowableError: ini_set() expects parameter 2 to be string, int given

This happened during the cache:clear step.

This bug is not on the `master` branch, which doesn't use strict_types


jaapio:
Thanks

Issue #552, opened 2012-07-31
@param Optional argument detection fails


0815Neo:
The documentor should detect optional arguments automaticly. But when i write this code:
public function __construct( $Host, $User, $UserPassword, $Database, $Persistent = false, $ConnectionCharset = self::STD_CONNECTION_CHARSET, $BufferSize = self::STD_BUFFER_SIZE ) { ...

The documentation is:
__construct(string $Host, string $User, string $UserPassword, string $Database, boolean $Persistent, string $ConnectionCharset, integer $BufferSize) 

But should be:
__construct(string $Host, string $User, string $UserPassword, string $Database, [ boolean $Persistent, string $ConnectionCharset, integer $BufferSize ])

brutalcrozt:
I really wait this one.

jaapio:
You could try to implement it yourself. There is no phpcode needed. And can be done purely in the template.
You can check if a param has a default value. Then render the bracket in the template.

brutalcrozt:
@jaapio that's a good idea, but i think it would be great if we as community follow the notion on the official php documentation. My hope that is become our base template, i am aware that probably not phpdoc2 user will like it, if that becomes an issue to you, we can make it configurable right?

The fact that not only me that involved in this issue, but it also saves to assume that it was necessary.
If this issue is labeled with pr wanted, i am willing to add dig up and code.

But i am still learning tho, not an expert so maybe will take some time.

jaapio:
Hi @brutalcrozt 

I think it makes a lot of sense to have a differentiation between required and optional params. So I don't think any of our users will have any issues with that. Beside that we do support custom templates so if they want to use something different they might just create one themselves.

I don't want to have to many configuration issues specialized on what is in the templates. Currently we don't have a config at template level and adding something like this in the overall config will look very odd. 

So please go a head, I am available to assist you if needed. You can create a new issue or just a PR so we can discuss the details when you are stuck. My own time is very limited so I don't mind if it takes weeks for you to finish this. On the other hand this issue is open for more than 7 years so I don't think any one is waiting for you to finish it. Please take the time you need. And don't hesitate to contact us when you have any questions.

PR #2102, opened 2019-07-30
Template, support optional argument

brutalcrozt:
- kill #552
- optional argument will be rendered between square bracket
- add literal `(optional)` on optional argument


brutalcrozt:
Of the topic,
1. is that the optional argument will always be placed after the mandatory one?
2. it changes the clean template, does it change the default template?
Also it untested, please guide me to properly testing it (if it necessary)

@jaapio , i think we agreed that we are going to changes the default template, right?

jaapio:
Thanks for this PR,

In php terms an optional argument MUST be placed after all mandatory ones. You cannot simply omit a mandatory argument after an optional. `function fooBar($optional = 'default', $required) {}` cannot be called without errors. 

The clean template is the default one.
At this moment we don't have any tests for templates since they are quite hard to write. There are a few of them but they are not working properly. 

Could you please rebase this PR so only your changes are included?

brutalcrozt:
Sure, captain @jaapio .
It would be a shame if the template doesn't change as we expect.
what i intent is manually test these changes since I don't have many experiences on composer.
I want to test it against my project. before it merged.
But i have no idea how to do it, i never release a project, 
but really excited to know how the process involved.

PR #2104, opened 2019-08-01
Template, support optional argument

brutalcrozt:
- kill #552
- optional argument will be rendered between square bracket
- add literal (optional) on optional argument

### DIFF
--- a/data/templates/clean/elements/method.html.twig
+++ b/data/templates/clean/elements/method.html.twig
@@ -5,7 +5,22 @@
             <article class="method">
                 <h3 class="{{ method.visibility }} {% if method.deprecated %}deprecated{% endif %}">{{ method.name }}()</h3>
                 <a href="#source-view" role="button" class="pull-right btn" data-toggle="modal" style="font-size: 1.1em; padding: 9px 14px"><i class="icon-code"></i></a>
-                <pre class="signature" style="margin-right: 54px;">{{ method.name }}({% for argument in method.arguments %}{{ argument.types ? argument.types|join('|')~' ' }} <span class="argument">{{ argument.isVariadic ? '...' }}{{ argument.name }}{{ argument.default ? ' = '~argument.default }}</span>{% if not loop.last %}, {% endif %}{% endfor %}) {{ method.response.types ? ': '~method.response.types|join('|') }}</pre>
+                <pre class="signature" style="margin-right: 54px;">{{ method.name }}
+                  ({% for argument in method.arguments %}
+                    {{ argument.types ? argument.types|join('|')~' ' }}
+                    {% if argument.default|length > 0 %}
+                      [
+                    {% endif %}
+                    <span class="argument">{{ argument.isVariadic ? '...' }}
+                      {{ argument.name }}{{ argument.default ? ' = '~argument.default }}
+                    </span>
+                    {% if argument.default|length > 0 %}
+                      ]
+                    {% endif %}
+                    {% if not loop.last %}, {% endif %}
+                  {% endfor %})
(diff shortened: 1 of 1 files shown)


brutalcrozt:
I end up failed to rebase properly, this is the new one.

@jaapio 
It would be a shame if the template doesn't change as we expect.
what i intent is manually test these changes since I don't have many experiences on composer.
I want to test it against my project. before it merged.
But i have no idea how to do it, i never release a project,
but really excited to know how the process involved.
Did you mind to assist me?

jaapio:
There are multiple ways to test your changes. 

First thing you could do is just make local changes in the vendor directory in your project where you required phpdocumentor. But since this will not run the actual pushed code you probably don't want to do it always like this.

A second way to achieve this it to create a local installation of phpdocumentor. Which is just a clone of this repo. And do a composer install. in this case you can run phpdocumentor from that directory.

There are other ways to do it. but I think these are the most easy ones to do.

brutalcrozt:
Kinda expecting we can install like [this](https://github.com/RobinHerbots/Inputmask#install-the-latest-beta-version)

FYI, try to run from local it yield `unable to find the `dot` command th`. Then the string is missing. Similar to  [this](https://stackoverflow.com/questions/23655520/phpdocumentor-unable-to-find-the-dot-command-of-the-graphviz-package/49630588), but without Graphviz on the error message. Sometimes the error is gone, but  I am having issues that the changes on template not real-time.
I am messing up with the white space, trying to fix it back, but either i don't know or twig not support multiline formatting. Also, we use twig 1.x, they have newer one 2.0.
do you think it's good to also update twig?

brutalcrozt:
Sorry to ping you @jaapio , but i need your response.
I can fix the twig related issue, i need your comment about elaborating the beta version or something like that on our lovely phpDocumentor2.

jaapio:
We cannot update twig at this moment.
I think you expirence issues with the twig cache. I need to check if you can reset it with a command. But you can always remove the directory. If you use the latest stable release it is in /tmp. Otherwise it should be in the cache directory which you can specify on the command line

brutalcrozt:
while I am aiming like the official php documentation like so:
image
currently, the changes i made produce:
image

I am feeling bad if i continue [this line](https://github.com/phpDocumentor/phpDocumentor2/blob/develop/data/templates/clean/elements/method.html.twig#L8) to be unreadable, the line length is already too long.

I rather add more line on our structure.xml, add optional_argument element.
below the argument element:
image
But sadly i don't know how to do that, I dig up on our documentation [project:parse](https://docs.phpdoc.org/references/commands/project_parse.html) and [project:transform](https://docs.phpdoc.org/references/commands/project_transform.html) but it not ready yet.
My biggest concern is related to code readability, i am waiting for your thought about this.

P.S
I am willing to make another PR related to our documentation (with your assistant) if needed.

mvriel:
Hi @brutalcrozt, adding a new flag like that to the structure feels a bit redundant to me as the 'default' indicates whether it is an optional argument or not. In your PR you have added newlines in the signature and because of the `<pre>` this may cause weird artefacts such as intended spacing, etc.

I like the way you are approaching this by mimicking the PHP documentation's style, and I can only encourage you to continue on this path.

As for the twig caching; if you add a parameter `twig-debug` to your template.xml for Clean then it will drop twig into a debugging mode. As I see now, it would be nice to also do this using a command-line argument of some sort but that does not exist now.

mvriel:
Soon, probably in the next couple of hours, I will be merging develop into master and removing the develop branch. I will post an update here but when that happens, can you change the target branch of this PR to point to master? I am unable to do this and I don't want to close this PR :)

PR #2118, opened 2019-08-17
Show dollar sign before argument name.

ADmad:
It feels alien seeing argument names without it.


mvriel:
Thank you for your contribution! I have merged your PR and I hope that you consider contributing more in the future

PR #2120, opened 2019-08-23
fix schema path in config/services.yaml

ptomulik:
Without this fix phpDocumentor is looking for ``data/xsd/phpdoc.xsd`` in the working directory. The file, in fact, is located in phpdoc's root directory.


jaapio:
Thanks

PR #2124, opened 2019-08-26
Fix loading V3 config from default locations

ptomulik:
A V3 config file containing errors is currently silently ignored, and either a default configuration is used or a next (possibly V2) file if exists. This seems to be quite counter-intuitive. An existing config file should rather be reported to contain errors, if it is the case. This PR provides a fix for this.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,6 +22,20 @@ var/*
 *.phar
 screenshots
 
+# IDEs and editors
+/.idea
+.project
+.classpath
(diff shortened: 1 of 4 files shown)


jaapio:
Please add some tests to show the need for these changes. And to be sure that we won't break this behavior in the future.

ptomulik:
Done

jaapio:
thanks

PR #2129, opened 2019-10-30
Template markdown

devcorrelator:
I created my own template to allow publishing of documentation on the Gitlab and Github wiki pages that go with projects.

Right now phpDocumentor will only create html files, and it needs to be able to create md files, maybe define it in the template.xml file?

Your project is so huge, and I'm not familiar enough with it to submit a pull request for allowing it to generate files with other extensions.

Here is my project that this template is based on.
[https://github.com/fr3nch13/phpdoc-markdown](https://github.com/fr3nch13/phpdoc-markdown)

Hopefully this can start a conversation about alternate formats like markdown?


mvriel:
Hi @devcorrelator, 

At this moment, our team does not have the development capacity to actively support more formats. We are open to such things because we do see the added value, but any PR we accept into the main repository also means we need to be able to support it from thereon.

I am working on a way to make it easier for community-supported contributions to be available in phpDocumentor, but this is not yet done.

Until that moment, I cannot accept this PR but I think it would make a splendid addition to community-managed templates and extensions

devcorrelator:
Thanks. I totally understand! I’ll keep an eye out for when community templates are supported, and adjust to those rules.

oliveratgithub:
I started a GitHub Wiki markdown template for phpDocumentor here: [oliveratgithub phpDocumentor-Template-ghwiki](https://github.com/oliveratgithub/phpDocumentor-Template-ghwiki/blob/master/README.md)

mvriel:
Doing the hard thing here today .. even though I very much appreciate the contribution, we as a team do not have the capacity to support this template in the past few years and as I can see it, not in the upcoming period.

Merging this in our codebase would mean it is 'official' and we need to update it to stay relative or to include new features in upcoming PHP versions.

Although I hate doing this, I am going to close this PR because we just cannot support this as part of maintenance. I can best recommend making a canonical repository for this template and we can talk about having a section on the website where we can list community-contributed templates, because I do appreciate the effort!

PR #2137, opened 2019-11-25
Remove XML Writer and Supporting classes

mvriel:


docs/references/writers/index.rst
```diff
@@ -7,4 +7,3 @@ Writers
    fileio
    twig
    checkstyle
```
jaapio: checkstyle was based on xml isn't it? So this should be removed

mvriel: I thought I checked that yesterday and checkstyle is it's own thing. Can't hurt to verify during my lunch break

PR #2141, opened 2019-11-29
Tighten up coding standards

mvriel:
In this change I have tightened the phpcs.xml, ran phpcbf and started
processing the remaining feedback by hand. Especially the tightening of
type checks is causing issues when running the application and tests.
This will potentially resolve a fair number of bugs or unintended
effects

This PR is not ready to be merged as the running of application still fails; but I needed a break


mvriel:
This PR introduces a lot of code style changes; I think there are a few left but these are quite a few important ones and the unit tests and local runs are green again

src/phpDocumentor/Compiler/Linker/Linker.php
```diff
@@ -157,22 +167,24 @@ public function substitute($item, $container = null)
...
 
                 // if the returned response is not an object it must be grafted on the calling object
-                if ($response !== null) {
-                    $setter = 'set' . ucfirst($fieldName);
-                    $item->{$setter}($response);
+                if ($response === null) {
+                    continue;
                 }
+
+                $setter = 'set' . ucfirst($fieldName);
```
jaapio: This hurts, needs to be changed at some point. But I need to think about how, please add a note that we fix this. in the future.

src/phpDocumentor/Compiler/Linker/Linker.php
```diff
@@ -267,17 +276,17 @@ protected function isDescriptorContainer($item): bool
...
      */
-    protected function replacePseudoTypes(string $fqsen, $container): string
+    protected function replacePseudoTypes(string $fqsen, ?DescriptorAbstract $container) : string
     {
         $pseudoTypes = ['self', '$this'];
         foreach ($pseudoTypes as $pseudoType) {
-            if ((strpos($fqsen, $pseudoType . '::') === 0 || $fqsen === $pseudoType) && $container) {
-                $fqsen = $container->getFullyQualifiedStructuralElementName()
-                    . substr($fqsen, strlen($pseudoType));
+            if ((strpos($fqsen, $pseudoType . '::') !== 0 && $fqsen !== $pseudoType) || !$container) {
```
jaapio: If we remove the call with a `null` value for `$container` we could remove the or here.

mvriel: That requirement for a null value comes from deeper in the Linker; I do not dare change that right now :D I have added a todo to revisit this

mvriel: To be fair, there are quite a few places where stuff is nullable that would be nice if it were not. But I prefer not to fix that at once but instead go over the codebase as soon as we have passing stylechecks and improve those locations one by one. Same goes for mixed types that we have in places

src/phpDocumentor/Descriptor/ConstantDescriptor.php
```diff
@@ -52,13 +52,13 @@ public function setParent($parent)
             ? $parent->getFullyQualifiedStructuralElementName() . '::' . $this->getName()
             : $this->getName();
 
-        $this->setFullyQualifiedStructuralElementName($fqsen);
+        $this->setFullyQualifiedStructuralElementName(new Fqsen($fqsen));
```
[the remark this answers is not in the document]
mvriel: All `*etFullyQualifiedStructuralElementName()` methods now yield and accept Fqsen objects; apparently this was already partly the case in many locations but with the added type-hinting it became painfully obvious. This went well in the past because without declare strict, the Fqsen was converted into a string

src/phpDocumentor/Transformer/Router/Router.php
```diff
@@ -39,86 +36,82 @@
...
         $this->configure();
     }
 
     /**
      * Configuration function to add routing rules to a router.
      */
-    public function configure()
+    public function configure() : void
     {
-        // Here we cheat! If a string element is passed to this rule then we try to transform it into a Descriptor
```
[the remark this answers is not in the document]
mvriel: Apparently, it was not necessary because of how the linker properly converts everything to an object and this was causing issues with type-hinting. I am still working on the routing and will probably reintroduce something that can route based on a string again in the future

src/phpDocumentor/Descriptor/Builder/Reflector/ArgumentAssembler.php
```diff
@@ -27,12 +25,10 @@ class ArgumentAssembler extends AssemblerAbstract
     /**
      * Creates a Descriptor from the provided data.
      *
-     * @param Argument $data
+     * @param mixed $data
```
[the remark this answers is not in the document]
mvriel: I agree :thinking: I have made it more explicit

PR #2153, opened 2019-12-06
More robust vendor directory detection

jclaveau:
This PR detects the vendor directory based on the presence of autoload.php. This allows the use of COMPOSER_VENDOR_DIR environment variable of Composer when installed as dependency


src/phpDocumentor/AutoloaderLocator.php
```diff
@@ -32,40 +32,31 @@ public static function autoload() : ClassLoader
...
-        $composerConfigurationPath = $rootFolderWhenInstalledWithComposer . 'composer.json';
-        if (file_exists($composerConfigurationPath)) {
-            $vendorDir = $rootFolderWhenInstalledWithComposer
-                . self::getCustomVendorPathFromComposer($composerConfigurationPath);
+        $vendorFolderWhenInstalledWithComposer = $baseDir . '/../../../../';
+        if (file_exists($vendorFolderWhenInstalledWithComposer.'/autoload.php')) {
+            $vendorDir = $vendorFolderWhenInstalledWithComposer;
         }
+        
+        $vendorDir = realpath($vendorDir);
```
mvriel: The use of realpath may interfere with running phpDocumentor's PHAR distribution. Realpath will break out of a PHAR file and consult the host filesystem. This is the reason why the original code did not use realpath to clean up the path :)

tests/unit/phpDocumentor/AutoloaderLocatorTest.php
```diff
@@ -84,32 +84,72 @@ public function testAutoloadAtDefaultLocation() : void
...
+
+/**
+ * This function overrides the native realpath($url) function, removing
+ * all the "..", ".", "///" of an url. Contrary to the native one, 
+ * 
+ * @see https://github.com/bovigo/vfsStream/issues/207
+ * @param string $url
+ * @param string|bool The cleaned url or false if it doesn't exist
+ */
+function realpath(string $url)
```
mvriel: This is unnecessary since realpath should not be used; the vfs filesystem has the same issues as the phar filesystem; with realpath you break out of virtual filesystem

mvriel:
Can you rebase master into this PR? Master contains a more robust build setup that will also test things like the PHAR distribution

src/phpDocumentor/AutoloaderLocator.php
jclaveau: I didn't know that, thanks. I just added realpath to ease the debug with more readability.

Thanks a lot

mvriel: No problem, I know this because I did the exact same thing when I originally built this class ;)

mvriel: I see you rebased the PR but the realpath is still there; do you still have to make the change or did something get lost in the rebase?

jclaveau: I just wanted to see the failing tests with realpath rebased on master (because your approach for integration tests interests me) but Behat's tests are passing loacally (`80 scenarios (80 passed)
283 steps (283 passed)`).

I'm surprised they are not run automatically with this PR and I get a warning/error in the Github's actions of my commit: 
+ https://github.com/jclaveau/phpDocumentor/runs/366043339
+ https://github.com/jclaveau/phpDocumentor/runs/366043344

So I wonder if 
+ Behat's is wright locally but wrong in Github's actions?
+ The opposit.
+ The tests you speak about are still not merged in master?

jaapio: The failing build you are seeing is using the phar which has been built in a step before we execute behat. As @mvriel already said your changes are breaking the phar and not the plain build which is also passing in your linked build.

Does that make it more clear what is going on? 

We are still searching for a solution to build the pr on this project as well. So we can see the results in the pr. Main issue is that we don't want to build on pr and push but only one of them

jaapio: The failing build you are seeing is using the phar which has been built in a step before we execute behat. As @mvriel already said your changes are breaking the phar and not the plain build which is also passing in your linked build.

Does that make it more clear what is going on? 

We are still searching for a solution to build the pr on this project as well. So we can see the results in the pr. Main issue is that we don't want to build on pr and push but only one of them

jclaveau:
@mvriel @jaapio I just removed the realpath and added the support of COMPOSER_VENDOR_DIR and COMPOSER environment variable with related tests

src/phpDocumentor/AutoloaderLocator.php
jclaveau: Thanks for the explanations. I removed the realpath() call

jaapio:
Thanks!

jclaveau:
Thanks for the review and all your work!

PR #2180, opened 2019-12-24
[WIP]: Add support for extensions

mvriel:
For phpDocumentor, we want to allow consumers and internal processes to
be able to manipulate what data will be rendered and with it the ability
to change what data is included.

Examples of this is the `@mixin` tag, with an extension you can read the
tags of all classes, determine whether it has a mixin tag and try to
locate an accompanying class that needs to be mixed in. Once found, you
can change the list of methods to include the mixed in methods; their
parent property remains a method to determine that they do not originate
in this class but instead from elsewhere.

This extension mechanism should also work when called from phpDocumentor
as a PHAR. Because you can register additional autoloading mechanisms
on runtime, it is able to query the real filesystem instead of inside
the PHAR. This means that projects could include their own extensions
and still refer to them from within the PHAR.

If this mechanism sticks, we could even attempt to move internal
processes (such as inheritence) to these extensions.

### DIFF
--- a/config/services.yaml
+++ b/config/services.yaml
@@ -144,6 +144,15 @@ services:
     phpDocumentor\Compiler\Linker\Linker:
         arguments: ['%linker.substitutions%']
 
+    # The locations and extensions are now coded into this service; but this should be injected using a pipeline
+    # stage with the registerExtension and registerLocation methods.
+    phpDocumentor\Extensions:
+        arguments:
+            $extensions:
+                - 'phpDocumentor\Extensions\Example'
+            $locations:
+                'phpDocumentor\Extensions\': 'data/Extensions'
+
     phpDocumentor\Compiler\Compiler:
         calls:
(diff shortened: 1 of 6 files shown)


mvriel:
This PR is a work in progress; the foundation is there, ready to be reviewed but the configuration should be adapted to allow for registering extension; and this should be carefully tested using an actual phar distribution and with behat tests

mvriel:
I agree that it does add a lot of responsibility to the user, I am not too concerned about that specific part as extensions are meant as a powerful measure and perhaps a bit as a last resort. Preventing deletions will restrict in a way for which I am not sure whether that will do much good.

Another solution may be to add a separation between the extension and the project descriptor; maybe introduce an API / Builder that has methods with which you can influence the project (but this also runs the risk of limiting the user in unforeseen ways or adding more complexity to maintain) or serialize the ProjectDescriptor to a data structure that is meant for manipulation and pass that to the writers. Comparable to the original 'Views' feature that we had planned.

mvriel:
The experiment was a success so far but we do not have the resources to pursue this at this moment; I will keep the branch open but close the PR for now

PR #2182, opened 2019-12-25
Introduce FlySystem into twig file loading

mvriel:
We want to load our twig templates using FlySystem's filesystems because
that enables us to more easily deal with templates inside and outside of
phpDocumentor's PHAR file and will simplify file handling.

TODO:

- [x] Unit tests
- [ ] Behat tests
- [x] Cleanup

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,7 +33,7 @@ phpstan:
 .PHONY: test
 test:
 	docker-compose run --rm phpunit ${ARGS}
-	docker-compose run --entrypoint=/usr/local/bin/php --rm phpunit tests/coverage-checker.php 68
+	docker-compose run --entrypoint=/usr/local/bin/php --rm phpunit tests/coverage-checker.php 69
 
 .PHONY: behat
 behat:
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -11,7 +11,6 @@ parameters:
     - '#.*NodeDefinition::fixXmlConfig.*#'
     - '#.*NodeDefinition::addDefaultsIfNotSet.*#'
 
-    - '#Constructor of class phpDocumentor\\Transformer\\Writer\\Twig\\Extension has an unused parameter \$transformation#'
     - '#Call to an undefined method League\\Flysystem\\Filesystem::find()#'
     # Bad design of descriptors.
     - '#Access to an undefined property phpDocumentor\\Descriptor\\Collection::\$[a-z]+#'
--- a/src/phpDocumentor/Transformer/Template.php
+++ b/src/phpDocumentor/Transformer/Template.php
@@ -18,7 +18,7 @@ use ArrayIterator;
 use Countable;
 use InvalidArgumentException;
 use IteratorAggregate;
-use League\Flysystem\FilesystemInterface;
+use League\Flysystem\MountManager;
 use phpDocumentor\Transformer\Template\Parameter;
 use function array_merge;
 use function count;
@@ -50,7 +50,7 @@ final class Template implements ArrayAccess, Countable, IteratorAggregate
     /** @var Parameter[] Global parameters that are passed to each transformation. */
     private $parameters = [];
 
-    /** @var FilesystemInterface */
+    /** @var MountManager */
     private $files;
 
     /**
@@ -58,7 +58,7 @@ final class Template implements ArrayAccess, Countable, IteratorAggregate
      *
      * @param string $name Name for this template.
      */
-    public function __construct(string $name, FilesystemInterface $files)
+    public function __construct(string $name, MountManager $files)
     {
         $this->name = $name;
         $this->files = $files;
@@ -129,7 +129,20 @@ final class Template implements ArrayAccess, Countable, IteratorAggregate
         $this->version = $version;
     }
 
-    public function files() : FilesystemInterface
+    /**
+     * FlySystem filesystem / MountManager containing the template files, base templates files
+     * and destination filesystem.
+     *
+     * This MountManager has three mounts:
+     *
+     * - template://, the files of this template
+     * - templates://, the base folder containing phpDocumentor's global templates (i.e. `/data/templates`)
+     * - destination://, the destination where the template needs to write to
+     *
+     * By combining this in one mount manager it is easier for writers to copy files between destinations (since
+     * MountManager's can copy between filesystems) and for writers to read and write from various locations.
+     */
+    public function files() : MountManager
     {
         return $this->files;
     }
--- a/src/phpDocumentor/Transformer/Writer/Twig/EnvironmentFactory.php
+++ b/src/phpDocumentor/Transformer/Writer/Twig/EnvironmentFactory.php
@@ -13,21 +13,12 @@ declare(strict_types=1);
 
 namespace phpDocumentor\Transformer\Writer\Twig;
 
-use InvalidArgumentException;
 use phpDocumentor\Descriptor\ProjectDescriptor;
-use phpDocumentor\Transformer\Template;
-use phpDocumentor\Transformer\Template\Parameter;
 use phpDocumentor\Transformer\Transformation;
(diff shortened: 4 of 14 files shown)


src/phpDocumentor/Transformer/Writer/Twig/FlySystemLoader.php
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public function getCacheKey($name)
+    {
+        $this->guardTemplateExistsAndIsFile($name);
+
+        return $name;
```
ashnazg: This just returns the unaltered string `$name` that was passed in?  Is the actual effect on this method call to do some state change in the background? At first glance, I thought maybe it should be a `void` return instead.

Or, is some method chaining/composing done in a way that a method wrapped around this method at runtime is also needing `$name`, and that's why it is returned like this?

mvriel: The main effect of this function is the guard; if that throws a loader error than Twig knows the cache key cannot be set. And the template name is a valid cache key in this case :)

phpstan.neon
```diff
@@ -11,7 +11,6 @@ parameters:
     - '#.*NodeDefinition::fixXmlConfig.*#'
     - '#.*NodeDefinition::addDefaultsIfNotSet.*#'
 
-    - '#Constructor of class phpDocumentor\\Transformer\\Writer\\Twig\\Extension has an unused parameter \$transformation#'
```
jaapio: :+1:

Makefile
```diff
@@ -33,7 +33,7 @@ phpstan:
 .PHONY: test
 test:
 	docker-compose run --rm phpunit ${ARGS}
-	docker-compose run --entrypoint=/usr/local/bin/php --rm phpunit tests/coverage-checker.php 68
+	docker-compose run --entrypoint=/usr/local/bin/php --rm phpunit tests/coverage-checker.php 69
```
jaapio: :+1:

src/phpDocumentor/Transformer/Template.php
```diff
@@ -129,7 +129,20 @@ public function setVersion(string $version) : void
...
 
-    public function files() : FilesystemInterface
+    /**
+     * FlySystem filesystem / MountManager containing the template files, base templates files
+     * and destination filesystem.
+     *
+     * This MountManager has three mounts:
+     *
+     * - template://, the files of this template
+     * - templates://, the base folder containing phpDocumentor's global templates (i.e. `/data/templates`)
```
jaapio: I think this is because you need some files from the global templates to be used during the rendering. It would have been nice if we could have this more like a transparent file system. Because you might know which directory you need to use to get the assets you need? So if `template://` doesn't contain the file you need look in `templates://`. And hide this from the actual implementation? So writers only need `template://`
If you see any options to do this that would be nice. but it is also ok to keep it like this for now.

src/phpDocumentor/Transformer/Writer/Twig/EnvironmentFactory.php
```diff
@@ -44,29 +35,18 @@ public function create(
...
-
-        // Clone twig because otherwise we cannot re-set the extensions on this twig environment on every run of this
-        // writer
-        $env = new Environment(new FilesystemLoader($templateFolders));
-
-        $this->addPhpDocumentorExtension($project, $transformation, $destination, $env);
-        $this->addExtensionsFromTemplateConfiguration($transformation, $project, $env);
+        $mountManager = $transformation->template()->files();
+        $env = new Environment(
+            new ChainLoader(
```
jaapio: You are kind of doing the thing I suggested earlier because twig cannot handle multiple filesytems. Would have been nice if it was abstracted here... Maybe our own `FileSystemChain`?

src/phpDocumentor/Transformer/Writer/Twig/Extension.php
```diff
@@ -147,14 +144,14 @@ static function (
             ),
             'sort' => new TwigFilter(
                 'sort_*',
-                static function ($direction, $collection) {
+                static function (string $direction, $collection) {
```
jaapio: can we add the return type on this closure as well?

tests/unit/phpDocumentor/Transformer/TemplateTest.php
```diff
@@ -159,8 +157,8 @@ public function testThatAllParametersArePropagatedToTheTransformationsWhenNeeded
         $this->assertSame(['key' => $parameter], $transformation->getParameters());
     }
 
-    private function givenExampleFilesystem() : Filesystem
+    private function givenExampleMountManager() : MountManager
     {
-        return $this->faker()->fileSystem();
```
jaapio: is `fileSystem` used anywhere now?

src/phpDocumentor/Transformer/Writer/Twig/FlySystemLoader.php
ashnazg: So, should it be a `void` return then?  The fact that it returns `$name` seems significant... and then the next realization that it returns the same `$name` _unaltered_ seems _extremely_ significant.  It's the kind of thing that, if I was in this code troubleshooting something, seeing this return done this way would make me sit up and take notice... and then go digging around calling code to try to understand its rationale, trying to understand why it was _important_ to return the same `$name` value that was passed in as an arg.  If there is so true significance for returning it that way, can it be `void` instead?

tests/unit/phpDocumentor/Transformer/TemplateTest.php
mvriel: Yep :)

src/phpDocumentor/Transformer/Writer/Twig/Extension.php
mvriel: Unfortunately not without looking into this some more; it seems that $collection may not always be a collection. I stayed on the safe side with this one because I didn't dare to :)

src/phpDocumentor/Transformer/Template.php
mvriel: Not entirely; the difference between these two is the difference between `/data/templates` and `/data/templates/clean`. When a path starts with `templates/` it actually refers to the 'global templates path'; where if `templates/` is not in front of it, it is considered to be the template folder itself. So there is no one-on-one overloading happening here.

src/phpDocumentor/Transformer/Writer/Twig/EnvironmentFactory.php
mvriel: As I mentioned, not really. But it does make me wonder if this chain loader is actually incorrect. Although my tests with the previous iteration showed the same paths. A future investigation may be nice :)

And I prefer not to do this earlier because then we mix Twig specific functionality with filesystems; these two are not the same.

For example, you may want to copy specific paths to your destination and for that, it is best if you refer to a specific source filesystem. Otherwise, you may get similar issues to Requests using a generic 'get' where you are not sure whether the value comes from the query, post data or cookies.

src/phpDocumentor/Transformer/Writer/Twig/FlySystemLoader.php
mvriel: This is a requirement of the base Twig Loader; not much I can do there I am afraid

ashnazg: Ah hah, there's the unseen significance... that might be worth an inline comment, or even an `@internal` tag, so that a later refactoring won't see the same thing I did and think "that's not necessary... let's just `void` this instead".

mvriel: Done and pushed to master

PR #2188, opened 2019-12-28
FQSEN objects did not link

mvriel:
The linker converts (arrays of) FQSENs to Descriptors and injects these
into other Descriptors. This is used to restore the object graph after
serialization (object graphs cannot be serialized, only trees).

With an earlier refactoring the FQSEN object was introduced but the
linker was not updated to resolve this type; meaning that anything that
used an FQSEN did not resolve.


mvriel:
In this PR I want to add a couple more readability improvements to the Linker

src/phpDocumentor/Compiler/Linker/DescriptorRepository.php
```diff
@@ -0,0 +1,170 @@
...
+
+    /** @var DescriptorAbstract[] */
+    private $elementList = [];
+
+    /**
+     * Attempts to find a Descriptor object alias with the FQSEN of the element it represents.
+     *
+     * This method will try to fetch an element after normalizing the provided FQSEN. The FQSEN may contain references
+     * (bindings) that can only be resolved during linking (such as `self`) or it may contain a context marker
+     * {@see CONTEXT_MARKER}.
```
jaapio: I think we need to check if this `CONTEXT_MARKER` is still working... because the `see` tag has now the requirement of a `FQSEN` which is always a valid element name. Maybe this is a broken tag :-)

mvriel:
I hope I will be able to make it even more clear in the next few commits; I am slowly working towards something more understandable. Also for me :D

src/phpDocumentor/Compiler/Linker/DescriptorRepository.php
mvriel: I'll write a test when I continue :)

composer-require-config.json
```diff
@@ -3,7 +3,8 @@
     "null", "true", "false",
     "static", "self", "parent",
     "array", "string", "int", "float", "bool", "iterable", "callable", "void", "object",
-    "Composer\\Autoload\\ClassLoader"
+    "Composer\\Autoload\\ClassLoader",
+    "League\\Uri\\Exceptions\\SyntaxError"
```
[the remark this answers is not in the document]
mvriel: If this PR is green I will merge it first and then fix that on master; I don't want to run another build just for this :sweat_smile:

PR #2193, opened 2019-12-30
Setup of resolve in DSN

jaapio:
This patch makes phpdoc using globs to collect files in the DSN of the source and it makes the source dsn relative to the loaded config.

### DIFF
--- a/src/phpDocumentor/Configuration/PathNormalizingMiddelware.php
+++ b/src/phpDocumentor/Configuration/PathNormalizingMiddelware.php
@@ -1,0 +1,111 @@
+<?php
+
+declare(strict_types=1);
+
+namespace phpDocumentor\Configuration;
+
+use League\Uri\Uri;
+use phpDocumentor\Dsn;
+use phpDocumentor\Path;
+use function ltrim;
+use function rtrim;
+use function strpos;
+use function substr;
+
+final class PathNormalizingMiddelware
+{
+    public function __invoke(array $configuration, ?Uri $uri) : array
+    {
+        $configuration = $this->makeDsnRelativeToConfig($configuration, $uri);
+        $configuration['phpdocumentor'] = $this->normalizePaths($configuration['phpdocumentor']);
+
+        return $configuration;
+    }
+
+    private function pathToGlobPattern(string $path) : string
+    {
+        if (strpos($path, '.') === 0) {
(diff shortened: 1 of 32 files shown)


src/phpDocumentor/Configuration/PathNormalizingMiddelware.php
```diff
@@ -0,0 +1,111 @@
...
+
+use League\Uri\Uri;
+use phpDocumentor\Dsn;
+use phpDocumentor\Path;
+use function ltrim;
+use function rtrim;
+use function strpos;
+use function substr;
+
+final class PathNormalizingMiddelware
```
mvriel: Nitpick: typo, should be PathNormalizingMiddleware

src/phpDocumentor/Dsn.php
```diff
@@ -57,9 +62,30 @@ final class Dsn
     /**
      * Initializes the Dsn
      */
-    public function __construct(string $dsn)
+    public function __construct(UriInterface $uri, array $parameters, string $dsn)
     {
-        $this->parse($dsn);
+        $this->dsn = $dsn;
```
mvriel: Isn't this redundant now that `uri` and `parameters` are in the value object?

src/phpDocumentor/Path.php
```diff
@@ -69,4 +72,12 @@ public static function isAbsolutePath(string $file) : bool
             )
             || parse_url($file, PHP_URL_SCHEME) !== null;
     }
+
+    public static function dirname(Path $input) : self
```
mvriel: shouldn't this be a non-static member of Path? It feels somewhat double to have a static method that accepts an instance

tests/unit/phpDocumentor/Configuration/PathNormalizingMiddelwareTest.php
```diff
@@ -0,0 +1,99 @@
...
+
+use League\Uri\Uri;
+use PHPUnit\Framework\TestCase;
+
+/**
+ * @coversDefaultClass \phpDocumentor\Configuration\PathNormalizingMiddelware
+ * @covers ::__invoke
+ * @covers ::<private>
+ */
+final class PathNormalizingMiddelwareTest extends TestCase
```
mvriel: Another nitpick: same sort of typo

PR #2195, opened 2020-01-02
Add initial e2e test for Default template

mvriel:
Because Behat is not able to comfortably test frontends, I have done an
experiment using Cypress as a test runner. This change is the result of
this experiment.

In this change I have introduced a new make command `make test-template`
that will install Cypress using NPM, run phpDocumentor against the
Mario's Pizzeria example, and test the output using Cypress.

I have chosen not to include this in the build pipeline yet as this is
an example meant for feedback; if we continue this way I want to clean
up the make commands a bit more and add it to Github Actions.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,11 +1,3 @@
-# IDE Shizzle; it is recommended to use a global .gitignore for this but since this is an OSS project we want to make
-# it easy to contribute
-.idea
-/nbproject/private/
-.buildpath
-.project
-.settings
-
 # No need to version the binary files of other tools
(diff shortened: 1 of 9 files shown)


tests/features/template/default/general.feature
```diff
@@ -0,0 +1,7 @@
+Feature: documentation index
```
jaapio: I think this was not intended to be included?

mvriel: Ah indeed, that can be removed again

PR #2196, opened 2020-01-03
Target should be relative to the current working dir

jaapio:
When --target is specified on the commandline it should be relative
to the current working dir.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,7 +36,7 @@ test:
 	docker-compose run --entrypoint=/usr/local/bin/php --rm phpunit tests/coverage-checker.php 69
 
 .PHONY: integration-test
-integration-test: node_modules/.bin/cypress data/examples/MariosPizzeria/build/default/index.html data/examples/MariosPizzeria/build/clean/index.html
+integration-test: node_modules/.bin/cypress build/default/index.html build/clean/index.html
 	docker run -it --rm -v ${CURDIR}:/e2e -w /e2e cypress/included:3.2.0
 
 .PHONY: behat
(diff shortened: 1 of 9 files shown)


src/phpDocumentor/Pipeline/Stage/Configure.php
```diff
@@ -50,7 +51,7 @@ public function __construct(
     public function __invoke(array $options) : array
     {
         $this->configFactory->addMiddleware(
-            new CommandlineOptionsMiddleware($options, $this->configFactory)
+            new CommandlineOptionsMiddleware($options, $this->configFactory, getcwd())
```
mvriel: Consider using the Kernel's `getWorkingDir` method; this is essentially a wrapper around getcwd but is useful for service injection using `@=service('kernel').getWorkingDir()`

Makefile
```diff
@@ -53,8 +53,8 @@ shell:
 node_modules/.bin/cypress:
 	docker run -it --rm -v ${CURDIR}:/opt/phpdoc -w /opt/phpdoc node npm install
 
-data/examples/MariosPizzeria/build/default/index.html: data/examples/MariosPizzeria/**/*
-	CURRENT_UID=$(shell id -u):$(shell id -g) docker-compose run --rm -v ${CURDIR}:/opt/phpdoc -w /opt/phpdoc/data/examples/MariosPizzeria phpdoc --template=default --target=build/default
+build/default/index.html: data/examples/MariosPizzeria/**/*
+	CURRENT_UID=$(shell id -u):$(shell id -g) docker-compose run --rm -v ${CURDIR}:/opt/phpdoc phpdoc --config=data/examples/MariosPizzeria/phpdoc.xml --template=default --target=build/default
```
mvriel: Nice improvement; I like it :)

PR #2197, opened 2020-01-04
Windows file scheme fix

jaapio:
php on windows cannot handle file:/// protocol. So we remove it.


src/phpDocumentor/Parser/FlySystemFactory.php
```diff
@@ -74,6 +79,12 @@ private function createAdapter(Dsn $dsn) : AdapterInterface
             );
         }
 
+        if (PHP_OS_FAMILY === 'Windows') {
```
mvriel: Maybe add a comment / extract to a method + docblock why you do this?

mvriel:
Build fails on Windows?

jaapio:
Need to check, but I think it is some broken unittest giving us false hope 😋

PR #2202, opened 2020-01-08
Ugly hacking to fix windows

jaapio:


src/phpDocumentor/Pipeline/Stage/Configure.php
```diff
@@ -51,7 +52,7 @@ public function __construct(
     public function __invoke(array $options) : array
     {
         $this->configFactory->addMiddleware(
-            new CommandlineOptionsMiddleware($options, $this->configFactory, getcwd())
+            new CommandlineOptionsMiddleware($options, $this->configFactory, 'file:///' . getcwd())
```
mvriel: getcwd already returns an absolute path; shouldn't there thus be only 2 slashes?

jaapio: That depends on the os😭

jaapio:
@mvriel I would be very happy if you are able to review this PR. I would like to improve the code a bit based on your feedback. I'm not very happy with the introduction of the `UriFactory` but I didn't want to duplicate the windows only checks.

src/phpDocumentor/Configuration/ConfigurationFactory.php
```diff
@@ -13,15 +13,16 @@
...
 use phpDocumentor\Configuration\Exception\InvalidConfigPathException;
+use phpDocumentor\UriFactory;
 use function file_exists;
 use function sprintf;
 
 /**
  * The ConfigurationFactory converts the configuration xml from a Uri into an array.
  */
-final class ConfigurationFactory
+/*final*/ class ConfigurationFactory
```
mvriel: Commented final? Shouldn't we remove the final or fix the tests?

src/phpDocumentor/Dsn.php
```diff
@@ -64,15 +62,15 @@ final class Dsn
...
-        $this->uri = $uri;
+        $this->uri        = $uri;
     }
 
     public static function createFromString(string $dsn) : self
     {
-        $parameters = explode(';', $dsn);
-        $uri = self::parseUri(array_shift($parameters));
+        $parameters       = explode(';', $dsn);
+        $uri              = UriFactory::createUri(array_shift($parameters));
```
mvriel: Maybe we should standardize whether to align or not to prevent back and forth changes :)

src/phpDocumentor/UriFactory.php
```diff
@@ -0,0 +1,68 @@
...
+use Throwable;
+use const DIRECTORY_SEPARATOR;
+use function preg_match;
+use function sprintf;
+use function str_replace;
+use function strlen;
+use function strpos;
+use function substr;
+
+final class UriFactory
```
mvriel: What you could do, is instead of making this a factory, is to make this a Proxy / Decorator around `League/Uri` and have the createUri method as a static factory method. That way:

1. you don't have to inject a Factory service into the DSN
2. we have an anti-corruption layer between phpDocumentor and League/Uri; possibly making it easier to replace league/uri in the future if necessary.

If you agree, I would recommend doing that change after merging this PR

src/phpDocumentor/UriFactory.php
```diff
@@ -0,0 +1,68 @@
...
+                    $uriString = substr($uriString, strlen('file:///'));
+                }
+
+                return LeagueUri::createFromWindowsPath($uriString);
+            }
+
+            return LeagueUri::createFromString($uriString);
+        } catch (Throwable $exception) {
+            throw new InvalidArgumentException(
+                sprintf(
```
mvriel: if you extract this into a `$message` local variable the code might look cleaner as you may not need to wrap onto multiple lines?

tests/unit/phpDocumentor/UriFactoryTest.php
```diff
@@ -0,0 +1,71 @@
...
+
+declare(strict_types=1);
+
+namespace phpDocumentor;
+
+use InvalidArgumentException;
+use PHPUnit\Framework\TestCase;
+
+/**
+ * @coversDefaultClass \phpDocumentor\UriFactory
```
mvriel: missing covers statement for the `<private>`

tests/unit/phpDocumentor/Pipeline/Stage/ConfigureTest.php
```diff
@@ -0,0 +1,97 @@
+<?php
+
+declare(strict_types=1);
+
```
mvriel: Missing file docblock

tests/unit/phpDocumentor/UriFactoryTest.php
```diff
@@ -0,0 +1,71 @@
+<?php
+
+declare(strict_types=1);
+
```
mvriel: Missing file docblock

src/phpDocumentor/UriFactory.php
```diff
@@ -0,0 +1,68 @@
...
+use function preg_match;
+use function sprintf;
+use function str_replace;
+use function strlen;
+use function strpos;
+use function substr;
+
+final class UriFactory
+{
+    public const WINDOWS_URI_FORMAT = '~^(file:\/\/\/)?(?<root>[a-zA-Z][:|\|])~';
```
mvriel: It feels weird to have this public and a value object consuming this; it makes for coupling between DSN and this service. Perhaps making this a proxy (as suggested earlier) with a helper method will resolve that?

src/phpDocumentor/UriFactory.php
```diff
@@ -0,0 +1,68 @@
+<?php
+
+declare(strict_types=1);
+
```
mvriel: Missing File DocBlock

src/phpDocumentor/Configuration/ConfigurationFactory.php
jaapio: Test needs to be fixed, but that needs to be done in a second pr. Because it feels odd to introduce an interface to be able to mock this class, without having it properly shaped. We still do need to merge the symfonyConfigFactory into this class.

Issue #2216, opened 2020-01-25
Move from phpstan level 5 to level 6


mvriel:
In the past few months, phpDocumentor has been pushing to upgrade the quality of the codebase by introducing tools such as phpstan. We have made enough fixes to get to phpstan level 5 and would like help to get to phpstan 6, and eventually to max level.

This specific ticket is for getting to phpstan level 6.

orklah:
This has been solved, it could be closed :)

PR #2238, opened 2020-01-27
fix some errors on PHPStan lvl 6

orklah:
This PR make some progress for #2216

It fixes some errors on level 6 (mainly generics description and some typos on property names in phpdoc)

There are 2 errors left on lvl 5 from recent commits but the first one implied to add checks in code and I didn't know how to fix the second one.

This PR reveals 2 new errors on level 5 that I didn't analyze for now.

I didn't run on docker and for some reason, I couldn't launch phpcbf. I'll fix errors from CI reported here.

### DIFF
--- a/src/phpDocumentor/Descriptor/ClassDescriptor.php
+++ b/src/phpDocumentor/Descriptor/ClassDescriptor.php
@@ -24,7 +24,11 @@ use function sprintf;
  */
 class ClassDescriptor extends DescriptorAbstract implements Interfaces\ClassInterface
 {
-    /** @var ClassDescriptor|null $extends Reference to an instance of the superclass for this class, if any. */
+    /**
+     * Reference to an instance of the superclass for this class, if any.
+     *
+     * @var DescriptorAbstract|Fqsen|string|null $parent
+     */
     protected $parent;
 
     /** @var Collection $implements References to interfaces that are implemented by this class. */
@@ -63,7 +67,7 @@ class ClassDescriptor extends DescriptorAbstract implements Interfaces\ClassInte
     }
 
     /**
-     * {@inheritDoc}
+     * @param DescriptorAbstract|Fqsen|string|null $parents
      */
     public function setParent($parents) : void
     {
@@ -71,7 +75,7 @@ class ClassDescriptor extends DescriptorAbstract implements Interfaces\ClassInte
     }
 
     /**
-     * {@inheritDoc}
+     * @return DescriptorAbstract|Fqsen|string|null
      */
     public function getParent()
     {
@@ -97,7 +101,7 @@ class ClassDescriptor extends DescriptorAbstract implements Interfaces\ClassInte
     /**
      * {@inheritDoc}
      */
-    public function setFinal($final) : void
+    public function setFinal(bool $final) : void
     {
         $this->final = $final;
     }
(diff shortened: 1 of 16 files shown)


orklah:
> It's looking good! I do see that the build currently fails on one last phpstan issue :)

Yeah, I'm investigating that. I assumed PHPStan was not part of the CI because it was failing on 2 errors locally. 1 was revealed by PHPStan in the version released this morning but I don't explain the other yet.

And locally I have 2 new errors (instead of one on CI) that are similar to the second one.

I'll try to fix that :)

mvriel:
Super! Good work!

jaapio:
Thanks for your efforts!

It might help to check your phpstan version. We are not running the latest version, but 0.12.3

orklah:
I don't get it, there's something wrong with the ChildInterface.

Its getParent method can return `DescriptorAbstract|Fqsen|string|null` but setParent can only take `?DescriptorAbstract`. It's already weird.

On top of that, setParent has a param hint so it seems right, but one of it's child (ClassDescriptor) does not repeat the type hint so it can technically take anything. And it seems it does take anything because in ClassAssembler, we push ?Fqsen in it, which is coherent with the ChildInterface::getParent return phpdoc.

It seems to me that ChildInterface is used twice but for two classes that share method names but not types. If that's the case, I see two solutions:
- Drop types in interface, drop inheritDoc from implemented classes and describe types in there. Con: you won't be able to put a param hint because the interface won't allow it
- Drop the interface

What do you think?

> It might help to check your phpstan version. We are not running the latest version, but 0.12.3

That must be it. I was surprised to see phpstan missing from composer.json but I didn't see it was installed elsewhere.

jaapio:
I do agree that the descriptors are a mess. They are on a list to be refactored, but that doesn't have any prio at the moment. I think it is ok to do your first option for now. 

We do not include any of the dev tools to composer since we had a lot of issues in the past by including those. Mainly because a number of our libraries are very commonly used in the php-eco-system. 3 of them are a dependency of phpunit for example. So we were not able to update our own code until phpunit did support our pre-released libs. Therefore we are now using phive to install the phars. You can have a look at our contributors' guide for more information.

orklah:
Ok, Phpstan is fixed now but I can't get phpcbf to run locally (I'm on windows and I'd prefer avoid Docker and Phive for now) and it catch an error that I don't know how to fix.

If a property doc comment is > 120 chars, it breaks CS but if I try to put two lines, there's another rule  that break (multi-line comment). What is the correct way to fix that?

src/phpDocumentor/Descriptor/ClassDescriptor.php
```diff
@@ -24,7 +24,9 @@
  */
 class ClassDescriptor extends DescriptorAbstract implements Interfaces\ClassInterface
 {
-    /** @var ClassDescriptor|null $extends Reference to an instance of the superclass for this class, if any. */
+    /** @var DescriptorAbstract|Fqsen|string|null $parent Reference to an instance of the superclass for this class,
```
mvriel: You can fix the CS issue like this:

```
    /**
     * Reference to an instance of the superclass for this class, if any.
     *
     * @var DescriptorAbstract|Fqsen|string|null $parent
     */
```

mvriel:
@orklah Sometimes it gets a bit tricky indeed; I have considered removing one of those rules but found that it frequently does give better results with it enabled.

In this case, by moving the description from the `@var` tag to the summary of the DocBlock; this is considered to be acceptable and will result in similar documentation

src/phpDocumentor/Descriptor/ClassDescriptor.php
orklah: Facepalm... Of course!

mvriel:
Thank you so much for checking and resolving the style warnings in the PR. These improvements are perfect and I have just merged them.

PR #2239, opened 2020-01-28
progress on phpstan level 6

orklah:
Still progressing to #2216

This adds a lot of generics notation.

However, I used the `list<>` notation, introduced in PhpStan on [0.12.4](https://github.com/phpstan/phpstan/releases/tag/0.12.4) I'm not sure how to proceed to bump phpstan version on phive?

(pretty sure I messed up CS sometimes. I'll fix that)

### DIFF
--- a/phpcs.xml.dist
+++ b/phpcs.xml.dist
@@ -24,6 +24,7 @@
         <exclude name="SlevomatCodingStandard.TypeHints.TypeHintDeclaration.MissingTraversableReturnTypeHintSpecification" />
         <exclude name="SlevomatCodingStandard.TypeHints.ParameterTypeHint.MissingNativeTypeHint" />
         <exclude name="SlevomatCodingStandard.TypeHints.TypeHintDeclaration.MissingParameterTypeHint" />
+        <exclude name="SlevomatCodingStandard.Namespaces.UnusedUses" />
     </rule>
 
     <rule ref="SlevomatCodingStandard.Functions.StaticClosure.ClosureNotStatic">
--- a/src/phpDocumentor/Descriptor/Builder/AssemblerInterface.php
+++ b/src/phpDocumentor/Descriptor/Builder/AssemblerInterface.php
@@ -16,6 +16,9 @@ namespace phpDocumentor\Descriptor\Builder;
 use phpDocumentor\Descriptor\Collection;
 use phpDocumentor\Descriptor\Descriptor;
 use phpDocumentor\Descriptor\ProjectDescriptorBuilder;
+use phpDocumentor\Reflection\DocBlock\Tag;
+use phpDocumentor\Reflection\Element;
+use phpDocumentor\Reflection\Php\Argument;
 
 /**
  * Interface for Assembler classes that transform data to specific Descriptor types.
@@ -25,7 +28,7 @@ interface AssemblerInterface
     /**
      * Creates a Descriptor from the provided data.
      *
-     * @param mixed $data
+     * @param Element|Tag|Argument $data
      *
      * @return Descriptor|Collection
      */
--- a/src/phpDocumentor/Descriptor/ClassDescriptor.php
+++ b/src/phpDocumentor/Descriptor/ClassDescriptor.php
@@ -31,7 +31,11 @@ class ClassDescriptor extends DescriptorAbstract implements Interfaces\ClassInte
      */
     protected $parent;
 
-    /** @var Collection $implements References to interfaces that are implemented by this class. */
+    /**
+     * References to interfaces that are implemented by this class.
+     *
+     * @var Collection<InterfaceDescriptor|Fqsen> $implements
+     */
     protected $implements;
 
     /** @var bool $abstract Whether this is an abstract class. */
@@ -40,16 +44,16 @@ class ClassDescriptor extends DescriptorAbstract implements Interfaces\ClassInte
     /** @var bool $final Whether this class is marked as final and can't be subclassed. */
     protected $final = false;
 
-    /** @var Collection $constants References to constants defined in this class. */
+    /** @var Collection<?ConstantDescriptor> $constants References to constants defined in this class. */
     protected $constants;
 
-    /** @var Collection $properties References to properties defined in this class. */
+    /** @var Collection<?PropertyDescriptor> $properties References to properties defined in this class. */
     protected $properties;
 
-    /** @var Collection $methods References to methods defined in this class. */
+    /** @var Collection<?MethodDescriptor> $methods References to methods defined in this class. */
     protected $methods;
 
-    /** @var Collection $usedTraits References to traits consumed by this class */
+    /** @var Collection<TraitDescriptor> $usedTraits References to traits consumed by this class */
     protected $usedTraits;
 
     /**
@@ -205,6 +209,7 @@ class ClassDescriptor extends DescriptorAbstract implements Interfaces\ClassInte
         /** @var Collection $methodTags */
         $methodTags = clone $this->getTags()->get('method', new Collection());
 
+        /** @var Collection<MethodDescriptor> $methods */
         $methods = new Collection();
 
         /** @var Tag\MethodDescriptor $methodTag */
@@ -279,6 +284,7 @@ class ClassDescriptor extends DescriptorAbstract implements Interfaces\ClassInte
         $propertyTags = $propertyTags->merge($this->getTags()->get('property-read', new Collection()));
         $propertyTags = $propertyTags->merge($this->getTags()->get('property-write', new Collection()));
 
+        /** @var Collection<PropertyDescriptor> $properties */
         $properties = new Collection();
 
         try {
--- a/src/phpDocumentor/Descriptor/Collection.php
+++ b/src/phpDocumentor/Descriptor/Collection.php
@@ -77,9 +77,7 @@ class Collection implements Countable, IteratorAggregate, ArrayAccess
      * @param string|int $index
      * @param T          $valueIfEmpty If the index does not exist it will be created with this value and returned.
      *
-     * @return mixed The contents of the element with the given index and the provided default if the key doesn't exist.
-     *
-     * @phpstan-return ?T
+     * @return ?T The contents of the element with the given index and the provided default if the key doesn't exist.
      */
     public function get($index, $valueIfEmpty = null)
     {
@@ -129,9 +127,7 @@ class Collection implements Countable, IteratorAggregate, ArrayAccess
     /**
      * Retrieves an item as if it were a property of the collection.
      *
-     * @return mixed
-     *
-     * @phpstan-return ?T
+     * @return ?T
      */
     public function __get(string $name)
     {
@@ -153,9 +149,7 @@ class Collection implements Countable, IteratorAggregate, ArrayAccess
      *
      * @param string|int $offset The offset to retrieve.
      *
-     * @return mixed
-     *
-     * @phpstan-return ?T
+     * @return ?T
      */
     public function offsetGet($offset)
     {
--- a/src/phpDocumentor/Descriptor/Collection.php
+++ b/src/phpDocumentor/Descriptor/Collection.php
@@ -77,7 +77,9 @@ class Collection implements Countable, IteratorAggregate, ArrayAccess
      * @param string|int $index
      * @param T          $valueIfEmpty If the index does not exist it will be created with this value and returned.
      *
-     * @return ?T The contents of the element with the given index and the provided default if the key doesn't exist.
+     * @return mixed The contents of the element with the given index and the provided default if the key doesn't exist.
+     *
+     * @phpstan-return ?T
      */
     public function get($index, $valueIfEmpty = null)
     {
@@ -127,7 +129,9 @@ class Collection implements Countable, IteratorAggregate, ArrayAccess
     /**
      * Retrieves an item as if it were a property of the collection.
      *
-     * @return ?T
+     * @return mixed
+     *
+     * @phpstan-return ?T
      */
     public function __get(string $name)
     {
@@ -149,7 +153,9 @@ class Collection implements Countable, IteratorAggregate, ArrayAccess
      *
      * @param string|int $offset The offset to retrieve.
      *
-     * @return ?T
+     * @return mixed
+     *
+     * @phpstan-return ?T
      */
     public function offsetGet($offset)
     {
--- a/src/phpDocumentor/Descriptor/Collection.php
+++ b/src/phpDocumentor/Descriptor/Collection.php
@@ -27,18 +27,19 @@ use function count;
  * The goal for this class is to allow Descriptors to be easily retrieved and set so that interaction in
  * templates becomes easier.
  *
- * @template-implements ArrayAccess<string|int, mixed>
- * @template-implements IteratorAggregate<string|int, mixed>
+ * @template T
+ * @template-implements ArrayAccess<string|int, T>
+ * @template-implements IteratorAggregate<string|int, T>
  */
 class Collection implements Countable, IteratorAggregate, ArrayAccess
 {
-    /** @var mixed[] $items */
+    /** @var array<T> $items */
     protected $items = [];
 
     /**
      * Constructs a new collection object with optionally a series of items, generally Descriptors.
      *
-     * @param DescriptorAbstract[]|mixed[] $items
+     * @param array<T> $items
      */
     public function __construct(array $items = [])
     {
@@ -48,7 +49,7 @@ class Collection implements Countable, IteratorAggregate, ArrayAccess
     /**
      * Adds a new item to this collection, generally a Descriptor.
      *
-     * @param DescriptorAbstract|mixed $item
+     * @param T $item
      */
     public function add($item) : void
     {
@@ -58,8 +59,8 @@ class Collection implements Countable, IteratorAggregate, ArrayAccess
     /**
      * Sets a new object onto the collection or clear it using null.
      *
-     * @param string|int                    $index An index value to recognize this item with.
-     * @param DescriptorAbstract|mixed|null $item  The item to store, generally a Descriptor but may be something else.
+     * @param string|int $index An index value to recognize this item with.
+     * @param T          $item  The item to store, generally a Descriptor but may be something else.
      */
     public function set($index, $item) : void
     {
@@ -74,9 +75,11 @@ class Collection implements Countable, IteratorAggregate, ArrayAccess
      * tree building operations.
      *
      * @param string|int $index
-     * @param mixed      $valueIfEmpty If the index does not exist it will be created with this value and returned.
+     * @param T          $valueIfEmpty If the index does not exist it will be created with this value and returned.
      *
      * @return mixed The contents of the element with the given index and the provided default if the key doesn't exist.
+     *
+     * @phpstan-return ?T
      */
     public function get($index, $valueIfEmpty = null)
     {
@@ -90,7 +93,7 @@ class Collection implements Countable, IteratorAggregate, ArrayAccess
     /**
      * Retrieves all items from this collection as PHP Array.
      *
-     * @return mixed[]
+     * @return array<T>
      */
     public function getAll() : array
     {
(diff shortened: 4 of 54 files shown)


mvriel:
For our build pipeline, we use https://github.com/phpDocumentor/quality-assurance/tree/master/phpstan. So I would need to trigger a new build. This can take a little while as I am in the middle of a story and still need to rerelease the phar of 2.9.1. before I get to this :sweat_smile:

jaapio:
Thanks again!

We will need to be careful since the github action is used in multiple repo's and phpstan is constantly adding more checks. So a new version might mean that we need to update the other repo's as well. This could be done, or we need to check how to create tag for that action.

orklah:
This seems more complex than I thought.

Here are the solutions I see:
- wait before merging this PR
- fallback to use `array<int, type>` instead of `list<type>` (this is a shame, Psalm use list to add more useful checks)
- add @psalm-param to keep `list<type>` while still using array<int, type> in @param

what do you think?

jaapio:
I think it is ok to start the upgrade, if you have time to help us with that ;-) 

If you don't want to commit your support for a longer time, which I do understand, I would recommend you to choose the latter. Since you are helping us forward with each contribution! We could always write a sniff or migration to move the `list<type>` to `@param` 

The introduction of psaml is on my wish list by the way ;-)

orklah:
I'd prefer focusing on static analysis for now, especially since I don't have a docker environment running (and had trouble in the past trying to set it up).

I tried to run psalm too but I had conflicts with dependancies as you said. I will try again later.

For now I'll do as you suggest concerning lists.

orklah:
I still have two issues.
- Some annotations don't play nice when it comes to running phpdocumentor against itself. If I interpreted my debugs correctly, It doesn't like "class-string" because it's not a Fqsen and it also doesn't like the array{key: value, key: value} notation either. I could drop them and try to use more standard types or notations...
- Codestyle issues when trying to import classes that are used in generics notations: If I try to add "use Namespace\SomeClass" and use it in an `@param array<SomeClass>`, the sniff won't parse the notation and will consider the import as useless. I'd recommand dropping this particular sniff...

orklah:
Turns out, the sniff can actually parse generics annotations and suppress this false positive:
https://github.com/slevomat/coding-standard/issues/251

However, this was released on version 5.0. Which is allowed by doctrine/doctrine-standard from version 6.0 and we actually use 5.0.

This would be another bump to plan if we want to avoid excluding this sniff...

jaapio:
Thanks for all your hard work. I delay the merge of this pr since we are preparing a new beta release. After that is done I will have a look what we can do to release some new dev tools to get this pr merged.

orklah:
No worries ;) Keep me updated !

jaapio:
@orklah could you please rebase your PR? 

The beta release is done, so I'm now picking up the work on our QA tools, think they require some extra attention to make this work. 

Regarding the issues you are facing of the new pseudotypes introduced by psalm and phpstan. We have an open issue for that in our docblock-reflection component. It will require extra work to get that support. I need to discuss that with @mvriel on how to improve this.

orklah:
I tried to rebase earlier but I get conflict and I'm not very comfortable with git... Is there a tool you use for that or just the command line?

jaapio:
I'm just using the command line for git. To resolve merge conflicts phpstorm is quite OK.

And I do hear good things about git kraken.

No problem if you can't rebase this. I can have a look.
First get this pr reviewed. And all  related stuff done:grinning:

orklah:
> I'm just using the command line for git. To resolve merge conflicts phpstorm is quite OK.

Nice, I didn't realise conflicts could be resolved in PhpStorm during a rebase. It worked well :)

phpcs.xml.dist
```diff
@@ -24,6 +24,7 @@
         <exclude name="SlevomatCodingStandard.TypeHints.TypeHintDeclaration.MissingTraversableReturnTypeHintSpecification" />
         <exclude name="SlevomatCodingStandard.TypeHints.ParameterTypeHint.MissingNativeTypeHint" />
         <exclude name="SlevomatCodingStandard.TypeHints.TypeHintDeclaration.MissingParameterTypeHint" />
+        <exclude name="SlevomatCodingStandard.Namespaces.UnusedUses" />
```
mvriel: Did you exclude this intentionally? I kinda like this one but I assume you had a good reason. If it was because of a single file, you can make your excludes limit itself to one file.

src/phpDocumentor/Descriptor/Builder/AssemblerInterface.php
```diff
@@ -25,7 +28,7 @@ interface AssemblerInterface
     /**
      * Creates a Descriptor from the provided data.
      *
-     * @param mixed $data
+     * @param Element|Tag|Argument $data
```
mvriel: I am surprised this works; there are so many things used :) 

This is a bad pattern, I know.. The descriptors and assemblers could use some refactoring to make them more type strict. Just so you know, after 3.0 launches we want to see if we can make these objects simple value objects in the domain instead of this complicated thing (especially the inheritance is complicated)

mvriel: (just to prevent confusion; no action is wanted. I just wanted to be transparent and express my surprise :) )

phpcs.xml.dist
orklah: I'm talking about it here:
https://github.com/phpDocumentor/phpDocumentor/pull/2239#issuecomment-579870267
https://github.com/phpDocumentor/phpDocumentor/pull/2239#issuecomment-579945218

tdlr: we use an old version of slevomat which can't read generics notation yet. This could be added back if we switch to Doctrine-standard ^5

src/phpDocumentor/Descriptor/ClassDescriptor.php
```diff
@@ -40,16 +44,16 @@ class ClassDescriptor extends DescriptorAbstract implements Interfaces\ClassInte
     /** @var bool $final Whether this class is marked as final and can't be subclassed. */
     protected $final = false;
 
-    /** @var Collection $constants References to constants defined in this class. */
+    /** @var Collection<?ConstantDescriptor> $constants References to constants defined in this class. */
```
mvriel: Aj, those nullables hurt .. we should really change the assemblers to stop them from inserting nulls into these collections :disappointed:

src/phpDocumentor/Descriptor/Collection.php
```diff
@@ -27,18 +27,19 @@
  * The goal for this class is to allow Descriptors to be easily retrieved and set so that interaction in
  * templates becomes easier.
  *
- * @template-implements ArrayAccess<string|int, mixed>
- * @template-implements IteratorAggregate<string|int, mixed>
+ * @template T
```
mvriel: A generic! How wonderful and distressing at the same time! :D Not because generics are bad; but because phpDocumentor does not support them yet :sweat_smile: Sounds like a good 'eat your own dogfood' case to build support for them :) (We do want this in 3.1)

src/phpDocumentor/Descriptor/Collection.php
```diff
@@ -149,6 +154,8 @@ public function offsetExists($offset) : bool
      * @param string|int $offset The offset to retrieve.
      *
      * @return mixed
+     *
+     * @phpstan-return ?T
```
mvriel: I need to read up why this can't go in a regular `@return` .. I would expect it in the `@return` :thinking:

src/phpDocumentor/Descriptor/Builder/AssemblerInterface.php
orklah: At work, we have a massive legacy project, the key is progress step by step. Documenting what the code do now is a good progress. It'll be easier to refactor later if we know exactly what's going on :)

PhpStan and Psalm are big help in that by checking the coherence of the whole and making sure we're not adding mistakes :)

src/phpDocumentor/Descriptor/DocumentationSetDescriptor.php
```diff
@@ -13,12 +13,17 @@
...
+use phpDocumentor\Dsn;
+
 abstract class DocumentationSetDescriptor
 {
     /** @var string */
     protected $name = '';
 
-    /** @var array */
+    /**
+     * @phpstan-var array{dsn?: Dsn, paths?: array<int, string>}
```
mvriel: @jaapio Do you see this? PHPStan actually implemented a variant of the associative array syntax that we proposed in PSR-5 :) Perhaps worth checking out the adoption rate of this

src/phpDocumentor/Descriptor/FileDescriptor.php
```diff
@@ -42,16 +38,16 @@ class FileDescriptor extends DescriptorAbstract implements Interfaces\FileInterf
     /** @var Collection $constants */
     protected $constants;
 
-    /** @var Collection $functions */
+    /** @var Collection<FunctionDescriptor> $functions */
```
mvriel: :thinking: Why do you type-hint against the concrete classes and not the interfaces? I am not saying that the interfaces are rather useful at this point (there were plans in the past) but didn't phpstan like that?

src/phpDocumentor/Descriptor/ClassDescriptor.php
orklah: Well, I didn't add them at first and it didn't seem to be an issue, I only saw later that there was special cases for null that PHPStan was raising. If there is really null in there, I don't know from where. Maybe we'll discover more with higher level analysis...

src/phpDocumentor/Descriptor/GuideSetDescriptor.php
```diff
@@ -13,8 +13,15 @@
 
 namespace phpDocumentor\Descriptor;
 
+use phpDocumentor\Dsn;
+
 final class GuideSetDescriptor extends DocumentationSetDescriptor
 {
+    /**
+     * @param array<Dsn|list<string>> $source
```
mvriel: Oh? I need to check where I made a mistake. I don't think I meant this to be either

src/phpDocumentor/Descriptor/InterfaceDescriptor.php
```diff
@@ -20,13 +20,13 @@
  */
 class InterfaceDescriptor extends DescriptorAbstract implements Interfaces\InterfaceInterface
 {
-    /** @var Collection $parents */
+    /** @var Collection<InterfaceDescriptor> $parents */
```
mvriel: Same question as before about using the concrete classes and not the interfaces

src/phpDocumentor/Descriptor/Collection.php
orklah: Yeah, really useful to use, but it must have been a pain to implement!

src/phpDocumentor/Descriptor/Interfaces/ConstantInterface.php
```diff
@@ -28,7 +28,9 @@ public function setTypes(Type $types) : void;
     /**
      * Returns the types that may be present in this constant.
      *
-     * @return Type[]
+     * @return list<?Type>
+     *
+     * @phpstan-return array<int, ?Type>
```
mvriel: Type is nullable? Again one to investigate later; I would not have expected this

src/phpDocumentor/Descriptor/DocumentationSetDescriptor.php
jaapio: Yes I know the existence of this notations. Although I don't like the notation it makes a lot of sense for legacy projects.
I do agree that we need to adopt this, all static analyzers (3) in php do support this.
We might want to check this with the guys from jetbrains. They are in the psr-5 group

src/phpDocumentor/Descriptor/PropertyDescriptor.php
```diff
@@ -100,7 +100,9 @@ public function setType(Type $type) : void
     }
 
     /**
-     * {@inheritDoc}
+     * @return list<string>
```
mvriel: Where does `list` come from? Do you know? I do not recognize this pseudo-type

src/phpDocumentor/Descriptor/Collection.php
orklah: I'm actually not sure why I did it that way. I'm looking into it

src/phpDocumentor/Descriptor/DocumentationSetDescriptor.php
orklah: Note that PHPStan and Psalm try to have similar syntaxes (both reads each other @xxx- notations while scanning).
Here are the notations supported by psalm: https://psalm.dev/docs/annotating_code/type_syntax/array_types/

phpcs.xml.dist
mvriel: :flushed: You are right, I did not read your comments carefully enough

src/phpDocumentor/Descriptor/Collection.php
mvriel: They are a completely new type of syntax that breaks with the way tags were parsed. In the past, you could split a tag based on whitespace to get its parts; but with things like generics you can no longer do that and need a more intelligent parser

src/phpDocumentor/Descriptor/FileDescriptor.php
orklah: Either I found the concrete class more easily or the interface didn't cover everything. For now PHPStan don't raise error for that. Maybe on higher levels...

src/phpDocumentor/Descriptor/Collection.php
jaapio: I think this is not nullable, but since a lot of methods in the descriptors are returning null. And we had a bug where we needed to filter those, it could be that phpstan is applying some reversed logic here.

mvriel:
@orklah what a great PR; I must honestly admit that I thought the build would fail on some of these notations but phpDocumentor is more resilient than I had expected :flushed: 

You do uncover interesting points with these changes, and improvements that are still pending. For this, I am very grateful.

src/phpDocumentor/Descriptor/Interfaces/ConstantInterface.php
orklah: ConstantDescriptor have a getType method which checks against null and could theorically return null. I don't know if it happens, but if I don't document it here, I'll have to handle all the null cases in every child method to keep coherence.

src/phpDocumentor/Descriptor/FileDescriptor.php
mvriel: Let's fix that in a followup; that we either remove the interfaces (if they are useless) or change it so we use those interfaces here

src/phpDocumentor/Descriptor/InterfaceDescriptor.php
mvriel: Answered earlier

src/phpDocumentor/Descriptor/Interfaces/ConstantInterface.php
mvriel: You did good; this just makes it painfully obvious that we need to fix something

mvriel:
I am now making new issues for the things you uncovered that we should look into / improve; since the build is green I would love to merge this as soon as you are satisfied too @orklah

orklah:
> I must honestly admit that I thought the build would fail on some of these notations but phpDocumentor is more resilient than I had expected 😳

The only syntaxes I had to revert to make it all green are:
- class-string: It tries to handle this as a Fqsen but the "-" make it obvious this is not a class. I suppose every notation in https://psalm.dev/docs/annotating_code/type_syntax/scalar_types/ could be handled the same way
- array{} notation

src/phpDocumentor/Descriptor/Collection.php
orklah: I do remember now why I use specific phpstan tags :D

```
----------------------------------------------------------------------
FOUND 3 ERRORS AFFECTING 3 LINES
----------------------------------------------------------------------
  82 | ERROR | [x] Method \phpDocumentor\Descriptor\Collection::get()
     |       |     does not have return type hint for its return
     |       |     value but it should be possible to add it based on
     |       |     @return annotation "?T".
 132 | ERROR | [x] Method
     |       |     \phpDocumentor\Descriptor\Collection::__get() does
     |       |     not have return type hint for its return value but
     |       |     it should be possible to add it based on @return
     |       |     annotation "?T".
 154 | ERROR | [x] Method
     |       |     \phpDocumentor\Descriptor\Collection::offsetGet()
     |       |     does not have return type hint for its return
     |       |     value but it should be possible to add it based on
     |       |     @return annotation "?T".
----------------------------------------------------------------------
PHPCBF CAN FIX THE 3 MARKED SNIFF VIOLATIONS AUTOMATICALLY
----------------------------------------------------------------------
```

CS checks will kill me!

src/phpDocumentor/Descriptor/Builder/AssemblerInterface.php
mvriel: Created issue https://github.com/phpDocumentor/phpDocumentor/issues/2245 to specifically target this

phpcs.xml.dist
mvriel: Commented on https://github.com/phpDocumentor/phpDocumentor/pull/2243 if we can include this

src/phpDocumentor/Descriptor/Collection.php
orklah: I believe PHPStan parser is a distinct repo on github and it can be reused. IIRC, they used it in slevomat to read annotations and fix the issue where it believed an import was not use (which I suppressed)

src/phpDocumentor/Descriptor/ClassDescriptor.php
mvriel: Created issue https://github.com/phpDocumentor/phpDocumentor/issues/2246 to further investigate this

src/phpDocumentor/Descriptor/Collection.php
mvriel: We have our own component that supports quite a few more features than other DocBlock parsers out there; but not (yet) generics :D

orklah:
If CI pass on this one, I think this can be merged. I'll probably open a new PR later with more changes to follow. (there's still a lot of work to pass Phpstan to level 6 :p)

src/phpDocumentor/Descriptor/Collection.php
mvriel: It would actually have been nice if other projects would contribute a missing feature instead of building another DocBlock parser :sweat_smile:

src/phpDocumentor/Descriptor/Collection.php
mvriel: Ah crap; we may need to report this to phpcs and/or the people who made these specific sniffs

src/phpDocumentor/Descriptor/Collection.php
orklah: True. I don't know why it was done that way :(

src/phpDocumentor/Descriptor/FileDescriptor.php
mvriel: created issue https://github.com/phpDocumentor/phpDocumentor/issues/2247 to check this

src/phpDocumentor/Descriptor/Collection.php
orklah: It's possible this is fixed already. It's kind of the same issue than parsing generics, It could have been fixed in a newer version of sniffs.

I'll probably try back later if the QA tools are updated.

src/phpDocumentor/Descriptor/Interfaces/ConstantInterface.php
mvriel: Made issue https://github.com/phpDocumentor/phpDocumentor/issues/2248 to resolve this

mvriel:
And it is merged! Thank you!

orklah:
I think this was my biggest PR ever :)

Thanks for reviews and advices!

PR #2243, opened 2020-01-31
Switch to phpDocumentor code style

jaapio:


mvriel:
Perhaps we can also fix this issue at the same time with this? :)

https://github.com/phpDocumentor/phpDocumentor/pull/2239#discussion_r373788920

PR #2244, opened 2020-02-01
experiment with Router

orklah:
Router extends ArrayObject but don't push anything in parent:__construct, don't override methods and don't use parent::

This is a test to see if the extends is useful.


orklah:
The test seems successful. I think this can be merged.

orklah:
Weird, Integrations tests failed. I don't think it's related

orklah:
Better now!

orklah:
> Good catch!

It was not so difficult to catch, PHPStan needs to know every generic template at level 6. Every use of Router was a warning :)

PR #2249, opened 2020-02-01
Add collections types for phpstan

orklah:
This is a draft PR to see CI results

This is probably broken :)

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -14,6 +14,10 @@ parameters:
     - '#Call to an undefined method League\\Flysystem\\Filesystem::find()#'
     # Bad design of descriptors.
     - '#Access to an undefined property phpDocumentor\\Descriptor\\Collection::\$[a-z]+#'
+    # Issue on the content of this collection https://github.com/phpDocumentor/phpDocumentor/pull/2249#issuecomment-581111947
+    - '#Parameter \#1 \$item of method phpDocumentor\\Descriptor\\Collection<phpDocumentor\\Descriptor\\TagDescriptor>::add\(\) expects phpDocumentor\\Descriptor\\TagDescriptor, phpDocumentor\\Descriptor\\Collection<phpDocumentor\\Descriptor\\DescriptorAbstract>\|phpDocumentor\\Descriptor\\Descriptor given\.#'
+    # Issue with nullability in generics. Maybe Phpstan bug?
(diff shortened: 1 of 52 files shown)


orklah:
@mvriel @jaapio I need a little help
See https://github.com/phpDocumentor/phpDocumentor/pull/2249/checks?check_run_id=421122194#step:6:23
I get an error on
https://github.com/phpDocumentor/phpDocumentor/pull/2249/files#diff-d14768b872eee4d6b9e87fd8ba402f36R51
because an element is not a TagDescriptor. Is this normal? If so, it make getTags() returning `Collection<Descriptor>` at best and it cause other errors later because Descriptor doesn't cover all of TagDescriptor.

Thanks

mvriel:
No, this is not normal. Everyone should be a Tag. If you look in the FunctionAssembler's code; there is even a special bit of business logic (marked with the TODO that is invalid) that re-adds a package tag

orklah:
Great! I'll keep the error in phpstan and ignore it if necessary until the underlying issue is fixed.

orklah:
Could you also look at
https://github.com/phpDocumentor/phpDocumentor/blob/master/tests/unit/phpDocumentor/Descriptor/ClassDescriptorTest.php#L381

I believe the "andReturn" should take a `new Collection([arguments])` instead of just $argument. Also, I think the mock should be a `phpDocumentor\Descriptor\ArgumentDescriptor` instead of a `phpDocumentor\Descriptor\Tag\ArgumentDescriptor`

Am I making a mistake somewhere?

mvriel:
$arguments (https://github.com/phpDocumentor/phpDocumentor/blob/master/tests/unit/phpDocumentor/Descriptor/ClassDescriptorTest.php#L372) is a misnomer here as it only contains 1 argument. It should indeed be a collection containing that argument

orklah:
> $arguments (https://github.com/phpDocumentor/phpDocumentor/blob/master/tests/unit/phpDocumentor/Descriptor/ClassDescriptorTest.php#L372) is a misnomer here as it only contains 1 argument. It should indeed be a collection containing that argument

Great! And should it be a Tag\ArgumentDescriptor or an ArgumentDescriptor?

orklah:
I think this should pass CI now. After this, I think there will be a last PR in which we should be able to bump to lvl 6

jaapio:
This is an other great PR! I'm currently working on the upgrade of the doctrine coding standard and a lot of fixes in here are fixing issues I'm facing right now.

PR #2251, opened 2020-02-02
Push phpstan to level 6

orklah:
draft for working with CI

### DIFF
--- a/src/phpDocumentor/Compiler/Linker/Linker.php
+++ b/src/phpDocumentor/Compiler/Linker/Linker.php
@@ -13,6 +13,7 @@ declare(strict_types=1);
 
 namespace phpDocumentor\Compiler\Linker;
 
+use ArrayAccess;
 use phpDocumentor\Compiler\CompilerPassInterface;
 use phpDocumentor\Descriptor\ClassDescriptor;
 use phpDocumentor\Descriptor\DescriptorAbstract;
@@ -22,7 +23,10 @@ use phpDocumentor\Descriptor\NamespaceDescriptor;
 use phpDocumentor\Descriptor\ProjectDescriptor;
 use phpDocumentor\Descriptor\TraitDescriptor;
 use phpDocumentor\Reflection\Fqsen;
+use phpDocumentor\Reflection\Types\Compound;
+use Webmozart\Assert\Assert;
 use function get_class;
+use function is_array;
 use function is_iterable;
 use function is_object;
 use function is_string;
@@ -112,11 +116,11 @@ class Linker implements CompilerPassInterface
      * This method will return null if no substitution was possible and all of the above should not update the parent
      * item when null is passed.
      *
-     * @param string|Fqsen|object|iterable $item
-     * @param DescriptorAbstract|null      $container A descriptor that acts as container for all elements
-     *                                                underneath or null if there is no current container.
+     * @param string|Fqsen|object|array<mixed> $item
+     * @param DescriptorAbstract|null          $container A descriptor that acts as container for all elements
+     *                                                    underneath or null if there is no current container.
      *
-     * @return string|DescriptorAbstract|array|iterable|null
+     * @return string|DescriptorAbstract|array<mixed>|null
      */
     public function substitute($item, ?DescriptorAbstract $container = null)
     {
@@ -129,6 +133,8 @@ class Linker implements CompilerPassInterface
         }
 
         if (is_iterable($item)) {
+            Assert::true(is_array($item) || $item instanceof ArrayAccess || $item instanceof Compound);
+
             return $this->substituteChildrenOfCollection($item, $container);
         }
 
@@ -141,8 +147,15 @@ class Linker implements CompilerPassInterface
         return null;
     }
 
+    //phpcs:disable Generic.Files.LineLength.TooLong
+    /**
+     * @param array<string|DescriptorAbstract|null>|(ArrayAccess<array-key, string|DescriptorAbstract>&iterable<array-key, string|DescriptorAbstract>) $collection
+     *
+     * @return array<string|DescriptorAbstract|null>|(ArrayAccess<array-key, string|DescriptorAbstract>&iterable<array-key, string|DescriptorAbstract>)|null
+     */
     private function substituteChildrenOfCollection(iterable $collection, ?DescriptorAbstract $container) : ?iterable
     {
+        //phpcs:enable Generic.Files.LineLength.TooLong
         $isModified = false;
         foreach ($collection as $key => $element) {
             $element = $this->substitute($element, $container);
--- a/src/phpDocumentor/Compiler/Pass/NamespaceTreeBuilder.php
+++ b/src/phpDocumentor/Compiler/Pass/NamespaceTreeBuilder.php
@@ -20,6 +20,7 @@ use phpDocumentor\Descriptor\DescriptorAbstract;
 use phpDocumentor\Descriptor\NamespaceDescriptor;
 use phpDocumentor\Descriptor\ProjectDescriptor;
 use phpDocumentor\Reflection\Fqsen;
+use Webmozart\Assert\Assert;
 use function strlen;
 use function substr;
 use function ucfirst;
@@ -102,13 +103,15 @@ class NamespaceTreeBuilder implements CompilerPassInterface
                 $this->addToParentNamespace($project, $namespace);
             }
 
+            Assert::isInstanceOf($namespace, NamespaceDescriptor::class);
+
             // replace textual representation with an object representation
             $element->setNamespace($namespace);
 
             // add element to namespace
             $getter = 'get' . ucfirst($type);
 
-            /** @var Collection $collection */
+            /** @var Collection<DescriptorAbstract> $collection */
             $collection = $namespace->{$getter}();
             $collection->add($element);
         }
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -127,7 +127,7 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         with:
-          args: analyse src tests --level 5 --configuration phpstan.neon
+          args: analyse src tests --configuration phpstan.neon
 
   build-phar:
     runs-on: ubuntu-latest
--- a/Makefile
+++ b/Makefile
@@ -30,7 +30,7 @@ phpcbf:
 
 .PHONY: phpstan
 phpstan:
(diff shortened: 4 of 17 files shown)


jaapio:
You might want to wait for #2243 because I had to ignore a number of warnings you are fixing right now. It would be awesome if you could address the missing type hints. Since you are already adding a lot of them I didn't want to add those.

orklah:
> You might want to wait for #2243 because I had to ignore a number of warnings you are fixing right now. It would be awesome if you could address the missing type hints. Since you are already adding a lot of them I didn't want to add those.

No worry, I'll fetch your commits and work on it afterwards ;)

jaapio:
My pr is ready and merged, please update your branch to have the new code-style enabled. This should solve your request to have Doctrine coding standard updated. So a proper detection of `use` usage is available.

orklah:
Could you tell me what type is supposed to go in "includes" and "partials" collections?

I couldn't find any example in the code itself...

I think after that, we'll be good for level 6

jaapio:
`includes` are `string`s they are just file names like a normal include in php. 
`partials` are currently unused. I don't know yet if they will be reintroduced. It could be that we introduce that concept again. But if that's the case it would be in a different form. Just handle them as a `string` for now. I will create a new ticket to remove that from the code.

orklah:
Phpstan is ready to be lvl6.

It seems you pass the phpstan level with command line but I'm not sure where I can bump it...

orklah:
Are you sure codestyle is checked correctly in CI? It seems to output the standard output like when we don't push arguments in command line

orklah:
Oops, the build fails in smoke tests. I'll look into it

jaapio:
There are 2 locations:
- https://github.com/phpDocumentor/phpDocumentor/blob/master/.github/workflows/push.yml#L128
- https://github.com/phpDocumentor/phpDocumentor/blob/master/Makefile#L33

If you are able to move that so we only need one location, be my guest :-)

Regarding the codestyle, I will check what is wrong. Thanks for pointing that out.

jaapio:
Fixed the code style check.

orklah:
> Fixed the code style check.

Quick!

I had less luck from my side. The build broke and I don't have a clue what I could change that would have this effect:
https://github.com/phpDocumentor/phpDocumentor/pull/2251/checks?check_run_id=421810264#step:8:1530
Do you have an idea?

src/phpDocumentor/Compiler/Linker/Linker.php
```diff
@@ -141,6 +146,13 @@ public function substitute($item, ?DescriptorAbstract $container = null)
         return null;
     }
 
+    /**
+     * @param array<string|DescriptorAbstract|null>|(ArrayAccess<array-key,
```
jaapio: This param seems to be very complicated. This might be the cause of the failing smoke test. 

But hang in, since I'm working on a fix to solve this issue. Since it was already reported by a user. :-D

orklah: :o I was totally expecting I broke the code, not a notation that was not read! Thanks for catching that!

It is a very complicated type indeed... I think a refactor of this class should use something like an ArrayObject and stick to that. Here we must pass an array or an iterable that is also ArrayAccess.

orklah: This is also causing issues with CS:
https://github.com/phpDocumentor/phpDocumentor/pull/2251/checks?check_run_id=421826643#step:6:18
I don't know how they expect me to put all that in 120 chars and still put variable name after that :D

jaapio: I would agree with a `//phpcs:ignore` above this docblock :-D
The linker is one of the most complicated classes we have. So that's a pain anyhow.

orklah: Ok. I tried `//phpcs:ignore` above the dockblock, I tried `//phpcs:ignore` the docblock and `phpcs:ignore` in the docblock too. Is there something I'm missing here?

jaapio:
`phpcs:ignore rule` rule should be replaced by the violated rule. 

So it would be `phpcs:ignore Generic.Files.LineLength.TooLong` in this case. 
Let me know if it doesn't work. I can have a look later.

orklah:
I can't make it work, even with the rule :(

src/phpDocumentor/Compiler/Linker/Linker.php
```diff
@@ -141,6 +146,12 @@ public function substitute($item, ?DescriptorAbstract $container = null)
         return null;
     }
 
+    //phpcs:ignore Generic.Files.LineLength.TooLong
```
jaapio:
```suggestion
    //phpcs:disable Generic.Files.LineLength.TooLong
    /**
     * @param array<string|DescriptorAbstract|null>|(ArrayAccess<array-key, string|DescriptorAbstract>&iterable<array-key, string|DescriptorAbstract>) $collection
     *
     * @return array<string|DescriptorAbstract|null>|(ArrayAccess<array-key, string|DescriptorAbstract>&iterable<array-key, string|DescriptorAbstract>)|null
     */
    //phpcs:enable Generic.Files.LineLength.TooLong
```

src/phpDocumentor/Compiler/Linker/Linker.php
```diff
@@ -129,6 +132,8 @@ public function substitute($item, ?DescriptorAbstract $container = null)
         }
 
         if (is_iterable($item)) {
+            assert(is_array($item) || $item instanceof ArrayAccess);
```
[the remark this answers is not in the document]
orklah: substituteChildrenOfCollection has a type hint iterable. However, in the code of the method, there is something like $item['somethin'] = $foo; which is only possible on arrays or ArrayAccess implementations. At first, I thought this was array only but it's not. I had to document the ArrayAccess part to allow said syntax, but then, Phpstan complained that I was not pushing an ArrayAccess.

The assert is making sure everything that pass here is either an array or an ArrayAccess so both can do $item['somethin'] = $foo;

jaapio:
@orklah also the fix for the smoke test has landed in master. So if you are able to rebase this PR, we can merge it on a green build :+1:

src/phpDocumentor/Compiler/Linker/Linker.php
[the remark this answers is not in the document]
jaapio: Please switch to the assert lib in this case. our pipeline has an extension installed to get knowledge about that lib. It will give a better error than the native php function.

src/phpDocumentor/Compiler/Linker/Linker.php
```diff
@@ -141,6 +146,18 @@ public function substitute($item, ?DescriptorAbstract $container = null)
...
 
+    //phpcs:disable Generic.Files.LineLength.TooLong
+    /**
+     * @param array<string|DescriptorAbstract|null>|(ArrayAccess<array-key, string|DescriptorAbstract>&iterable<array-key, string|DescriptorAbstract>) $collection
+     *
+     * @return array<string|DescriptorAbstract|null>|(ArrayAccess<array-key, string|DescriptorAbstract>&iterable<array-key, string|DescriptorAbstract>)|null
+     */
+    //phpcs:enable Generic.Files.LineLength.TooLong
+    /**
+     * @param array<string|DescriptorAbstract|null>|(ArrayAccess<array-key, string|DescriptorAbstract>&iterable<array-key, string|DescriptorAbstract>) $collection
```
jaapio: hm github suggestions failed here :-)

src/phpDocumentor/Compiler/Pass/NamespaceTreeBuilder.php
```diff
@@ -102,13 +103,15 @@ protected function addElementsOfTypeToNamespace(ProjectDescriptor $project, arra
                 $this->addToParentNamespace($project, $namespace);
             }
 
+            assert($namespace instanceof NamespaceDescriptor);
```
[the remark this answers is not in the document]
orklah: $project->getIndexes() return `Collection<DescriptorAbstract>`. I didn't find a static analysis magic yet to explain to phpstan that `->get('namespaces', new Collection())` mean it will get a `Collection<NamespaceDescriptor>`. The assert is only there to explain that to Phpstan, but it also controls that there isn't anything we don't want in the Collection. Ok, I'll use the Assert library

src/phpDocumentor/Compiler/Linker/Linker.php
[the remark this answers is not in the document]
orklah: fixed

src/phpDocumentor/Compiler/Linker/Linker.php
orklah: yes, too bad, it was a cool feature !

orklah:
I was not expecting that... Assertion fail on in Linker. assert() must not have been enabled (which is very weird BTW, it's not the only one you have in your code...)

That means there is a type in there that : 
-is an iterable
-is not an array
-is not a class implementing ArrayAccess
-does not break on https://github.com/phpDocumentor/phpDocumentor/blob/bafaa233db2b5a60349de15ae75d1ffbede057e8/src/phpDocumentor/Compiler/Linker/Linker.php#L165

I have no idea what is it. I'll look at it later. Probably tomorrow.

Also CS break again. (I'm starting to regret telling you it was not active :D )

jaapio:
Please ping me if you think @mvriel or me should have a look. I don't want you to spend to much time on the linker :-D

orklah:
Ok, I fixed the issue with Linker by expanding the Assert

However, the weird type was an instance of Compound. It is weird that it's here. PHPStan do not report it yet (probably on higher level) but this is an issue because substituteChildrenOfCollection will expect this instance to implement ArrayAccess to work correctly. 

Secondly, I can't manage to fix CS. PHPStan can't read phpdoc if there's an inline comment between doc and function so I had to put //phpcs::enable lower. This seems to work(but it's ugly) but CS still complains for spacing issue between functions. I think it does not like the //phpcs::disable. I tried to put it in the phpdoc block but it then complains that the order is not right (but lowering the disable would lead to disable too late).

Can I let you look at that?

jaapio:
Thanks for the update. I will have a look at the code style.
I think I need to get in to the first issue you are describing. It sounds like a bug in the linker. But I'm not sure what actually happened in this situation. All additional information you could provide to reproduce this issue would help me a lot. Since you already had a deep dive into this context.

orklah:
Well, I'm not actually sure what's going on, I went printing var_dump in CI to debug :D. All I can tell is 
"Qa workflow / Unit tests [7.2 | ubuntu-latest]" will reproduce the issue. You can add an instanceof Compound in the if(is_iterable()) and it will be triggered.

jaapio:
The changes in this pr are merged into #2265 which also addresses the issue in the linker found here.
Closing this pr on behave of the new pr. :-)

PR #2263, opened 2020-02-04
experiment with psalm detection of unused code

orklah:
I ran Psalm with it's unused code detection on src/

I will run the CI on this and restore all broken functionnality and see what's left.

I know some code is probably needed for BC purposes but it will at least highlight untested code if there is any...

(This is actually the first time I use the option in psalm, I also want to see what it can do :) )

### DIFF
--- a/src/phpDocumentor/Descriptor/ArgumentDescriptor.php
+++ b/src/phpDocumentor/Descriptor/ArgumentDescriptor.php
@@ -64,18 +64,6 @@ class ArgumentDescriptor extends DescriptorAbstract implements Interfaces\Argume
         return $this->type;
     }
 
-    /**
-     * @return list<?Type>
-     *
-     * @phpstan-return array<int, ?Type>
-     */
-    public function getTypes() : array
-    {
-        trigger_error('Please use getType', E_USER_DEPRECATED);
-
-        return [$this->getType()];
-    }
-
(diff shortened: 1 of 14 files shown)


orklah:
I was starting to worry I'd have to revert everything :)

There's still some interesting catches though.

Mind to take a look at it? Is there anything interesting to delete?

jaapio:
I'm not sure if we can use this output. Because we do not test everything in phpDocumentor, so I'm not sure if it would break anything :-)

orklah:
@jaapio you may still want to look at https://github.com/phpDocumentor/phpDocumentor/pull/2263/files#diff-5844ff81b5eb5ff658aabe38d06279fdL70
I'm not sure what the intent was but it seems weird. We cast a Dsn to string a put the result in a variable while being in a try block.

Said variable is not used later so it could be dropped. It leaves only the cast to string in the try catch.

The __toString method of Dsn don't declare throwing Exception and it doesn't seem like it does. Even if it did, throwing exception in __toString is only possible on PHP 7.4 and PhpDocumentor must be compatible with ^7.2.5

I posted this on psalm issues because while this code *could* be valid, I don't think it is in this case.

jaapio:
I think you are right, this catch block should be removed. You can do that in a separate PR. Other things are not removable I think.

orklah:
PR created, I close this one

PR #2264, opened 2020-02-05
drop unused code

orklah:
Drop unused code identified in https://github.com/phpDocumentor/phpDocumentor/pull/2263


jaapio:
Thanks again for your great contribution! We are very happy with all the work you do!

PR #2265, opened 2020-02-05
Orklah phpstan6

jaapio:


jaapio:
@orklah It looks like you found an optimization in your pr of phpstan level 6 :-D

orklah:
yep, I handled that in config file, it's much easier than command line, especially when you had two separate commands ;)

orklah:
How did you get rid on Compound type?

orklah:
> No I was point at: https://github.com/phpDocumentor/phpDocumentor/pull/2265/files#diff-b55c2daf004c2f28e40c91df76f64044R127
> The compount type was an invalid substitute in the linker 👍

oh nice :)
Static analysers are great for that!

jaapio:
Our linker is trying to resolve all FQSEN object to their actual descriptor. Lets say ClassDescriptor for class `A` has a link to ClassDescriptor `B` but to be able to cache that `B` is a `FQSEN` in the parsed structure. The actual link is added by the linker. Which does a lookup for all FQSEN objects. But I think this is only needed for descriptors. The `Type` objects should be ignored.

PR #2269, opened 2020-02-05
Detect null elements in Collection

orklah:
Draft for working with CI


orklah:
The only null I spotted was in a recent test.

Either the issue with null introduction was fixed earlier, or the issue is not reproduced with the tests nor smoke test.

The assertion I added ensure we don't ever add null into Collections but I'm not sure if you want to keep them on production.

jaapio:
I mentioned the filters as a possible hotspot of null values. I need to check what is exactly implemented there. I think it is OK to have the assets in place because that reduces the number of checks we have to do at other locations. If we can get rid of these null value issues that would be very nice. 

Keep up the good work!

orklah:
I'll put that one ready, I'll look at filters on another PR

jaapio:
Thanks for these improvements

PR #2271, opened 2020-02-06
Wrong handling of getInheritedElement for Constant and Property

orklah:
Hi,

While adding types in one of my PR, I stumbled on those:
https://github.com/phpDocumentor/phpDocumentor/blob/master/src/phpDocumentor/Descriptor/ConstantDescriptor.php#L133
https://github.com/phpDocumentor/phpDocumentor/blob/master/src/phpDocumentor/Descriptor/PropertyDescriptor.php#L152

For me, it's not handled correctly.
If I understand everything, here's what happens: 

- getParent on ConstantDescriptor or PropertyDescriptor will return ClassDescriptor, InterfaceDescriptor or null.
- we check if we have a Class or Interface so we exclude null
- getParent on ClassDescriptor could return ClassDescriptor, never InterfaceDescriptor
- getParent on InterfaceDescriptor could return a collection of InterfaceDescriptor, never a single InterfaceDescriptor.

Conclusion: There's no way we can end up with a single instance of InterfaceDescriptor inside the "if".

This PR drops the impossible case.

I think I get the original intent, and it is better handled in MethodDescriptor::getInheritedElement : https://github.com/phpDocumentor/phpDocumentor/blob/master/src/phpDocumentor/Descriptor/MethodDescriptor.php#L220
If I'm right, this behaviour could be extracted in DescriptorAbstract to avoid code repetition

### DIFF
--- a/src/phpDocumentor/Descriptor/ClassDescriptor.php
+++ b/src/phpDocumentor/Descriptor/ClassDescriptor.php
@@ -28,7 +28,7 @@ class ClassDescriptor extends DescriptorAbstract implements Interfaces\ClassInte
     /**
      * Reference to an instance of the superclass for this class, if any.
      *
-     * @var ClassDescriptor|InterfaceDescriptor|Fqsen|string|null $parent
+     * @var ClassDescriptor|Fqsen|string|null $parent
      */
     protected $parent;
 
@@ -72,7 +72,7 @@ class ClassDescriptor extends DescriptorAbstract implements Interfaces\ClassInte
     }
 
     /**
-     * @param ClassDescriptor|InterfaceDescriptor|Fqsen|string|null $parents
+     * @param ClassDescriptor|Fqsen|string|null $parents
      */
     public function setParent($parents) : void
     {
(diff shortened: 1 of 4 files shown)


jaapio:
I think you are right but, let me first explain en write-out the intended behavior.

Docblocks are automatically inherited from parents. It depends on the type of the descriptor were the "parent" docblock can be found. `MethodDescriptor` can inherit information from a parent `ClassDescriptor` when the containing class extends another class. 

```
class A {
  /** 
   *  Parent docblock
   */ 
  public function myMethod()
}

class B extends A {
  //this method will get a docblock from its parent. defined in A
  public function myMethod()
}
```
If A would implement one or more interfaces the `inheritedDescriptor` could also be found in the interfaces. 

But concider this, because this is valid php...

```
interface A
{
    /**
     * @return mixed
     */
    public function a();
}

interface B
{
    /**
     * @return void
     */
    public function a();
}

interface C extends A, B
{
    //What would be the return type :-D
    public function a();
}

class Foo implements C
{
    public function a() {
        return 'AA';
    }
}
```
It would require better testing to do any assumptions on this. Because the same will apply to const. But since properties can only be defined in classes it would be a different implementation. It could be that we can share this between descriptors. However, we do have plans to simplify the descriptors to dump value objects with less behavior and move this logic to a compile pass it would be a useless effort to improve this right now.

orklah:
You're right! I forgot Interfaces couldn't have properties. Also about MethodDescriptor. I'm not so sure about what the current code do in regard of your example...

jaapio:
I'm closing this PR, but I will create a new issue that references this issue. Because we should have a better look at the inheritance behavior of phpdocumentor. It would require some functional tests to prove that we need to rework this.

PR #2272, opened 2020-02-07
nullable get in collection

orklah:
I presume fetching null when specifying a specific offset in collection is never wanted. You either expect some element or specify a default value.

If I'm right, this PR will remove the null return type from Collection::get() and ensure every call to get() will either return a value  or use the default value provided (or throw an Exception)

This will also ensure that 
```php
$coll = new Collection(['a'=>$collA, 'b'=>$collB]);
$coll->c->count()
```

returns a meaningful exception instead of a null pointer error

I create this in draft to see the results of the CI

### DIFF
--- a/src/phpDocumentor/Descriptor/Collection.php
+++ b/src/phpDocumentor/Descriptor/Collection.php
@@ -86,15 +86,22 @@ class Collection implements Countable, IteratorAggregate, ArrayAccess
      *
      * @phpstan-param T $valueIfEmpty
      *
-     * @phpstan-return ?T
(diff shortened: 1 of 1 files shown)


orklah:
Well my assumption was off. There are cases in the code where we test for null after fetching a specific offset. Also a lot of tests are broken too.

Not sure this is worth pursuing for now. It may be better to create a new method for calls that must never return null (kinda like filter() but for a single element).
If this new method was to be created, we will have to investigate if __get should start using it to solve the null pointer issue.

I'm closing this for now

PR #2273, opened 2020-02-07
Static Analysis improvements

orklah:
This PR adds improvements based on Psalm and PHPStan
- It removes an issue ignored in phpstan.neon by changing the syntax of successive merges
- It remove the long and ugly type we had to handle in Linker when we could receive Compound in the method
- It slightly change the filter() template documentation to allow filtering with a subtype (F) of the Collection type(T)
and other small changes


orklah:
Also removed unnecessary ignored files on phpstan

orklah:
I'll rework on this PR after https://github.com/phpDocumentor/phpDocumentor/pull/2275 is resolved. 
That kind of notation https://github.com/phpDocumentor/phpDocumentor/pull/2273/files#diff-560ddfa4810bca66ce8d287a75582337R289 is great for PHPStan but Psalm is complaining it doesn't know the type of the initial Collection

jaapio:
Does it help you to have psalm in ci as well?

orklah:
Not for now, Psalm's config is pretty different from PHPStan.

PHPStan focus on level of increased strictness and is very good at that. You can put it to lowest level and having only simple issues to solve.

Psalm has a much more finer config where each rule is an xml tag but it's less good in handling difficulty on low levels. I still have errors not so simple to handle even when I use the lowest. Given that, I have to use Psalm baseline from the start and I try to solve easy things first, even if it's not the first level. I think it will be better to get to a state where we only have almost unsolvable errors on low level and then put all that in the baseline before adding Psalm in CI.

orklah:
This is ready for review. I think it's best not to create PR too big.

PR #2275, opened 2020-02-07
Collection: Named Constructor

orklah:
Adds a way to generate Collection instance from the class-string which will populate it.

It allows replacing this:
```php
/** @var Collection<RandomClass> */
$coll = new Collection($elements);
```
with 
```php
$coll = new Collection(RandomClass::class, $elements);
```

and still keep every benefit of having a generic'd Collection. For now, the class-string is not saved inside the Collection. If we are one day in need of more control, we could save the class-string and use it to validate every insertion in the Collection.


jaapio:
Nice improvement, this could help at more places to make clear what is happening.

PR #2276, opened 2020-02-07
Remove unnecessary clones

orklah:
This PR removes unnecessary cloning in the code.
The Collection::merge method creates a new Collection with all the items from both Collections and doesn't affect original Collection


jaapio:
Thanks, for this cleanup

PR #2278, opened 2020-02-08
experiment with AssemblerInterface::create

orklah:
It seems to only return Descriptor and never a Collection. If CI confirm that, it means there is some dead code dedicated to handle a Collection returned  from create that could be deleted.

### DIFF
--- a/src/phpDocumentor/Descriptor/Builder/AssemblerInterface.php
+++ b/src/phpDocumentor/Descriptor/Builder/AssemblerInterface.php
@@ -13,9 +13,7 @@ declare(strict_types=1);
 
 namespace phpDocumentor\Descriptor\Builder;
 
-use phpDocumentor\Descriptor\Collection;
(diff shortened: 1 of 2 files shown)


orklah:
Seems like I was right.

Collection in create was introduced in ttps://github.com/phpDocumentor/phpDocumentor/pull/1214
to accomodate a DescriptorCollection returned by TypeCollectionAssembler::create.

However, TypeCollectionAssembler doesn't exist anymore, neither DescriptorCollection.

orklah:
I had to disable CS in the interface because there was an error because create should declare its return type. However, every implementation of the interface already have a Child and I'd have to change all that to just "Descriptor" which would probably cause errors in other places.

It could be revisited if the Interfaces become better or if we support only ^PHP7.4 one day (covariance and contravariance allow to declare compatible types in interfaces and implementations)

mvriel:
Looks good to me! Thanks for checking it out and removing the dead code :sunglasses:

PR #2283, opened 2020-02-10
Check version on Run

mvriel:
With this change, phpDocumentor will fetch the releases from github and
check if the current version is the latest. If it is not, a message will
be shown to ask the user to upgrade.

In this change, I have also refactored away the VERSION file as we now
use PrettyVersions; and I have refactored the Application class away
since it did not contain anything useful or that should actually have
been elsewhere.

- [ ] Add test for RunCommand to actually check whether the version check is done
- [ ] Add test for VersionChecker class
- [ ] Set timeout for CURL to a low value as we don't want to block execution too long when there is no internet
- [ ] Add error handling on the cURL interactions when there is no internet; an INFO message should be emitted but not a warning or error since some run phpDocumentor on an internetless server.


src/phpDocumentor/Console/Command/Project/RunCommand.php
```diff
@@ -269,6 +277,13 @@ protected function execute(InputInterface $input, OutputInterface $output) : int
         $output->writeln('phpDocumentor ' . $this->getApplication()->getVersion());
         $output->writeln('');
 
+        try {
+            $this->releaseChecker->check(Application::VERSION()->getShortVersion());
```
jaapio: maybe we should provide an option to disable this feature on the commandline. For ci environments without an internet connection.

mvriel:
Not going to do this anytime soon; needs rethinking

PR #2285, opened 2020-02-10
php inspections improvements

orklah:
Apply improvements suggested by the PHPStorm native inspections and plugin "PHP Inspections EA".

Those rules can be quite opiniated sometimes, don't hesitate to point me something you don't like, I'll revert it.

I did not run PHPStan after that so I'll fix errors with CI.


orklah:
This is ready for review

jaapio:
Thank you for all the nice improvements.

PR #2287, opened 2020-02-11
small fixes for static analysis

orklah:
I need to make some checks, Then, if everything is okay, I should be able to commit a psalm config file we can use to add psalm in CI

### DIFF
--- a/psalm.xml
+++ b/psalm.xml
@@ -1,0 +1,211 @@
+<?xml version="1.0"?>
+<psalm
+    totallyTyped="false"
+    resolveFromConfigFile="true"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xmlns="https://getpsalm.org/schema/config"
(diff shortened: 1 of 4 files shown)


orklah:
This should pass CI.

I added the psalm.xml file that could be used for CI, it should be error free.

To create the file, I made these steps:
- launch psalm --init on highest level (It's the other way than PHPStan)
- I added some rules as errors at the highest levels
- I added suppression rules for things that I couldn't fix below a `<!-- User defined -->` block
- When everything was fixed, I enabled some other rules that didn't trigger new errors

This should make a good first file.
How we should proceed if we want to increase code quality:
- Try to tackle some of the comments I made in suppressed rules
- Replace "info" by "error" in some rules (preferably starting from the bottom of the file) and fix every error triggered
- Activate "totallyTyped" in header to enable rules forbidding "mixed" type (This is reaaally difficult to fully comply with those rules but it can find errors when Psalm had no vision before)

PR #2288, opened 2020-02-13
getType shouldn't return null

orklah:
Trying to see if anything will break when getType can't return null


orklah:
I'm closing this, I messed up with git. I'll start again more slowly, it seems every class handle its type differently...

Issue #2248, opened 2020-02-01
getTypes should not contain nullable entries


mvriel:
While upgrading PHPStan's coverage to level 6 it reported that entries in the `getTypes` method of the ConstantDescriptor could be null values (https://github.com/phpDocumentor/phpDocumentor/pull/2239#discussion_r373789636); this should not be the case as it makes it harder to iterate that collection.

For this issue, we need to investigate why PHPStan believes there could be null values and ensure there aren't

jaapio:
I'm not sure if null values are even possible with the new reflection component. I think it could be part of our legacy system, needs to be checked to make sure I am right

mvriel:
Yeah, I did not expect this either. So I am really curious how this happens

jaapio:
@orklah I think you already addressed this issue?

orklah:
@jaapio I did not try to tackle this yet but this is a good idea for a next step :)

One of the source cause is there:
https://github.com/phpDocumentor/phpDocumentor/blob/master/src/phpDocumentor/Descriptor/ConstantDescriptor.php#L85
if somehow the instanceof is failing, we will return null, and the method getTypes on this descriptor is just a return [$this->getType()] so it will return null as well.

I don't know if it actually happens though. A couple of Assert and the CI should tell us what's going on

PR #2289, opened 2020-02-13
experiment with getTypes

orklah:


orklah:
Well, best I could do is use the same trick as in TypedAbstract and array_filter the getType result to be sure not to return null.

orklah:
This will close #2248

orklah:
This should be good to merge too if you think it's worth it. The CI fail is not related

orklah:
I don't get it, build fails spectacularly but there's almost nothing on this PR. It didn't seem to affect other PR though...

jaapio:
I think we need to update our ci setup. But due to different priorities I didn't get to that yet. 

No worries as soon as I have fixed the ci I will notify you.

jaapio:
Please rebase, let's see if we can merge this one :-)

orklah:
I really don't get it. I'm starting to think there's something wrong with this specific PR... Other PR pass without issue...

orklah:
replaced by #2315

PR #2290, opened 2020-02-13
experiment with Collection::get

orklah:
I would like to transform Collection::get() so it can never return null again.

There are four reasons for that:
- There are a lot of places where we chain get(). If it can return null, there is a risk of null pointer exception everywhere
- get() is used in __get(). Magic is obscure enough without adding the fact that sometimes an attribute will return null
- It's used with a default value and it will change the state of the Collection (semantically, it should not be a getter)
- Overall, it contributes to having null everywhere, Psalm complains very early and PHPStan will start complaining about that next level I believe.

To do that, I remove the possibility to pass a default value to get() and I check the existence of a value for the given offset. If there's not, an exception is thrown.

To cover all the cases I removed, I added a fetch() method who do pretty much the same thing as the old get() but I'll probably rework it a little. Every case where we use the default value or we expect that the collection may return null, I replaced get() by fetch().
Semantically, it means that every call to fetch() without default value should be checked for null if needed.

This is still a draft. I expect CI will throw some exceptions to me where I forgot to replace some get().

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -15,10 +15,11 @@ parameters:
     - '#Call to an undefined method League\\Flysystem\\Filesystem::find()#'
     # Bad design of descriptors.
     - '#Access to an undefined property phpDocumentor\\Descriptor\\Collection<phpDocumentor\\Descriptor\\Collection<phpDocumentor\\Descriptor\\DescriptorAbstract>>::\$[a-z]+#'
-    # Issue on the content of this collection https://github.com/phpDocumentor/phpDocumentor/pull/2249#issuecomment-581111947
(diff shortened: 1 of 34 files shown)


orklah:
Note: with this change, we're starting to stand on the edge of what can be understood perfectly by both Psalm and PHPStan.

Ideally, fetch() would return either an instance of T (it's class template) or D (where D is a template of it's default value). I'd like them to understand that D is a subtype of T. I'd also like them to understand that if a default value (D) is given, the return type is T|D (without null) so it won't complain chaining calls to fetch if we have a default value.
If I trust the tests I made so far, I'm not sure any tool will react correctly. I'll do more tests to be sure

orklah:
This looks good. We still have a majority of get() so in all those cases, we know there's a content in the collection. For all the other cases, we will either have a default value or I expect static analysis to warn us about a checks if they're missing

PR #2293, opened 2020-02-16
Type definitions

orklah:
Add param and return types when possible (+added a template annotation on Matcher)


orklah:
This should be good to merge

PR #2296, opened 2020-02-23
add strict_types on Guides

orklah:
I think strict_types checks should be part of CS rules so it will never be forgotten


jaapio:
Appart form that is would be required, I'm not sure what the sability of this part is. Is was added as an expiriment. @mvriel could you have a look what to do with this?

mvriel:
Looks good to me; the guides section is excluded from the checks as I prototype in there and the code that I took from Ryan Weaver (with permission) did not match our standards yet. But improvements are always welcome.

PR #2301, opened 2020-03-04
Improvements

orklah:
This PR introduce a MiddlewareInterface to standardize Middlewares in phpDocumentor/Configuration.

I realized later that there's also a Middleware Interface in phpDocumentor/Reflection. They seem unrelated if I'm not mistaken.

I replaced some usage of League/Uri by UriInterface that was already used at some point to standardize.

I replaced the type "callable" by Matcher when applicable. It's much more clear what's going on.

I replaced some assert() usage by Assert also to standardize things between classes.

There's only one class that implements Command so I asserted that every command is a CreateCommand but I expect this to fail at some point. It's very unclear what's done here for me but I think there was an attempt to have other Commands at some point, should we keep this possibility?

EDIT: forgot to create as draft. This is not ready to be merged. And possibly never will :)

### DIFF
--- a/psalm.xml
+++ b/psalm.xml
@@ -12,6 +12,8 @@
         <directory name="src" />
         <ignoreFiles>
             <directory name="vendor" />
+            <!-- Symfony file -->
+            <file name="src/phpDocumentor/Kernel.php"/>
             <!-- This specific file seems a work in progress -->
             <file name="src/phpDocumentor/Guides/Generator/HtmlForPdfGenerator.php"/>
         </ignoreFiles>
@@ -48,6 +50,7 @@
         <PossiblyInvalidIterator errorLevel="info" />
         <PossiblyInvalidMethodCall errorLevel="info" />
         <PossiblyNullArgument errorLevel="info" />
+        <PossiblyNullIterator errorLevel="info" />
         <PossiblyNullOperand errorLevel="info" />
         <PossiblyNullPropertyAssignmentValue errorLevel="info" />
         <PossiblyNullReference errorLevel="info" />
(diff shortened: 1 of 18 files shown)


orklah:
Well, I always forgot to check types on tests.

There's a whole lot of tests that broke because I'm now expecting a MiddlewareInterface instead of any callable. I'll check what can be done about it. Either I can refactor tests to use anonymous classes that extends MiddlewareInterface, or I'll revert this change.

I also have an issue with one Middleware that didn't expect a second param. This is weird because this second param is pushed elsewhere in code. PHP allow passing more parameters than needed but it's ugly. I'll either create a fake Uri to pass or I'll make this parameter optional on all implementations of MiddlewareInterface...

mvriel:
The middlewares were callables on purpose because, originally, we wanted consumers to add their own middlewares and having to extend another class, or implement an interface, might have been difficult.

This is probably no longer the case so making it stricter should not be an issue. However, originally I had intended the Configuration object to be manipulated by the middlewares and have pass the Command along the middleware chain.

We may need to revise which parameters to pass along; as the URL is only used in specific middlewares and it feels more of a workaround.

My recommendation would be to write a documentation page on what these middlewares are and redesign their interface to be consistent. Though, this can be done after merging this PR

mvriel:
Request from me: you are doing multiple things in one PR. Please try to make your PRs as small and focused as possible. 

This will make it easier to review and if one element doesn't fit or needs to be altered, then the whole cannot be merged. If it were separate PRs, we could have merged the other parts and discuss the one that needs elaboration

orklah:
> Request from me: you are doing multiple things in one PR. Please try to make your PRs as small and focused as possible.
> 
> This will make it easier to review and if one element doesn't fit or needs to be altered, then the whole cannot be merged. If it were separate PRs, we could have merged the other parts and discuss the one that needs elaboration

You're right. I should have structured this better. I'll make sure next PR are more focused on one fix.

orklah:
So, I fixed test by making UriInterface optional and by creating callable classes instead of callable functions.

orklah:
I'll close this. It should be splitted before merge and It's not fresh enough in my head to be merged right now anyway.

PR #2303, opened 2020-03-06
add Visibility on ConstantDescriptor and ConstantAssembler

orklah:
This adds methods and tests for Visibilities on Constants for the Descriptor and Assembler


jaapio:
Look good to me

PR #2305, opened 2020-03-07
Fix build

orklah:
On last version, Psalm detects a new error.

It comes from calling getMethods, getConstants and getProperties on a DescriptorAbstract.

DescriptorAbstract implements __call method so Psalm thinks any undeclared method will return __call's return type.

To avoid that, I checked with instanceof each specific Descriptor to let Psalm knows what he's dealing with.


jaapio:
Thanks

PR #2359, opened 2020-05-19
Add Docker to .dependabot/config.yml

HonkingGoose:
Dependabot will check for Docker :whale: updates on the same weekly schedule as the other dependencies. 

The updates will need to **be manually approved and merged** though, as `automerged_updates` are not supported by Dependabot when using Docker.


jaapio:
Thanks for your contribution

PR #2364, opened 2020-05-26
Composer v2.0: remove `--no-suggest flag` as its going to be deprecated

HonkingGoose:
From the upgrade guide for Composer v2.0:
> Deprecated --no-suggest flag as it is not needed anymore

https://github.com/composer/composer/blob/master/UPGRADE-2.0.md#for-composer-cli-users

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -55,7 +55,7 @@ jobs:
 
       - name: Install Composer dependencies
         run: |
-          composer install --no-progress --no-suggest --prefer-dist --optimize-autoloader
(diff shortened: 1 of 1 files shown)


HonkingGoose:
This might not be a problem/issue yet, but it serves as a reminder for when Composer is updated to 2.0

I think the GitHub action uses the latest version of Composer by default, but I might be wrong about that, it's not explicitly defined in the `push.yml` file what version of Composer is used.

jaapio:
Thanks!

HonkingGoose:
Thanks for merging! 😄

PR #2367, opened 2020-06-05
Docs: improvements for "getting started - installing"

jrfnl:
PHP version:
* Refer to the same minimum PHP version as specified in the `composer.json` file.
* Make it clear that it is a _minimum_ version.

Phive:
* As this was rendered as one paragraph, the sentences felt disconnected and confusing.
    Reversing the order and turning it into two paragraphs hopefully fixes that.

Phar:
* Invoking php isn't something you do "simply" if you're relatively new to PHP (or never use the command-line). Ablist language.

Docker:
* I think the "read next" sentence was intended to be at the end of the page ? Not at the end of this section ? Moved it.

Composer:
* Add instruction on how to invoke PHPDocumentor after installation.
    Most other installation methods already mention how to invoke after installation, but the Composer instructions did not.

Read next:
* Change the link to next section within the "Getting started" guide as this page is the first section, so the reader has already read half of it.

Other:
* Use consistent indentation for code samples.
* Minor punctuation fixes.

### DIFF
--- a/docs/getting-started/installing.rst
+++ b/docs/getting-started/installing.rst
@@ -7,7 +7,7 @@ System Requirements
 phpDocumentor has several dependencies on other software packages. Please make sure that you have these
 installed before installing phpDocumentor.
 
--  `PHP 7.2.0`_
+-  `PHP 7.2.5`_ or higher
 -  Graphviz_ (optional)
 -  PlantUML_ (optional)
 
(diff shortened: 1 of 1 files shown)


jrfnl:
Eh.. @jaapio is the build not passing a known issue ? I haven't touched any files which are within the scope of these checks... AFAIK

jaapio:
There seems to be some issue with the setup step. Don't worry about that, I will have a look later

jaapio:
Thanks for these changes.

jrfnl:
@jaapio You're welcome and please do let me know if anything is unclear, needs further improvement or when I've taken something too far, (more PRs upcoming)

PR #2368, opened 2020-06-05
Docs: improvements for "getting started - your first set of documentation"

jrfnl:
Note: _the changes look more extensive than they actually are as I've tried to maintain the line length which roughly appeared to be adhered to._

Writing a DocBlock
* Removed the word "inline" as it can be confusing to people. `//` and `/* */` style comments are generally regarded as inline comments, while docblocks are, well, docblocks, and can not be used "inline", in the sense of trailing comments.

What does a DocBlock look like:
* I've tried to make the language a little more readable. (still quite technical though)

File-level Docblock:
* Changed `function` to `role`.
* Removed some redundant text.

Summary:
* The explanation of when a Summary ends was unclear.
    - What is a dot ? In human writing language, this is called a "period" or a "full stop". Also see: https://www.thepunctuationguide.com/
    - Two subsequent line breaks is in conflict with the statement at the start that every line should start with an asterisk.
        I'm hoping to make it clearer by calling it a "blank (comment) line".

Description:
* Adjusted the phrasing for when a description ends.
    It originally said "when the first tag is encountered", I've qualified this now as "when the first tag is encountered on a new line", to distinguish between
    inline tags and stand-alone tags.
    Still not 100% happy with the phrasing, but it's a start.

Example:
* The lines in the code sample were causing a scroll bar, even on a wide screen.
* Aligned the second line of the parameter description.
* Updated the line numbers referenced to match.
* Aligned the phrasing about the closer with the phrasing used for the opener, i.e. "sequence" not "statement".
    IMO that phrasing is less confusing than statement, as in development a "statement" (in code) has a different meaning.

Closing:
* Added "Read more" subheading.
* Tried to clarify and simplify the paragraph a little again.

Other:
* Various inline markup, grammar and punctuation fixes.
* Limit the line length where relevant.

**Open question**:

Running phpDocumentor
* I personally think that the Hint about the bin folder is misplaced in this tutorial and can confuse people as it is referring to a unsupported method of installation (basically git clone).
    I'd like to suggest to remove it. Alternatively, it could be moved to the "installing" page.
    Note: I have *not* done so in this PR as I'd like a second opinion.

### DIFF
--- a/docs/getting-started/your-first-set-of-documentation.rst
+++ b/docs/getting-started/your-first-set-of-documentation.rst
@@ -4,20 +4,20 @@ Your First Set of Documentation
 Overview
 --------
 
-The goal of this tutorial is to introduce you in writing and subsequently generating effective documentation with
-phpDocumentor.
+The goal of this tutorial is to introduce you to writing and subsequently generating effective documentation
+with phpDocumentor.
 
 Writing a DocBlock
 ------------------
 
-A DocBlock is a piece of inline *documentation* in your source code that informs you what a class, method or other
-Structural Element its function is.
+A DocBlock is a piece of *documentation* in your source code that informs you what the function of
+a certain class, method or other *Structural Element* is.
 
 Which elements can be documented?
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Before we discuss what a DocBlock looks like, let's first zoom into what you can document with them. phpDocumentor
-follows the PHPDoc definition and recognizes the following Structural Elements:
+Before we discuss what a DocBlock looks like, let's first zoom in on what you can document with them.
+PhpDocumentor follows the PHPDoc definition and recognizes the following *Structural Elements*:
 
 * Function_
 * Constant_
@@ -28,19 +28,19 @@ follows the PHPDoc definition and recognizes the following Structural Elements:
 * Property_
 * Method_
 
-In addition to the above the PHPDoc standard also supports DocBlocks for *Files* and include/require statements,
-even though PHP itself does not know this concept.
+In addition to the above, the PHPDoc standard also supports DocBlocks for *Files* and ``include``/``require``
+statements, even though PHP itself does not know this concept.
 
-Each of these elements can have exactly one DocBlock associated with it, which directly precedes it. No code or
-comments may be between a DocBlock and the start of an element's definition.
+Each of these elements can have exactly one DocBlock associated with it, which directly precedes it.
+No code or comments may be between a DocBlock and the start of an element's definition.
 
 What does a DocBlock look like?
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-DocBlocks are always enclosed in a comment-type, called DocComment, that starts with ``/**`` and ends
-with ``*/``. Each line in between the opening and closing statement should start with an asterisk (``*``). Every
-DocBlock precedes exactly one Structural Element and all contents of the DocBlock apply to that
-associated element.
+DocBlocks are always enclosed in a particular comment-type, called *DocComment*. A DocComment starts
+with ``/**`` (opener) and ends with ``*/`` (closer). Each line in between the DocComment opener and
+its closer should start with an asterisk (``*``). Every DocBlock precedes exactly one *Structural Element*
+and all contents of the DocBlock apply to that associated element.
 
 For example:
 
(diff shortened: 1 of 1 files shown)


jrfnl:
I'm closing this PR temporarily. Some other PRs go across the complete documentation and would conflict with this PR.
I want to give priority to those PRs and will rebase this PR to solve the conflicts after those PRs have been merged (or rejected).

jrfnl:
Rebased, updated & reopened.

mvriel:

docs/getting-started/your-first-set-of-documentation.rst
```diff
@@ -58,10 +58,10 @@ For example:
 
    **File-level DocBlocks**
 
-   Quite often projects will want to document the license or function for an entire file instead of a single element.
-   This can be accomplished by having a DocBlock as the first element encountered in a file. It is important to note that
-   whenever another Structural Element directly follows the DocBlock that it is no longer recognized as a
-   File-level DocBlock but belonging to the subsequent element.
+   Quite often projects will also want to document the license or role for an entire file.
+   This can be accomplished by having a DocBlock as the first element encountered in a file.
+   It is important to note that whenever another *Structural Element* directly follows the DocBlock
```
ashnazg, line 63: s/ 
follows the DocBlock
that it is
/ 
follows the DocBlock,
it is
/

docs/getting-started/your-first-set-of-documentation.rst
```diff
@@ -96,22 +96,22 @@ DocBlocks are divided into the following three parts. Each of these parts is opt
...
 Summary
-  Sometimes called a short description, provides a brief introduction into the function of the associated element.
-  A Summary ends
-  in one of these situations:
+  The Summary, sometimes called a short description, provides a brief introduction into the function
+  of the associated element. A Summary ends when it encounters either of the below situations::
 
-    1. A dot is following by a line break, or
-    2. Two subsequent line breaks are encountered.
+    1. A period ``.``, followed by a line break, or
```
ashnazg, line 102: I think the `1. / 2.` implies a list that itself implies _do **all** of these_ rather than _either / or_, so maybe switch to `*`

```
* a period `.` followed by a line break
* or a blank comment line
```

docs/getting-started/your-first-set-of-documentation.rst
```diff
@@ -195,18 +197,21 @@ For example::
...
-option, in which case the output will be written to a subfolder called ``output``.
+What the above example does, is scan all files in the ``src`` directory and its subdirectories, perform
+an analysis and generate a website containing the documentation in the folder ``docs/api``. If you want,
+you can omit the ``-t`` option, in which case the output will be written to a subfolder called ``output``.
 
-.. hint::
+Read more
+~~~~~~~~~
+
+PhpDocumentor features several templates_ with which you can change the appearance of your documentation.
```
ashnazg, line 207: Do we still have "several" templates?

docs/getting-started/your-first-set-of-documentation.rst
```diff
@@ -96,22 +96,22 @@ DocBlocks are divided into the following three parts. Each of these parts is opt
 may not exist without a Summary.
 
 Summary
-  Sometimes called a short description, provides a brief introduction into the function of the associated element.
-  A Summary ends
-  in one of these situations:
+  The Summary, sometimes called a short description, provides a brief introduction into the function
+  of the associated element. A Summary ends when it encounters either of the below situations::
```
ashnazg, line 100: I noticed the double colon here earlier, but assumed it was formatting... is it actually just a duplicate?

jrfnl: The colon has special meaning in the markup language used, so for a literal colon it needs to be doubled in most cases.

ashnazg: I wasn't looking for that previously... does it need to be doubled above on line 82 also?

jrfnl: Looking at the generated docs, that one seems to be fine, though doubling it probably doesn't do any harm either.

docs/getting-started/your-first-set-of-documentation.rst
jaapio: We do, however we are dropping most of them :-)

ashnazg (approved):

docs/getting-started/your-first-set-of-documentation.rst
mvriel: The double colon actually means: the next quoted or indented block is a Literal Block :) I think a slash is the one for escaping and I am not even sure that you need to escape a single colon

PR #2370, opened 2020-06-06
Template Default: normalize the CSS section headers

jrfnl:
Saving the files as utf-8 is one thing, using unicode characters as part of the chapter markers is another (and not necessary).

I realize the diff may be confusing, so attaching some screenshots from an editor which doesn't support unicode:

**Before**:
image

**After**:
image

### DIFF
--- a/data/templates/default/css/base.css.twig
+++ b/data/templates/default/css/base.css.twig
@@ -71,7 +71,7 @@
 }
 
 /* Grid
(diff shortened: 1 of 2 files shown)


jaapio:
Thanks, small steps to make the developer experience better!

jrfnl:
@jaapio Yes, this one seemed a bit trivial, but I noticed it while testing something else, so figured I may as well fix it.

PR #2379, opened 2020-06-07
CI: allow Cypress tests to pass

jrfnl:
This:
* Fixes one test - class-details / "Show methods with return type in the Table of Contents"
* Skips four tests which are currently completely broken.
* Partially skips two test which are currently partially broken.

The (partially) skipped tests examine the breadcrumbs and look for a `Home` link which doesn't exist.
These tests should either be removed or the behaviour should be fixed.

I'm kind of assuming this removal of the "Home" breadcrumb is intentional and the tests should be removed, but I didn't want to do so without verifying this. Please let me know if I can update the PR to remove these tests.

### DIFF
--- a/cypress/integration/default/class-details.spec.js
+++ b/cypress/integration/default/class-details.spec.js
@@ -7,13 +7,16 @@ describe('Class Detail Page', function() {
         cy.get('.phpdocumentor-content__title').contains("Pizzeria");
     });
 
-    it('Has a breadcrumb featuring "Home" and "Marios"', function() {
-        cy.get('.phpdocumentor-breadcrumbs').contains("Home");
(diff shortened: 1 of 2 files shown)


jaapio:
I think it is ok for now to have the tests available. @mvriel is still working on the actual design and optimizing it to make it better. 
So I'm going to merge this PR as is. We can have a look later to fix these tests. 

Since this is the first PR that needed a comment, I'm thanking you so much for all the small changes you are currently doing. This helps us so much. Thanks for doing this!

jrfnl:
@jaapio You're very welcome ;-)

As we discussed of GH, I'm also going to be sending in a slew of bug reports/feature requests soon and I'm still old school: when you ask a question in an open source project, make sure you also answer at least some too :sunglasses:  Or in this case: fix some things you know how to fix.

PR #2380, opened 2020-06-07
Docs: remove `:term:` annotations

jrfnl:
_Discussed with @jaapio prior to creating this PR._

Commit https://github.com/phpDocumentor/phpDocumentor/commit/2d7a2a6af470b7a79f3d31006cb51da7e0ab620e removed the `glossary.rst` file.

The net result of that was that the automatic linking to the Glossary via `:term:...` annotations was removed. However, due to something inscrutable in Sphinx, the "_:term:`Type`_" references were now being auto-linked to the Python documentation, which I don't believe was the intention.
This effect can be clearly seen on: https://docs.phpdoc.org/3.0/references/phpdoc/types.html

To prevent this unintentional linking, this commit removes all `:term:` annotations throughout the documentation.

If needs be, they can always be brought back in the future by reverting this commit.

Side-note: this also solves a significant amount of the warnings the build was generating - see: https://github.com/phpDocumentor/phpDocumentor/runs/747054719?check_suite_focus=true#step:5:1973

### DIFF
--- a/docs/getting-started/changing-the-look-and-feel.rst
+++ b/docs/getting-started/changing-the-look-and-feel.rst
@@ -5,7 +5,7 @@ Overview
 --------
 
 With this tutorial, I want to show you how the look and feel of phpDocumentor can be changed using one of the
-existing :term:`templates` or by selecting a custom-made :term:`template`.
+existing templates or by selecting a custom-made template.
 
 What is a template?
 -------------------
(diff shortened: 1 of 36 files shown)


jaapio:
Could you please rebase this PR so we have 1 commit to revert this PR when we want to reintroduce the glossary at some point. I'm still a bit in doubt about removing it. Maybe we never will reintroduce it but for now, it is ok like this.

jrfnl:
Rebased to remove imaginary conflict.

PR #2385, opened 2020-06-08
Template files: fix various links

jrfnl:
### Template files: fix various links

Verified all remaining links in template files and updated them where necessary.

Note: I've stayed away from updating links in the docblocks of external dependencies, like JS libraries. Those should be automatically updated when those libraries are updated (which they probably should be).

### Template files: remove a broken link

This link would be added to documentation websites via the template, but the URL is no longer valid.

As this is a URL on the phpdoc.org website, this may need investigation/fixing instead of the link being removed.


jaapio:
The responsive template is removed. so it should not be linked in any way :+1:

PR #2386, opened 2020-06-08
File docblocks: fix license tags

jrfnl:
The URL used in the `@license` tags in various places was outdated. Fixed now.

~~Alternatively these could be removed as these tags only seems to exist in a very limited number of files in the codebase.~~

**Updated PR**: As these tags only seems to exist in a very limited number of files in the codebase, they have now been removed.

### DIFF
--- a/bin/phpdoc.php
+++ b/bin/phpdoc.php
@@ -6,7 +6,6 @@
  * PHP Version 5.3
  *
  * @copyright 2010-2018 Mike van Riel / Naenius (http://www.naenius.com)
- * @license   http://www.opensource.org/licenses/mit-license.php MIT
  * @link      https://phpdoc.org
  */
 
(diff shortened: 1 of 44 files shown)


jaapio:
You are right these links should be removed. The files you updated are not covered by our cs checks. Since most of them are test assets.

jrfnl:
@jaapio I will update the PR to remove most of these. I'm a bit wary removing all as it may be removing legitimate test case data to test the handling of the `@license` tag.

jrfnl:
@jaapio I've updated the PR to remove the tags instead. I've locally ran the PHPUnit tests and those stil seem to pass without problem.

Please note that I have not ran the Cypress or Behat tests, so some critical checking of both the CI output and if these aren't run via CI, running of these by a dev with a working setup would be highly recommended and appreciated!@jaapio I've updated the PR to remove the tags instead. I've locally ran the PHPUnit tests and those stil seem to pass without problem.

Please note that I have not ran the Cypress or Behat tests, so some critical checking of both the CI output and if these aren't run via CI, running of these by a dev with a working setup would be highly recommended and appreciated!

PR #2388, opened 2020-06-09
add template annotations on Compiler

orklah:
This should fix some errors reported by phpstan on the last build.

The last error on the build seems related to https://github.com/phpstan/phpstan/issues/3415. Let's see if it's fixed soon, otherwise I'll create a PR to ignore this error.


orklah:
Hmm, Psalm doesn't seem to be aware of the template on SplPriorityQueue. Solving the issue for phpstan creates one for Psalm.

I'll investigate a little to see if there are simple solution, otherwise this could be ignored in psalm configs, Phpstan seems more detailed in this case.

orklah:
I ignored the issue in Psalm. It could be added as a stub to add the missing functionality but I don't know the class well enough to try this right now

PR #2397, opened 2020-06-14
Docs: improvements for "references - phpdoc - tags - copyright"

jrfnl:
Other:
* Minor text synchronization with PSR 19.

Ref: https://github.com/php-fig/fig-standards/blob/master/proposed/phpdoc-tags.md#53-copyright

**Open question**:

I wonder if the "Effects in phpDocumentor" section is still correct as in the current "Default" template, I don't see any `@copyright` tags displayed.
* Is this on the to do list ?
* Is this no longer supported ?
* Should the section text be updated ?
* Should an issue be opened about support for this tag in phpDocumentor 3 ?

### DIFF
--- a/docs/references/phpdoc/tags/copyright.rst
+++ b/docs/references/phpdoc/tags/copyright.rst
@@ -1,18 +1,20 @@
 @copyright
 ==========
 
-The @copyright tag is used to document the copyright information for
-Structural Elements.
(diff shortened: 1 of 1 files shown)


jaapio:
Also for this one please create a new issue for the template support. I think this is just not implemented yet in the template.

PR #2402, opened 2020-06-14
Docs: improvements for "references - phpdoc - tags - license"

jrfnl:
Description:
* Rephrased the "NOT RECOMMENDED" phrase to use "RECOMMENDED ... ONLY" for clarity.

Effects in phpDocumentor:
* Minor text clarification.

Example:
* Added the PHP open tag to reinforce the recommendation to only use the tag in a file-level PHPDoc.
* Added a blank comment line to make it clearer that these are alternatives and not tags to be used together in the same PHPDoc.

Other:

### DIFF
--- a/docs/references/phpdoc/tags/license.rst
+++ b/docs/references/phpdoc/tags/license.rst
@@ -1,33 +1,34 @@
 @license
 ========
 
-The @license tag is used to indicate which license is applicable for the associated
-Structural Elements.
+The ``@license`` tag is used to indicate which license is applicable for the associated
+*Structural Elements*.
 
 Syntax
 ------
+
+.. code-block::
 
     @license [<url>] [name]
 
 Description
 -----------
 
-The @license tag provides the user with the name and URL of the license that is
-applicable to Structural Elements and any of their child elements.
+The ``@license`` tag provides the user with the name and URL of the license that is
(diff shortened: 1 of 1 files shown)


docs/references/phpdoc/tags/license.rst
```diff
@@ -1,33 +1,34 @@
...
 
-The @license tag provides the user with the name and URL of the license that is
-applicable to Structural Elements and any of their child elements.
+The ``@license`` tag provides the user with the name and URL of the license that is
+applicable to a *Structural Element* and any of its child elements.
 
-It is NOT RECOMMENDED to apply @license tags to any PHPDoc other than
-file-level PHPDocs as this may cause confusion which license applies at which
-time.
+It is RECOMMENDED to apply @license tags ONLY to file-level PHPDocs, to prevent
```
jaapio, line 20: `NOT RECOMMENDED` was used because php-fig is following this rfc: https://www.ietf.org/rfc/rfc2119.txt

I would like to keep it that way for now.

jrfnl: The license tag is no longer in PSR19, but the sentence as was, contains an implied double negative.

`RECOMMENDED` is part of the RFC you reference, it is just the `ONLY` which isn't, but then can be made proper case to still comply with the RFC.

PR #2403, opened 2020-06-14
Docs: improvements for "references - phpdoc - tags - link"

jrfnl:
Summary:
* Moved the note about support in phpDocumentor to the "Effects in phpDocumentor" section.

Effects in phpDocumentor:
* Split the text into two paragraphs for improved readability.

Example:
* Modernized the Type used in the `@return` tag.

Other:
* Verified the text with PSR 19.
    This also fixes the "internal" syntax resulting in an invalid email link.

Ref: https://github.com/php-fig/fig-standards/blob/master/proposed/phpdoc-tags.md#56-link

**Open question**:

I wonder if the note about support in phpDocumentor (now moved to "Effects in phpDocumentor") is still correct.
* Based on the specs the `@link` tag should only have support for URIs, not FQSEN, so why mention that as missing support ?
* IIRC, both stand-alone as well as inline `@link` tags seem to be supported in phpDocumentor 3, though some tweaking may still be needed, so is this note still relevant at all ?
* And if it is, it may be more helpful to be more specific about what is and isn't supported, so what can you tell me as input to update this ?

### DIFF
--- a/docs/references/phpdoc/tags/link.rst
+++ b/docs/references/phpdoc/tags/link.rst
@@ -1,44 +1,48 @@
 @link
 =====
 
-.. important::
-
-   The effects of the inline version of this tag are not yet fully implemented
-   in phpDocumentor 3. There's only URI support (i.e. no support for
-   Structural Elements), and even that is available only in long descriptions.
-
-The @link tag indicates a custom relation between associated
-Structural Elements and a website, which is identified by an absolute
-URI.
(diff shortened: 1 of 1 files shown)


jrfnl:
@jaapio Any feedback on the open questions ?

jaapio:
Uh, oh excuse me. I will extract those to an issue, because it might need some discussion.

PR #2404, opened 2020-06-14
Docs: improvements for "references - phpdoc - tags - method"

jrfnl:
Description:
* Minor textual improvements.
* Added links to the PHP manual pages covering the relevant magic methods.
    As the link _text_ starts with an underscore, I wasn't sure having those links at the bottom of the page would work as there they would need to be prefixed with an underscore, so, in this case, for these specific links, I've left them inline.
* Rephrased the "MUST NOT" phrase to use "can ONLY" for clarity.

Examples:
* Modernized the Types used in the tags.

Other:

Ref: https://github.com/php-fig/fig-standards/blob/master/proposed/phpdoc-tags.md#57-method

### DIFF
--- a/docs/references/phpdoc/tags/method.rst
+++ b/docs/references/phpdoc/tags/method.rst
@@ -1,45 +1,50 @@
 @method
 =======
 
-The @method allows a class to know which 'magic' methods are callable.
+The ``@method`` tag allows a class to know which 'magic' methods are callable.
 
 Syntax
 ------
+
+.. code-block::
 
     @method [[static] return type] [name]([[type] [parameter]<, ...>]) [<description>]
 
 Description
 -----------
 
-The @method tag is used in situation where a class contains the ``__call()``
-or ``__callStatic()`` magic method and defines some definite uses.
+The ``@method`` tag is used in situations where a class contains the
+`__call() <https://www.php.net/language.oop5.overloading#object.call>`_ or
+`__callStatic() <https://www.php.net/language.oop5.overloading#object.callstatic>`_
+magic method and defines some definite uses.
 
-An example of this is a child class whose parent has a __call() to have dynamic
-getters or setters for predefined properties. The child knows which getters and
-setters need to be present but relies on the parent class to use the __call()
-method to provide it. In this situation, the child class would have a @method
-tag for each magic setter or getter method.
+An example of this is a child class whose parent has a ``__call()`` method defined
+to have dynamic getters or setters for predefined properties. The child knows
+which getters and setters need to be present, but relies on the parent class to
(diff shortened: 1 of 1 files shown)


docs/references/phpdoc/tags/method.rst
```diff
@@ -1,45 +1,50 @@
...
 
 If the intended method is static, the ``static`` keyword can be placed before
 the return type to communicate that.
-There must be a return type, ``static`` on its own would mean that the method
-returns an instance of the child class which the method is called on.
+In that case, a return type MUST be provided, as ``static`` on its own would mean
+that the method returns an instance of the child class which the method is called on.
 
-@method tags MUST NOT be used in a PHPDoc that is not associated with
+``@method`` tags can ONLY be used in a PHPDoc that is associated with
```
jaapio, line 39: Same rfc term's are here applicable https://www.ietf.org/rfc/rfc2119.txt

jrfnl: Except that the sentence as it was contained a double negative... decreasing readability and making it hard to understand.

Note: The adjusted text is in line with PSR19 where this was already fixed two years ago in https://github.com/php-fig/fig-standards/commit/b44a5fccab69eceff95e098284e1f425cf4e3750

mvriel:

docs/references/phpdoc/tags/method.rst
```diff
@@ -1,45 +1,50 @@
...
+An example of this is a child class whose parent has a ``__call()`` method defined
+to have dynamic getters or setters for predefined properties. The child knows
+which getters and setters need to be present, but relies on the parent class to
+use the `__call() <https://www.php.net/language.oop5.overloading#object.call>`_
+method to provide this functionality. In this situation, the child class would have
+a ``@method`` tag for each magic setter or getter method.
 
-The @method tag allows the author to communicate the type of the arguments and
-return value by including those types in the signature.
+The ``@method`` tag allows the author to communicate the :doc:`type <../types>` of
```
ashnazg, line 28: Capitalize `Type` here, to be consistent with other PR changes?

docs/references/phpdoc/tags/method.rst
ashnazg: Ah, so `ONLY` is not a defined keyword in that RFC.  Let's try this in order to lose the double negative:

`@method tags MUST NOT be used in a PHPDOC unless it is associated with`

I'll need to fix this in PSR-19 too.

jrfnl: Fixed

ashnazg (approved):

PR #2406, opened 2020-06-14
Docs: improvements for "references - phpdoc - tags - param"

jrfnl:
Summary:
* Added a link to the PHP manual page covering function arguments.

Description:
* Removed a redundant phrase - "Exceptions to this recommendation are:". Might have been orphaned after a previous edit ?

Example:
* Added a second argument with a different Type and a multi-line description.

Other:
* Verified the text with PSR 19.

Ref: https://github.com/php-fig/fig-standards/blob/master/proposed/phpdoc-tags.md#59-param

**Open question**:

The specs feel outdated. If the function declaration has parameter type declarations, those should be leading and the tags can be omitted.
The specs currently do not explicitly allow for that.

In PSR-19, the following phrase has been added to the first paragraph of the description, which sort of allows for that.
Should this phrase be added or not ?

> The "name" is required only when some @param tags are omitted due to all useful info
> already being visible in the code signature itself.

### DIFF
--- a/docs/references/phpdoc/tags/param.rst
+++ b/docs/references/phpdoc/tags/param.rst
@@ -1,56 +1,58 @@
 @param
 ======
 
-The @param tag is used to document a single argument of a function or method.
+The ``@param`` tag is used to document a single `argument of a function or method`_.
 
 Syntax
 ------
+
+.. code-block::
 
     @param [Type] [name] [<description>]
 
 Description
 -----------
 
-With the @param tag it is possible to document the type and function of a
-single argument of a function or method. When provided it MUST contain a
-Type to indicate what is expected; the description on the other hand is
-OPTIONAL yet RECOMMENDED in case of complicated structures, such as associative
-arrays.
+With the ``@param`` tag it is possible to document the :doc:`Type <../types>`
+and function of a single `argument of a function or method`_. When provided it
+MUST contain a :doc:`Type <../types>` to indicate what is expected.
+The description is OPTIONAL yet RECOMMENDED, for instance, in case of
+complicated structures, such as associative arrays.
 
-The @param tag MAY have a multi-line description and does not need explicit
+The ``@param`` tag MAY have a multi-line description and does not need explicit
 delimiting.
 
 It is RECOMMENDED when documenting to use this tag with every function and
-method. Exceptions to this recommendation are:
+method.
 
 This tag MUST NOT occur more than once per argument in a PHPDoc and is
-limited to Structural Elements of type method or function.
+limited to *Structural Elements* of type method or function.
 
 Effects in phpDocumentor
 ------------------------
 
-Structural Elements of type method or function, that are tagged with the
-@param tag, will have additional information in the section regarding Arguments.
+*Structural Elements* of type method or function, that are tagged with the
+``@param`` tag, will have additional information in the section regarding Parameters.
 
-If the return Type is a class that is documented by phpDocumentor, then
-a link to that class' documentation is provided.
+If the parameter *Type* is a class that is documented by phpDocumentor,
+then a link to that class' documentation is provided.
 
 .. note::
 
-   phpDocumentor supports @param tags which omit the name, this is
+   phpDocumentor supports ``@param`` tags which omit the name, this is
    NOT RECOMMENDED but provided for compatibility with existing projects.
 
 .. note::
 
-   phpDocumentor will try to analyze correct usage and presence of the @param
+   phpDocumentor will try to analyze correct usage and presence of the ``@param``
    tag; as such it will provide error information in the following scenarios:
 
-   * An @param is provided but no argument was found matching the @param.
-   * The name of the @param does not match the name of argument at the same
+   * A ``@param`` is provided but no argument was found matching the ``@param``.
+   * The name of the ``@param`` does not match the name of argument at the same
      position.
    * A mismatch between the type hint, if present, and the type declaration was
      detected.
-   * The type declaration is *type*; this is an invalid type and often provided
+   * The type declaration is ``type``; this is an invalid type and often provided
      by IDEs.
 
 Examples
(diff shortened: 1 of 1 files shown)


jrfnl:
Per https://github.com/phpDocumentor/phpDocumentor/issues/2444#issuecomment-644944907 , phpDocumentor 3 does not currently contain any verification of the `@param` tags.

I've added a second commit to this PR to update the documentation to reflect this.

docs/references/phpdoc/tags/param.rst
```diff
@@ -1,57 +1,53 @@
...
+and function of a single `argument of a function or method`_. When provided it
+MUST contain a :doc:`Type <../types>` to indicate what is expected.
+The description is OPTIONAL yet RECOMMENDED, for instance, in case of
+complicated structures, such as associative arrays.
 
-The @param tag MAY have a multi-line description and does not need explicit
+The ``@param`` tag MAY have a multi-line description and does not need explicit
 delimiting.
 
 It is RECOMMENDED when documenting to use this tag with every function and
```
jaapio, line 25: @mvriel what do you think? should we rephrase this?  This recommendation was from the time before typehints in php. 

I would recommend usage of a `@param` when the type is a collection, or cannot be expressed by a native typehint in php. The tag should also be used when the purpose of a parameter is unclear, like boolean values, which are often featuere flags.

jrfnl: Side-note: as discussed off GH, I intend to pull any significant changes we make to the tag descriptions to PSR-19, so if we can come up with a better description, let's :+1: 

Also see: https://twitter.com/ashnazg/status/1273572427935756290

mvriel: I agree that we can relax the recommendation; though I would focus on the description side of things as the main function is shifting more and more towards documenting the argument's function rather than its type. I believe we can win so much more in phpDocumentor itself as well if we were to document our own elements better

docs/references/phpdoc/tags/param.rst
```diff
@@ -1,57 +1,53 @@
...
 Description
 -----------
 
-With the @param tag it is possible to document the type and function of a
-single argument of a function or method. When provided it MUST contain a
-Type to indicate what is expected; the description on the other hand is
-OPTIONAL yet RECOMMENDED in case of complicated structures, such as associative
-arrays.
+With the ``@param`` tag it is possible to document the :doc:`Type <../types>`
+and function of a single `argument of a function or method`_. When provided it
```
ashnazg, line 17: `function` already is a domain term, so maybe `intent` of the arg would be better wording?

docs/references/phpdoc/tags/param.rst
```diff
@@ -1,57 +1,53 @@
...
+MUST contain a :doc:`Type <../types>` to indicate what is expected.
```
ashnazg, line 18: I think we can indeed relax the `Type` to `MAY`, though I think `name` has to be required so that the parser can know how to tie a given param to what's in the code signature, in the case where some params are omitted in the docblock:

```
/**
 * I'm a FOO BAR function
 *
 * @param this array should contains strings... 
 *        but with no Type or Name given,
 *        how does phpDocumentor know I'm tied to $baz?
 * @return void
 */
public function foo (string $bar, array $baz, object $quz)
```

docs/references/phpdoc/tags/param.rst
ashnazg: I agree with the relaxing:

`At a minimum, it is RECOMMENDED to use this tag with each argument whose code signature does not provide Type information.`

Giving a Description has always been at the discretion of the developer, and the Name is only needed to remove all ambiguity over which argument belongs to that param tag.  But Type is important, and if it's not in the code signature, we should stress the importance of including it in the param tag.

docs/references/phpdoc/tags/param.rst
```diff
@@ -1,57 +1,53 @@
...
 
-If the return Type is a class that is documented by phpDocumentor, then
-a link to that class' documentation is provided.
+If the parameter *Type* is a class that is documented by phpDocumentor,
+then a link to that class' documentation is provided.
 
 .. note::
 
-   phpDocumentor supports @param tags which omit the name, this is
+   phpDocumentor supports ``@param`` tags which omit the name, this is
```
ashnazg, line 42: Are we able to do this only in cases where there are param tags given for every arg in the code signature?  If so, we should highlight that requirement... otherwise we're saying here that my earlier `foo()` isn't true, and that we _can_ resolve this arg the one param is tied to.

So:
`phpDocumentor supports @param tags which omit the name, but only if every argument is given a param tag.`

docs/references/phpdoc/tags/param.rst
```diff
@@ -1,57 +1,53 @@
...
-   * The name of the @param does not match the name of argument at the same
-     position.
-   * A mismatch between the type hint, if present, and the type declaration was
-     detected.
-   * The type declaration is *type*; this is an invalid type and often provided
-     by IDEs.
+   phpDocumentor will not verify whether a documented type is in line with
+   a declared type.
+   Such verification, if desired, can be executed by static analyzer tools
+   like `PHPStan`_ or `PHP_CodeSniffer`_.
```
ashnazg, line 50: Does Psalm do this too?

docs/references/phpdoc/tags/param.rst
```diff
@@ -62,11 +58,19 @@ Examples
     /**
```
ashnazg, line 58: I'd prefix this example with `Here is a fully formed example, where all arguments are given param tags even when not strictly required.`

Then, below the example, I'd suggest a second example showing the minimum required:

```
    /*
     * Counts the number of items in the provided array.
     *
     * @param $items     Array structure to count the elements of.
     * @param $recursive Whether or not to recursively count elements in nested arrays.
     *
     * @return int Returns the number of elements.
     */
    function count(array $items, bool $recursive = false)
```

docs/references/phpdoc/tags/param.rst
jaapio: Yes they do

mvriel:
I am merging this because it is an improvement nonetheless and I will resolve the feedback on master :)

jrfnl:
@mvriel Sorry, I hadn't realized there was consensus yet on what the phrasing should be (which is why I hadn't updated the PR yet).

mvriel:
No problem! I felt it was time to merge this PR and we can always open a new one if there are more changes needed :) I already liked these :)

PR #2407, opened 2020-06-14
Docs: improvements for "references - phpdoc - tags - property[<-read|-write>]"

jrfnl:
I've combined the contents of the `@property`, `@property-read` and `@property-write` pages into _one_ documentation page as it makes more sense to read about these tags together.
The added benefit is, that this removed some (near) duplication in the documentation.

Note:
* The pages will still show up individually in the table of contents, with the title for the `@property` documentation page being adjusted for the purposes of the TOC.
* Existing (external/internal) links to these pages will not be broken as the `@property-read` and `@property-write` pages still exist, they will now just show the same contents as the `@property` page.

Summary:
* Synced the text with the current text in PSR-19.
* Added a link to the PHP manual page covering class properties.

Syntax:
* Adjusted the specs to include `@property-read` and `@property-write`.

Description:
* Based the text on PSR-19, with a rewritten "example" from the original text.
* Added a table to make it more obvious when to use which tag.
* Added links to the PHP manual pages covering the relevant magic methods.
    As the link _text_ starts with an underscore, I wasn't sure having those links at the bottom of the page would work as there they would need to be prefixed with an underscore, so, in this case, for these specific links, I've left them inline.
* Rephrased the "MUST NOT" phrase which contained a double negative to use "can ONLY" for clarity.

Effects in phpDocumentor:
* Adjusted the text to include `@property-read` and `@property-write`.

Examples:
* Expanded the original example to cover all three tags.
* Added the example from PSR 19 as a secondary, more concrete example.

Other:

Ref: https://github.com/php-fig/fig-standards/blob/master/proposed/phpdoc-tags.md#510-property


mvriel:

docs/references/phpdoc/tags/property.rst
```diff
@@ -1,34 +1,49 @@
...
+============================ =====================
+``__get()`` and ``__set()``  ``@property``
+``__get()`` only             ``@property-read``
+``__set()`` only             ``@property-write``
+============================ =====================
 
-@property tags MUST NOT be used in a PHPDoc that is not associated with
-a *class* or *interface*.
+The ``@property``, ``@property-read`` and ``@property-write`` tags can ONLY be used
+in a PHPDoc that is associated with a *class* or *trait*.
```
ashnazg, line 38: Add `interface` to this list?

jrfnl: Eh... interfaces cannot declare properties, so these tags don't apply.

ashnazg: Ah, it's _constants_ that I'm thinking of on interfaces.  Ok, then `*interface*` can be removed from line 43 below... or maybe _replaced_ by `*trait*`.

jrfnl: Good catch, not sure how I missed that one. Fixed now.

ashnazg (approved):

PR #2408, opened 2020-06-14
Docs: improvements for "references - phpdoc - tags - return"

jrfnl:
Summary:
* Added a link to the PHP manual page covering function return values.

Description:
* Minor text synchronization with PSR 19.

Examples:
* Modernized the Type used in the tag.

Other:

Ref: https://github.com/php-fig/fig-standards/blob/master/proposed/phpdoc-tags.md#511-return

**Open question**:

The specs feel outdated. If the function declaration has return type declarations, that should be leading and the tag can be omitted.
The specs currently do not allow for that (both here as well as in PSR19).

### DIFF
--- a/docs/references/phpdoc/tags/return.rst
+++ b/docs/references/phpdoc/tags/return.rst
@@ -1,43 +1,48 @@
 @return
 =======
 
-The @return tag is used to document the return value of functions or methods.
+The ``@return`` tag is used to document the `return value of functions or methods`_.
 
 Syntax
 ------
+
+.. code-block::
 
     @return [Type] [<description>]
 
 Description
 -----------
 
-With the @return tag it is possible to document the return type and function of a
-function or method. When provided it MUST contain a Type to indicate
-what is returned; the description on the other hand is OPTIONAL yet
-RECOMMENDED in case of complicated return structures, such as associative arrays.
+With the ``@return`` tag it is possible to document the return :doc:`Type <../types>`
+of a function or method. When provided it MUST contain a :doc:`Type <../types>`
+to indicate what is returned; the description on the other hand is OPTIONAL yet
+RECOMMENDED, for instance, in case of complicated return structures, such as
+associative arrays.
 
-The @return tag MAY have a multi-line description and does not need explicit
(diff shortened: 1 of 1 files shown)


jaapio:
> The specs feel outdated. If the function declaration has return type declarations, that should be leading and the tag can be omitted.

In phpDocumentor, we treat it the other way around. Docblocks can be more specific than PHP's native return types. I think we should update the sentence so it states that, by default native return type is used, thus the return tag may be omitted. When a more specific return type is needed the tag can be added, and will be leading.

ashnazg:
I've gone back and forth on this thought of the return tag overriding the return in the code, and rewrote my comment here six times already.

Although my general thought is "if the tag says X, go with that no matter what the code says", for both return and param tags, I'm getting less and less convinced in the _use case utility_ of having phpDocumentor do that.  Maybe we do it because it's easy enough to allow, as us being "super flexible to the developer", even if we can personally argue against the developer who chooses to do tags that way when _we_ know that there are better ways _in the code_ to express what the tag override is trying to say (e.g. covariance / contravariance).

Won't PHPStan, Psalm, and even phpStorm argue with the developer that writes the tag with the intention of overriding the return in the code signature?

Maybe the only real use case left for this is PHP code with a code signature return type whose *runtime* is pre-covariance, and thus the code signature _cannot_ yet be changed... yes, I think this is the biggest reason to keep our existing behavior of "tag _blindly_ overrides code".

... and I also agree that the PSRs should highlight this use case to justify allowing such override behavior.

docs/references/phpdoc/tags/return.rst
```diff
@@ -1,43 +1,48 @@
...
     @return [Type] [<description>]
 
 Description
 -----------
 
-With the @return tag it is possible to document the return type and function of a
-function or method. When provided it MUST contain a Type to indicate
-what is returned; the description on the other hand is OPTIONAL yet
-RECOMMENDED in case of complicated return structures, such as associative arrays.
+With the ``@return`` tag it is possible to document the return :doc:`type <../types>`
```
ashnazg, line 16: Capitalize `Type`?

ashnazg (approved):

PR #2410, opened 2020-06-14
Docs: improvements for "references - phpdoc - tags - since"

jrfnl:
Summary:
* Moved the note about support in phpDocumentor to the "Effects in phpDocumentor" section.
* Made a hybrid of the current text in PSR-19 and the original text.

~~Syntax:~~
~~* Replaced "version" with "Semantic Version" as per PSR 19.~~

Description:
* Combined existing text with the current description in PSR-19 with minor adjustments for clarification.

Examples:
* Added a `@version` tag to show the different uses.
* Modernized the Type used in the `@return` tag.

Other:

Ref: https://github.com/php-fig/fig-standards/blob/master/proposed/phpdoc-tags.md#513-since

**Open question**:

I wonder if the note about support in phpDocumentor (now moved to "Effects in phpDocumentor") is still correct.
* IIRC, the `@since` tag seems to be supported in phpDocumentor 3, though some tweaking may still be needed, so is this note still relevant at all ?
* And if it is, it may be more helpful to be more specific about what is and isn't supported, so what can you tell me as input to update this ?

### DIFF
--- a/docs/references/phpdoc/tags/since.rst
+++ b/docs/references/phpdoc/tags/since.rst
@@ -1,39 +1,43 @@
 @since
 ======
 
-.. important::
-
-   The effects of this tag are not yet fully implemented in phpDocumentor 3.
-
-The @since tag indicates at which version did the associated
-Structural Elements became available.
+The ``@since`` tag is used to denote _when_ an element was introduced or modified.
 
 Syntax
 ------
 
-    @since [version] [<description>]
+.. code-block::
+
+    @since [<version>] [<description>]
 
 Description
 -----------
 
-The @since tag can be used to indicate since which version specific
-Structural Elements have become available.
+The ``@since`` tag can be used to document the "version" of the introduction
+or modification of a *Structural Element*.
+
+It is RECOMMENDED that the version matches a semantic version number (x.x.x)
+and it MAY have a description to provide additional information.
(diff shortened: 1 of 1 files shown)


mvriel:

docs/references/phpdoc/tags/since.rst
```diff
@@ -1,39 +1,43 @@
...
-Structural Elements became available.
+The ``@since`` tag is used to denote _when_ an element was introduced or modified.
 
 Syntax
 ------
 
-    @since [version] [<description>]
+.. code-block::
+
+    @since [<Semantic Version>] [<description>]
```
ashnazg, line 11: Given the many cases of SHOULD/MAY flexibility we _have_ to allow for many tags, I'd say this should remain just `[version]`, to account for the many places where a version# is used that doesn't actually qualify as _Semantic_.

This is in line with the use mentioned just below, where _Semantic_ is `RECOMMENDED` rather than _required_.  If we chose to _require_ it, any projects using non-Semantic for their own reasons (many historical) would not be PSR compliant, if only because of this one requirement.

jrfnl: Fixed

ashnazg (approved):

PR #2414, opened 2020-06-14
Docs: improvements for "references - phpdoc - tags - var"

jrfnl:
Summary:
* Synced the text with the current text in PSR-19.

Syntax:
* Removed the `$` from the element_name to allow for it potentially being a constant name.

Description:
* Combined existing text with the current description in PSR-19 with minor adjustments for clarification.

Effects in phpDocumentor:
* Added the section and described the effects. (Hope I got it right)

Examples:
* Added in the global variable examples from PSR-19.
* Synced the existing property example with PSR-19.
* Added an example of a constant with a full docblock and no `@var` description.
* Added in the `foreach` example from PSR-19.
* Minor tweaks to the compound example.
    Note: in this case I have not taken over the example as used in PSR-19 as it conflicts with the text in the description.

Other:

Ref: https://github.com/php-fig/fig-standards/blob/master/proposed/phpdoc-tags.md#517-var

### DIFF
--- a/docs/references/phpdoc/tags/var.rst
+++ b/docs/references/phpdoc/tags/var.rst
@@ -1,46 +1,108 @@
 @var
 ====
 
-You may use the @var tag to document the "Type" of properties, sometimes called class variables.
+You may use the ``@var`` tag to document the :doc:`Type <../types>` of
+the following *Structural Elements*:
+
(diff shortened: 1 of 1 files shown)


mvriel:

ashnazg (approved):
LGTM

PR #2418, opened 2020-06-14
remove ignored error in phpstan fixed in 0.12.29

orklah:


orklah:
I forgot that tool versions are controlled elsewhere :)

Guess you can keep this PR for when the phpstan version will be updated ;)

jaapio:
phpstan is updated :+1:

PR #2472, opened 2020-07-09
Issue #2306 - Option for creating class diagram in the docs

### LINKED ISSUE #2306: [phpDocumentor 3] Bug : no classes.svg for classes graphs
Hello,

I was trying to see the class diagram generated by GraphViz, and it seems the v3 of phpDocumentor don't create the image `classes.svg` which should be in the `graphs` directory of the report. `graphs/classes.html` call this image.

All reports where generated on Linux.

I tried with the docker image phpdoc/phpdoc:3.0 and with the phar file too (installing `apt install graphviz` before), and same for both no `classes.svg` are created.
Running verbose mode (-vvv) don't bring any clues and it even says the file is written! But the file is not physically in the `graphs` folder:
```
21:45:22 NOTICE    [app]   Writer Graph  on graphs/classes.svg
21:45:22 NOTICE    [app]   Writer Graph  on graphs/classes.svg
```
It is not a linux file permission problem as all others files/folders of the reports are well created.

So it seems there is a bug somewhere preventing this image to be created.

**I got two side questions please:**
- where is the link to this `classes.html` file on the generated reports home page or somewhere else ? I found the `graphs` directory by navigating directly on the result reports folder, but i imagine there is somewhere an html link on the report to this diagram.
- graphviz is only used for this classes.html page ?

Thanks.

subhojit777:
Fixes #2306

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ However, code of earlier PHP versions can be analyzed.
 All templates provided with phpDocumentor have support for Class diagrams based on the read code base.
 This will require the application [Graphviz] to be installed on the machine running phpDocumentor.
 Rendering the class diagrams using [Graphviz] is optional, and warnings about missing [Graphviz] can be ignored.
-However, your documentation will contain some dead links in this case.
+However, your documentation will contain some dead links in this case. Class diagram will be created with option `--setting=graphs.enabled=true`.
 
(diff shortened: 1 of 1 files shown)


jaapio:
Thanks for this contribution!

PR #2579, opened 2020-10-03
Update to php 7.4

marvincaspar:


jaapio:
I think it is a bit to early to drop php 7.2 for an application like this one. It doesn't really harm us right now to support php 7.2 - 7.4. So I don't want to bump this right now.

Issue #2217, opened 2020-01-25
Rewrite Mockery calls to Prophecy


mvriel:
Many phpDocumentor tests use Mockery as a mocking/stubbing library but with the advent of prophecy, we are now working to adopt that as our library or method of choice.

To reduce the number of dependencies that we have and simplify our tests, we need all mockery uses to be rewritten to use prophecy.

Any help here is much appreciated. Please issue a PR per test as that will help in reviewing and merging.

simivar:
@jaapio As prophecy integration in PHPUnit is deprecated from version 9 and will be removed in version 10, do you still want to migrate to Prophecy or would it make sense to migrate to PHPUnit's built-in mocking library?

See sebastianbergmann/phpunit#4141

jaapio:
However prophecy is removed from the phpunit core I think it is still superior to the phpunit mocks. I tried recently to improve the phpunit mocks and I was a bit disappointed about the internals. There are much edge cases which are not nicely covered.

So yes I do still prefer to migrate to prophecy.

simivar:
> To reduce the number of dependencies that we have and simplify our tests

As I see in the issue description the point of this task was as well to simplify dependencies of phpDocumentator. If we got with `prophecy` we will just change dependency from `Mockery` to `Prophecy`. Isn't that right?

jaapio:
Since phpdocumentor libraries are used by phpunit, mockery and prophecy it is always a pain to install more dependencies via composer.

We will need to find a way to install prophecy via phar like we do with phpunit right now. Removing mockery is a step forward. Phars can be scoped which prevents issues.

simivar:
I've opened two more PRs and if those get merged what will be left to do is:
- [ ] Migrate phpDocumentor\Descriptor from Mockery to Prophecy (11 classes)
- [ ] Migrate phpDocumentor\Descriptor\Builder\Reflector from Mockery to Prophecy (7 classes)
- [ ] Migrate phpDocumentor\Descriptor\Builder\Reflector\Tags to Prophecy (9 classes)
- [ ] Migrate phpDocumentor\Descriptor\Filter to Prophecy (4 classes)
- [ ] Migrate/refactor phpDocumentor\Faker\Provider class from Mockery

PR #2580, opened 2020-10-03
Migrate test cases that don't use Mockery to PHPUnit's TestCase

simivar:
As the first step with #2217, I have decided to migrate all test classes that don't use any `Mockery` methods from `MockeryTestCase` to the original `TestCase` from PHPUnit. Thanks to that we will easily know which of the classes still needs some work.


jaapio:
Thanks, I had to merge this one manually, because of some conflicts but seems to work as expected.

PR #2582, opened 2020-10-03
Migrate phpDocumentor\Console from Mockery to Prophecy

simivar:
#2217


jaapio:
Thanks for this next step!

PR #2583, opened 2020-10-03
Migrate phpDocumentor\Compiler\Pass from Mockery to Prophecy

simivar:
#2217


jaapio:
And thanks again for this one!

PR #2587, opened 2020-10-04
Migrate phpDocumentor\Transformer from Mockery to Prophecy

simivar:
#2217

`Transformer/Template/CollectionTest.php` was left as child of `MockeryTestCase` instead of PHPUnit's `TestCase` as it uses `phpDocumentor\Faker\Provider` class which is as well dependent on Mockery. 

To keep this PR simple `Provider` **must** be migrated (refactored) later and all child classes must be changed to `TestCase`.


jaapio:
Nice, thanks again :-)

PR #2607, opened 2020-10-05
Migrate phpDocumentor\Descriptor\Builder\Reflector to Prophecy

martzd:
#2217

### DIFF
--- a/tests/unit/phpDocumentor/Descriptor/Builder/Reflector/ClassAssemblerTest.php
+++ b/tests/unit/phpDocumentor/Descriptor/Builder/Reflector/ClassAssemblerTest.php
@@ -13,9 +13,6 @@ declare(strict_types=1);
 
 namespace phpDocumentor\Descriptor\Builder\Reflector;
 
-use Mockery as m;
-use Mockery\Adapter\Phpunit\MockeryTestCase;
-use Mockery\MockInterface;
 use phpDocumentor\Descriptor\ConstantDescriptor;
 use phpDocumentor\Descriptor\MethodDescriptor;
 use phpDocumentor\Descriptor\ProjectDescriptorBuilder;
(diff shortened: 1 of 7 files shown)


tests/unit/phpDocumentor/Descriptor/Builder/Reflector/ClassAssemblerTest.php
```diff
@@ -140,37 +140,35 @@ protected function getClassReflectorDescriptor(?Fqsen $classFqsen = null, ?Fqsen
...
-                        break;
-                }
-
-                return $mock;
+        $projectDescriptorBuilderMock = $this->prophesize(ProjectDescriptorBuilder::class);
+        $projectDescriptorBuilderMock->getDefaultPackage()->shouldBeCalled()->willReturn('\\');
+        $projectDescriptorBuilderMock->buildDescriptor(Argument::any(), Argument::any())->will(function ($param) {
+            switch ($param) {
+                case 'Properties':
+                    $mock = $this->prophesize('phpDocumentor\Descriptor\PropertiesDescriptor');
```
jaapio, line 150: While we are here, Could you please update this to a class ref. `::class`

tests/unit/phpDocumentor/Descriptor/Builder/Reflector/FunctionAssemblerTest.php
```diff
@@ -13,50 +13,39 @@
...
-                }
-            }
-        );
-        $this->argumentAssemblerMock = m::mock(ArgumentAssembler::class);
-        $this->argumentAssemblerMock->shouldReceive('getBuilder')->andReturnNull();
-        $this->argumentAssemblerMock->shouldReceive('setBuilder')->with($this->builderMock);
-        $this->fixture = new FunctionAssembler($this->argumentAssemblerMock);
-        $this->fixture->setBuilder($this->builderMock);
+        $this->builderMock = $this->prophesize(ProjectDescriptorBuilder::class);
+        $this->builderMock->buildDescriptor(ProphecyArgument::any(), ProphecyArgument::any())->shouldBeCalled();
```
jaapio, line 43: This mock has been changed in behavior... however no tests are failing I think it would be good to keep it as it was. Because right now there is a validation missing, which was here because of some failures. 
The expected argument should be of type `DocBlock\Tags\Generic::class` and has a name `package` 

You could do this with `Argument::that(callback)`

jaapio (changes requested):
Thanks for this one. I think we are almost there, this needs some small improvements before we can merge it.

tests/unit/phpDocumentor/Descriptor/Builder/Reflector/ClassAssemblerTest.php
martzd: Such class does not exists anymore so I've deleted the whole switch `case`.

tests/unit/phpDocumentor/Descriptor/Builder/Reflector/FunctionAssemblerTest.php
martzd: Done.

jaapio:
Thanks!

PR #2623, opened 2020-10-12
Psalm.xml simplification

orklah:
This PR propose to simplify the psalm.xml file.

It deletes outdated suppressions and attributes with their default values. It also uses the errorLevel attribute instead of listing each issue


jaapio:
Thanks for the cleanup.

PR #2624, opened 2020-10-13
changing templates for coherence

orklah:
This PR propose changing some templates annotations for psalm.

Most of the time, it shouldn't be needed to add local @var with templates, it means there is something weird going on.

In the case of ProjectDescriptorBuilder.php, it was just an omission in getAssembler documentation.

The AssemblerFactory.php is more complex though. If I understand it fully, The class is a collection of AssemblerMatcher built in advance. When needed, we send a Descriptor, and it returns a matching AssemblerInterface.

That means the Descriptor given and the one retrieved is not actually the same and Psalm detects it. We know it's basically the same because it matched, so I had to cheat a little with an annotation saying that the returned element is the one given in parameter.

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -24,6 +24,9 @@ parameters:
     # PHPStan doesn't support @template D of T but will still complain if we put a D in a Collection of T !
     - '#Parameter \#2 \$value of method phpDocumentor\\Descriptor\\Collection<T>::offsetSet\(\) expects T, D given\.#'
 
+    # phpstan/phpstan/issues/3951
+    - '#Method phpDocumentor\\Descriptor\\ProjectDescriptorBuilder::filterDescriptor\(\) should return TDescriptor of phpDocumentor\\Descriptor\\Descriptor\|null but returns phpDocumentor\\Descriptor\\Filter\\Filterable\|null\.#'
+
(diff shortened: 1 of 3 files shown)


orklah:
I only checked with psalm. Phpstan is failing, I'll look into it

orklah:
The remaining error is a bug in phpstan: https://github.com/phpstan/phpstan/issues/3951

I'll try a workaround if this is not fixed in a few days

phpstan.neon
```diff
@@ -1,5 +1,4 @@
 includes:
-    - /composer/vendor/phpstan/phpstan-webmozart-assert/extension.neon
```
mvriel, line 2: Why did you remove this extension? Is it not needed for the assertion support?

orklah: arg, shouldn't have included that in the commit

orklah: I remove it locally because the path is not valid, but I revert that before committing usually.

I'm a little surprised it didn't seem to cause issue with the build though...

I reverted it

orklah:
I ignored the issue for now, this should be ready

jaapio:
Thanks for cleaning up this part

PR #2628, opened 2020-10-16
use Assert lib instead of native assert

orklah:
this PR propose replacing remaining native assert() function with the Assert class


mvriel:
How do you get phpstan and phpstorm to play nice? I have used assert instead of the WebMozart assert library because of these two :)

orklah:
Psalm has a notion of assertions. It looks like that: `@psalm-assert string $value`. This example means that the param $value of the method is a string, unless the method throws an exception.

Webmozart\assert has a pretty good coverage of those directly in the code so it plays nicely with Psalm.

PHPStan don't have assertions built-in but they have a plugin (that you already use: https://github.com/phpDocumentor/phpDocumentor/blob/master/phpstan.neon#L2)

The rest is just using the right methods in the library to make sure static analysis understand what you're doing.

PR #2629, opened 2020-10-16
Cleanup unused things

orklah:
This PR propose removing some redundant/unused things in the codebase. There's 4 commits detailing each change


mvriel:
@orklah phpstan doesn't like some of these removals :)

orklah:
Yeah, it considers array<mixed> is more detailed than just array!

It's not totally wrong, I'll change some things

jaapio:
Less is better :+1:

PR #2632, opened 2020-10-17
push psalm to level 6

orklah:
This PR fix a few issues and ignore others to upgrade Psalm to level 6. PHPStan will probably be broken, I'll look at that in CI because I think I don't have the same version locally

### DIFF
--- a/src/phpDocumentor/Descriptor/Builder/Reflector/Docblock/DescriptionAssemblerReducer.php
+++ b/src/phpDocumentor/Descriptor/Builder/Reflector/Docblock/DescriptionAssemblerReducer.php
@@ -43,7 +43,7 @@ final class DescriptionAssemblerReducer extends AssemblerAbstract implements Ass
     /**
      * @param Tag[] $tags
      *
-     * @return TagDescriptor[]
+     * @return list<TagDescriptor|null>
      */
     private function createTags(array $tags) : array
     {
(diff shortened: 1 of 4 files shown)


src/phpDocumentor/Descriptor/Builder/Reflector/Docblock/DescriptionAssemblerReducer.php
```diff
@@ -43,7 +43,7 @@ public function create(object $data, ?Descriptor $descriptor = null) : ?Descript
     /**
      * @param Tag[] $tags
      *
-     * @return TagDescriptor[]
+     * @return list<TagDescriptor|null>
```
mvriel, line 46: This is weird .. I would not expect nulls in this result. Perhaps we should resolve that :)

orklah: I'll look into it

orklah: Well, we go through buildDescriptor which has the filter feature inside so it can technically return null. If you say that it can't happen here, we can add an Assert to check that.

jaapio: Null can be used here, the new description implementation requires all array indexes to be there.
Since we do filter tags null can be a result.

mvriel: Hmm, would a Null Object not be a good fit here? I always get a bit uneasy with null values in arrays 😅

jaapio: Yes we could introduce that

PR #2634, opened 2020-10-17
replace static in final class by the class name

orklah:
The static keyword is redundant in final classes, it create unnecessary confusion. I replaced static by the class name.

Aso, would you be open to mark as final every class in the library that is not inherited? You told me once that this library is not supposed to be extended. This would warrant that and allow more simplifications like this PR

### DIFF
--- a/src/phpDocumentor/Descriptor/Builder/Matcher.php
+++ b/src/phpDocumentor/Descriptor/Builder/Matcher.php
@@ -15,15 +15,15 @@ final class Matcher
     private $type;
 
     /**
-     * @param class-string<StaticT> $type
+     * @param class-string<SelfT> $type
      *
-     * @return static<StaticT>
+     * @return self<SelfT>
      *
-     * @template StaticT
+     * @template SelfT
      */
     public static function forType(string $type) : self
     {
-        return new static($type);
+        return new self($type);
     }
 
     /**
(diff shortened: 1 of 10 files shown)


src/phpDocumentor/Descriptor/Builder/Matcher.php
```diff
@@ -15,15 +15,15 @@ final class Matcher
...
-     * @return static<StaticT>
+     * @return self<SelfT>
      *
-     * @template StaticT
+     * @template SelfT
      */
     public static function forType(string $type) : self
     {
-        return new static($type);
+        return new Matcher($type);
```
mvriel, line 26: please use `self` instead; makes for less changes when refactoring and is more expressive than re-using the classname.

mvriel:
> Also, would you be open to mark as final every class in the library that is not inherited? You told me once that this library is not supposed to be extended. This would warrant that and allow more simplifications like this PR

I would be cautious with this; final is our default but various of these classes are not final due to mocking. And some of the classes may be used (but should not)

src/phpDocumentor/Descriptor/Builder/Matcher.php
orklah: done :)

orklah:
> I would be cautious with this; final is our default but various of these classes are not final due to mocking. And some of the classes may be used (but should not)

Oh, didn't thought about mocking. Nevermind then :)

mvriel:
Thank you @orklah!

PR #2637, opened 2020-10-18
Replace null by NullDescriptor when filtering

orklah:
This is a draft of a PR that aims to replace null when filtering Descriptors.

### DIFF
--- a/src/phpDocumentor/Descriptor/Builder/Reflector/Docblock/DescriptionAssemblerReducer.php
+++ b/src/phpDocumentor/Descriptor/Builder/Reflector/Docblock/DescriptionAssemblerReducer.php
@@ -8,6 +8,7 @@ use phpDocumentor\Descriptor\Builder\AssemblerAbstract;
 use phpDocumentor\Descriptor\Builder\AssemblerReducer;
 use phpDocumentor\Descriptor\Descriptor;
 use phpDocumentor\Descriptor\DocBlock\DescriptionDescriptor;
+use phpDocumentor\Descriptor\NullDescriptor;
 use phpDocumentor\Descriptor\TagDescriptor;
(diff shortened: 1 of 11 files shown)


src/phpDocumentor/Descriptor/Builder/Reflector/Docblock/DescriptionAssemblerReducer.php
```diff
@@ -49,7 +50,13 @@ private function createTags(array $tags) : array
     {
         $result = [];
         foreach ($tags as $tag) {
-            $result[] = $this->builder->buildDescriptor($tag, TagDescriptor::class);
+            $tagDescriptor = $this->builder->buildDescriptor($tag, TagDescriptor::class);
+
+            if ($tagDescriptor instanceof NullDescriptor) {
```
jaapio, line 55: we should not apply the filtering here, the descriptionDescriptor expects all tags to be in. NullDescriptor should just result into a empty replacement. 
So I think we would need a specialized NullTagDescriptor to make this possible

orklah: Can do, but the templates will be ugly.

That means the return of Filter::filter will look like `@return TDescriptor|NullDescriptor|NullTagDescriptor`. That implies we will have to handle both NullDescriptor and NullTagDescriptor in every call to comply with static analysis.

Unless we use conditional returns like `@return ($descriptor is TagDescriptor ? NullTagDescriptor : NullDescriptor)`. This will be perfect for static analysis, but I understood you weren't big fans of this syntax...

orklah: You have an opinion on this @jaapio ?

PR #2650, opened 2020-10-24
try to run on php 8

orklah:
This is a draft PR to try to run CI on PHP8.

This will allow to see what's failing and take actions if necessary


jaapio:
However this pr is not perfect yet I want to merge this anyway. So we can improve it while we are improving phpDocumentor.

PR #2657, opened 2020-10-30
fix static analysis build

orklah:
This should fix the build for both Psalm an PHPStan. I had to refactor the use of template. I believe I went overboard in a previous commit and some of it was not useful and too complex to maintain.


orklah:
The build still fails but the static analysis part is OK, so I think this can be merged

mvriel:
Thanks @orklah; it seems Phive is acting up somehow. I think I heard @jaapio about it this morning. So I am indeed merging this as it will resolve the psalm/phpstan issues

PR #2673, opened 2020-11-08
Lvl up Psalm to 5

orklah:
This PR propose a serie of fixes to upgrade Psalm to lvl 5.

The build will fail because of 4 last errors which should be fixed by https://github.com/phpDocumentor/Reflection/pull/181, however, this would need a new release of Reflection.


orklah:
This should be ready now

jaapio:
Nice :-D Level 5!

PR #2674, opened 2020-11-08
Psalm improvements

orklah:
This PR fix a bunch of issues detected by Psalm. Mainly documentation.

After that PR, there should be only a few issues left to upgrade to level 4.

This PR also tries to update Psalm to the new major release

### DIFF
--- a/phive.xml
+++ b/phive.xml
@@ -4,5 +4,5 @@
   <phar name="composer-require-checker" version="^2.0.0" installed="2.1.0" location="./tools/composer-require-checker" copy="true"/>
   <phar name="php-coveralls/php-coveralls" version="^2.2.0" installed="2.2.0" location="./tools/php-coveralls" copy="true"/>
   <phar name="scrutinizer-ci/ocular" version="^1.6.0" installed="1.6.0" location="./tools/ocular" copy="true" force-accept-unsigned="true"/>
-  <phar name="psalm" version="^3.9.3" installed="3.17.1" location="./tools/psalm" copy="true"/>
+  <phar name="psalm" version="^4.1" installed="^4.1" location="./tools/psalm" copy="true"/>
 </phive>
--- a/src/Guides/Markdown/Parser.php
+++ b/src/Guides/Markdown/Parser.php
@@ -137,7 +137,7 @@ final class Parser implements ParserInterface
             }
 
             echo 'DOCUMENT CONTEXT: I am '
-                . ($event->isEntering() ? 'entering' : 'leaving')
+                . 'leaving'
                 . ' a '
                 . get_class($node)
                 . ' node'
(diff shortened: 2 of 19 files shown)


orklah:
If the review is OK, this can be merged as-is. I'll now work on Configuration to see if I can improve the documentation and my comprehension before considering a refactor

src/phpDocumentor/Configuration/Definition/Version2.php
```diff
@@ -152,7 +152,12 @@ public function getConfigTreeBuilder() : TreeBuilder
      *
      * @param array<string, string|array<string, mixed>> $values
      *
-     * @return array<string, array<int|string, mixed>|string>
+     * @return array{
+     * configVersion: string, title: array<string, mixed>|string, paths: array{output: mixed, cache: mixed},
```
jaapio, line 156: I'm wondering why `title` could be an array. I think this should only be a `string` 
`output` and `cache` should be `string`

src/phpDocumentor/Configuration/Definition/Version2.php
```diff
@@ -152,7 +152,12 @@ public function getConfigTreeBuilder() : TreeBuilder
...
+     * version: array{0: array{number: string, api: array{0: array{default-package-name: mixed,
```
jaapio, line 157: `default-package-name` should be `string|null` 

I do not think we should define `0` as a key here, it is just an array with `int` index type. Is there a specific reason to define it this way? Maybe how we are using it right now in the application?

src/phpDocumentor/Configuration/Definition/Version2.php
```diff
@@ -152,7 +152,12 @@ public function getConfigTreeBuilder() : TreeBuilder
...
+     * source: array{paths: array<string>}, ignore: array{paths: array<string>}, extensions: array{extensions: mixed},
+     * visibilities: non-empty-list<string>|null, markers: array{markers: mixed}}}}}, templates: array
```
jaapio, line 159: `markers: mixed` should be `markers: non-empty-list<string>` 
`templates: non-empty-list<string>`

jaapio (commented):
In general this like perfect. 
Just some expected changes in the config definition. But those are based on what I would expect from the knowledge I have about the config. It could be that we need to change the code to make these statements true :-)

src/phpDocumentor/Configuration/Definition/Version2.php
orklah: This was Psalm inferring what it can based on the code. I did not look too deep in the array because my goal was to have the first few levels right. I'll revert this part anyway, I already rewrote that in another branch

orklah:
I reverted the big block. Mind merging this so I can push another PR with the documentation of Config?

jaapio:
Done :+1:

PR #2685, opened 2020-11-10
improve configuration doc

orklah:
I think this is pretty much done.

However, I'm gonna get pinned by CS rules for lines that are way too long! I'm not sure what should be done about that. Should I indent them on multiple lines or should it stay like this for now while refactoring...

There's probably some mistakes in here, maybe some keys that can't yet be defined were I put them or that are optional and I flagged them mandatory but I think this is mostly correct.

It draws a big picture and it doesn't look good! I think the refactoring should begin by a Version object and we'll see how that goes from here

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -32,6 +32,11 @@ parameters:
 
     # PHPStan doesn't support inheritance of TDescriptor
     - '#Parameter \#2 \$assembler of method phpDocumentor\\Descriptor\\Builder\\AssemblerFactory::(register|registerFallback)\(\) expects phpDocumentor\\Descriptor\\Builder\\AssemblerInterface.* given\.#'
+
+    # Advanced transformations of array. PHPStan doesn't seem to keep nonEmpty flag through some array functions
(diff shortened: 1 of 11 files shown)


voku:
😲 This arrays looks very "special". Thanks for improving this step by step.

orklah:
> 😲 This arrays looks very "special". Thanks for improving this step by step.

Thanks! It's pretty complex yeah, There are multiple sources and arrays are transformed from one version of the config to the next so array keys are not always there.

Let's hope I can make it clearer at the end :)

jaapio:
Thanks @orklah, I think it is ok to put these lines on the ignore list for phpcs. Making this multi line will not make it look better.

What you are facing here is basically all configuration phpdocumentor is able to use. So yeah, it is complex. The migration from v2 to v3 doesn't make it better. Introducing some value object here would make a lot of sense.

orklah:
This shouls be ready 👍 

I'm also ready to push another PR with the first Value Object when this is merged

jaapio:
Thanks a lot for the time you are spending on this. Looking forward to the next pr :-D

PR #2687, opened 2020-11-13
introduce VersionObject

orklah:
This PR propose introducing a VersionObject to replace a big Version array in configuration.

I wasn't very inspired for the name so it can be changed if needed (just not in "Version" to avoid confusion with Version2 and Version3)

### DIFF
--- a/src/phpDocumentor/Configuration/CommandlineOptionsMiddleware.php
+++ b/src/phpDocumentor/Configuration/CommandlineOptionsMiddleware.php
@@ -63,12 +63,12 @@ final class CommandlineOptionsMiddleware implements MiddlewareInterface
         $configuration = $this->overwriteSettings($configuration);
 
         if (!isset($configuration['phpdocumentor']['versions'])) {
-            $configuration['phpdocumentor']['versions'][] = $this->createDefaultVersionSettings();
+            $configuration['phpdocumentor']['versions']['1.0.0'] = $this->createDefaultVersionSettings();
         }
 
         if ($this->shouldReduceNumberOfVersionsToOne($configuration)) {
             $configuration['phpdocumentor']['versions'] = [
-                end($configuration['phpdocumentor']['versions']),
+                '1.0.0' => end($configuration['phpdocumentor']['versions']),
             ];
         }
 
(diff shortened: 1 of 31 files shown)


orklah:
I'll check CI failures this afternoon

orklah:
This should be ready for review.

Please note that the second commit fix what I'd consider a bug. The index of the versions array was used incrementally where it's supposed to be a string. I replaced by 1.0.0

orklah:
This should be ready. I overlooked a lot of things in the previous PR. This should be much better now

src/phpDocumentor/Configuration/ValueObjects/VersionObject.php
```diff
@@ -0,0 +1,72 @@
...
+ *
+ * @link https://phpdoc.org
+ */
+
+namespace phpDocumentor\Configuration\ValueObjects;
+
+use phpDocumentor\Dsn;
+use phpDocumentor\Path;
+
+class VersionObject
```
jaapio, line 19: I think this class should be named `phpDocumentor\Configuration\VersionSpecification`  There is no need to have a separate `ValueObjects` namespace for this.

src/phpDocumentor/Configuration/ValueObjects/VersionObject.php
```diff
@@ -0,0 +1,72 @@
...
+{
+    /** @var string */
+    public $number;
```
jaapio, line 22: I prefer getters for all properties in this class. In the end, this object should be immutable. The only location where mutations are allowed is in the `Configuration` namespace.

jaapio (changes requested):

src/phpDocumentor/Configuration/ValueObjects/VersionObject.php
orklah: Getters were added, but I had difficulties for setters because subsets of the api property are overwritten dozens times which make the process really heavy. You have to make a copy of the whole array, changing what need to be changed locally and then push the changes.

It may be easier once we have more Specification objects because we'll be able to write a subset via a sub-object.

src/phpDocumentor/Configuration/ValueObjects/VersionObject.php
orklah: Done :)

jaapio:
Thanks a lot for this improvement.

PR #2688, opened 2020-11-13
remove Font Awesome all.min.js and leave only all.min.css

### LINKED ISSUE #2686: Font Awesome is loaded twice in the HTML template, with conflicting technologies
## Expected behavior
Font Awesome should be loaded at most one time, and it should be either Webfont technology or SVG technology, not both.

Ironically, I was publishing the latest version the Font Awesome WordPress plugin when I noticed this problem in its own generated API docs 😄.

## Actual behavior
Both are being loaded.

See [here in layout.html.twig](https://github.com/phpDocumentor/phpDocumentor/blob/8332a760a3e8c280aa1996194eadb9c5c4d840d2/data/templates/default/layout.html.twig#L14) and [here](https://github.com/phpDocumentor/phpDocumentor/blob/8332a760a3e8c280aa1996194eadb9c5c4d840d2/data/templates/default/layout.html.twig#L19).

It's self-evident from reading the code.

## Your environment

N/A

* Version used: the problem is currently on the master branch

---
I'd gladly put up a PR to fix this, but I'm not sure which to remove since I don't know what is the desired behavior here. The choice of Font Awesome technology depends upon the use cases. But it's never correct to load it twice. Probably the simplest approach, unless you know that you require SVG/JS-only features, like [Power Transforms](https://fontawesome.com/how-to-use/on-the-web/styling/power-transforms), is to drop the `all.min.js` and just keep the Webfont/CSS technology via `all.min.css`.

mlwilkerson:
Closes #2686 

- [x] remove all.min.js
- [x] audit uses of Font Awesome throughout to ensure there are no SVG/JS-specific uses
- [x] determine whether the presence of the "svg-inline--fa" classes in [`index.html.dist`](https://github.com/mlwilkerson/phpDocumentor/blob/2563c52743d553a8b814e2e73c54a3bc4849e827/data/demo/index.html.dist#L7) are accidental or essential

### DIFF
--- a/data/templates/default/layout.html.twig
+++ b/data/templates/default/layout.html.twig
@@ -16,7 +16,6 @@
     {% block javascripts %}
         <script src="https://cdn.jsdelivr.net/npm/fuse.js@3.4.6"></script>
         <script src="https://cdn.jsdelivr.net/npm/css-vars-ponyfill@2"></script>
-        <script src="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/5.12.0/js/all.min.js" integrity="sha256-0vuk8LXoyrmCjp1f0O300qo1M75ZQyhH9X3J6d+scmk=" crossorigin="anonymous"></script>
         <script src="js/search.js"></script>
         <script defer src="js/searchIndex.js"></script>
     {% endblock %}


mlwilkerson:
@jaapio after doing some quick, basic searching for uses of `fa-` classes in the HTML templates in this repo, I only find basic usage that is perfectly compatible with the Font Awesome Webfont technology. So if those were the only things to see, I'd say this PR is ready to go.

However, the `data/demo/index.html.dist` file [includes a style element](https://github.com/mlwilkerson/phpDocumentor/blob/2563c52743d553a8b814e2e73c54a3bc4849e827/data/demo/index.html.dist#L7) that is normally injected by the Font Awesome SVG/JS framework, but is not used with the Webfont technology.

This file looks like it's part of a demo app, and if I had to guess, I'd say that it is the result of taking a snapshot of DOM elements after they'd been rendered and subsequently transformed by the Font Awesome SVG/JS auto-replacement process. Maybe they were copied from a browser after that rendering process, or possibly from some server-side rendering process. I'd guess that in part because this `<style>` element is normally inserted into the DOM automatically when the SVG/JS framework loads and kicks off its process of detecting and auto-replacing icon elements like `<i>` tags with `fa-classes` with the corresponding `<svg>` elements. The `svg-inline--fa` CSS class is added to those resulting `<svg>` elements to ensure that they are correctly sized.

While the `svg-inline--fa` class does appear on some `<svg>` elements in that file, it appears to me that this is only the case because this file represents a snapshot of the HTML _after_ the SVG/JS framework has done its SVG auto-replace work because that `all.min.js` was loaded, albeit unnecessarily so.

The way the Webfont tech works is that it just directly styles the `<i>` tags. It doesn't _replace_ `<i>` tags with `<svg>` tags.

So if my hunch is right that `index.html.dist` depends on the JS/SVG framework only because this snapshot was taken after a loaded instance of the JS/SVG did its magic on it, then it should work fine, after removing `all.min.js`, to rebuild/regenerate this `index.html.dist`. I expect the resulting template to no longer have that extra `<style>` element in it. And I would expect that all `<i>` elements with classes like `<i class="fab fa-github"></i>` would remain as `<i>` tags in the final HTML. Then, when loaded in the browser with only `all.min.css`, I expect that they'll look just like you intend them to.

But I'm not sure how to rebuild that `index.html.dist` in the demo app in order to verify all of this. Can you help with that part?

jaapio:
Hi,

Thanks for investigating this. The demo index is not used by phpdocumentor itself. It is the front page which you see at demo.phpdoc.org. The page is not regenerated by phpdocumentor.

I think this pr is good to go.
Thanks again for taking the time to create a pr. With contributions like this we can move forward to a better phpdocumentor.

mlwilkerson:
OK, I'll consider it done, then.

jaapio:
Thanks!

PR #2689, opened 2020-11-13
Try to fix phive

orklah:
It seems a previous change I made in phive for psalm broke something in the build 
https://github.com/phpDocumentor/phpDocumentor/runs/1394684226#step:9:14

This is a test to see if I can fix it.


orklah:
Ok, this is fixed. There is still an error in phive, but it's not my fault anymore :D

PR #2712, opened 2020-11-30
Fix typo in document

takaram:
Current version of the reference says that `callback` is a valid type in PHPDoc, which should be `callable`.
This PR fixes this issue.


jaapio (approved):

jaapio:
Thanks!

PR #2752, opened 2021-01-05
Typo in the number of ways to install

guinru:
There are 4 ways listed, not 3.


jaapio:
Thanks for your PR. But, your change is not correct. There are 4 ways listed, but only 3 are supported.

We do not support composer. As you can read on our website: https://phpdoc.org

PR #2775, opened 2021-01-18
Introduce api object to refactor filters

jaapio:


src/phpDocumentor/Configuration/ApiSpecification.php
```diff
@@ -226,7 +226,7 @@ public function getIgnoredTags() : array
         return $this->ignoreTags;
     }
 
-    private function calculateVisablity() : int
+    public function calculateVisablity() : int
```
orklah, line 229: Maybe an opportunity to fix the type on Visibility?

PR #2779, opened 2021-01-29
Leverage newer API of PrettyVersion

Jean85:
With those changes, you'll stop leveraging the transient dependency of `ocramius/package-version` and use the dedicated API of PrettyVersion. This also allows PrettyVersion 2.0 which drops the sub dependency alltogether, and hooks into the Composer 2 APIs directly.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -34,7 +34,7 @@
         "ext-xml": "*",
         "doctrine/event-manager": "^1.1",
         "jawira/plantuml": "^1.27",
-        "jean85/pretty-package-versions": "^1.2",
+        "jean85/pretty-package-versions": "^1.5 || ^2.0.1",
         "league/commonmark": "^1.5",
         "league/flysystem": "^1.0",
         "league/pipeline": "^1.0",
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -34,7 +34,7 @@
         "ext-xml": "*",
         "doctrine/event-manager": "^1.1",
         "jawira/plantuml": "^1.27",
-        "jean85/pretty-package-versions": "^1.2",
+        "jean85/pretty-package-versions": "^1.5 || ^2.0.1",
```
jaapio, line 37: Is there any specific reason why we would need to support both versions? If possible we prefer to have just one version :-D

jaapio (commented):
Thanks for your PR, it is really appreciated that you took the time to help us improve phpDocumentor! 

It would be nice if you could answer my question about the supported versions. From there I can take this PR over and fix the missing parts regarding our code style :-D Unless you have time to do it yourself.

composer.json
Jean85: Apart from the common problem of installing multiple projects with conflicting constraints, 2.0 requires Composer 2, that's the only difference. If you're fine with imposing that to your end users, you can drop 2.0.

jaapio: Thanks, I will take a look :+1:

PR #2827, opened 2021-02-19
Uml directive 

jaapio:
Introduces the new UML directive that will use plantuml to render diagrams in your documentation

image


mvriel:
Is this something I can help with, while you focus on the Version/DocumentationSet refactor?

PR #2833, opened 2021-02-24
Admonition contents are not rendered

mvriel:
In this change, I have made a start with rendering admonitions. The
prior method did not include the body of the admonition and as such
didn't really show anything.

This can still be improved, both the docs and the rendering style; and I
will do so in a subsequent commit

### DIFF
--- a/src/Guides/RestructuredText/HTML/Directives/AbstractAdmonitionDirective.php
+++ b/src/Guides/RestructuredText/HTML/Directives/AbstractAdmonitionDirective.php
@@ -39,29 +39,23 @@ abstract class AbstractAdmonitionDirective extends SubDirective
         array $options
     ) : ?Node {
         $environment = $document->getEnvironment();
-        $wrapperDiv = function () use ($environment, $options) {
-            return $environment->getRenderer()->render(
-                'directives/admonition.html.twig',
-                [
-                    'name' => $this->name,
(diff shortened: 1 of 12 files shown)


src/Guides/RestructuredText/HTML/Directives/AbstractAdmonitionDirective.php
```diff
@@ -39,18 +39,23 @@ final public function processSub(
...
+                    return $environment->getRenderer()->render(
+                        'directives/admonition.html.twig',
+                        [
+                            'name' => $this->name,
+                            'text' => $this->text,
+                            'class' => $options['class'] ?? null,
+                        ]
+                    );
+                },
+                '</article></div>'
```
jaapio, line 57: I guess the reason why this is here is because directives can be nested?

jaapio (approved):

src/Guides/RestructuredText/HTML/Directives/AbstractAdmonitionDirective.php
mvriel: Unfortunately yes .. I would have loved a different approach but the Guides library does nesting with callables and promises; which make things a bit oddish. And I do not have enough knowledge on the internals yet to make a solution that matches my expectations better

PR #2836, opened 2021-02-27
Guides: Render code blocks

mvriel:
In this change, I have made the necessary changes to make sure code
blocks are rendered correctly. The existing code was a weird hybrid
between Scrivo/Highlight.php and, I think, a Javascript library.

I have moved entirely to using prism as a javascript library as that is
a well-maintained, good working, library that includes features that are
very practical to have. This includes line numbering, which is not
supported through another route.

In this change, I have introduced support for the `code` and
`code-block` directives; with their respective options. With the `linenos`
option, you enable line numbering and with lineno-start you can set a
starting linenumber (the latter will automatically enable the former).
There is an alternate option called number-lines, that acts similar to
lineno-start as it is an integer that will enable line numbering _and_
set a starting number.

See
https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-code-block
and https://docutils.sourceforge.io/docs/ref/rst/directives.html#code.

I do not know why there are two ways .. but nonetheless. We support
both.


mvriel:
As a side-note: I am introducing several extra CDN roundtrips in this PR. I am researching how we could setup webpack and build a locally hosted version. This is not intended to be done by users, but as a development tool for us to generate a minified optimal JS file

jaapio (approved):

PR #2837, opened 2021-02-28
Project builder improvements

jaapio:


mvriel:
@jaapio what is the status of this PR; I am fearful that more incoming changes for the RST parser might influence this PR and it contains a lot of changes

jaapio:
Please don't care about this PR, I will update it once in a while. Some changes in here are already merged, it was more like a work in progress branch to see where the problems are to start supporting multiple versions.

One of the things that came out was the file handling which has been changed. The table of contents changes also found their origin in here. When I have time to pick this one up, I will have to review the changes anyway.

PR #2840, opened 2021-03-02
Guides: Copy asset when using Image directive

mvriel:
In this change, I have introduced an `asset` twig function that will
copy the provided asset and return a canonical path to that asset. This
will, at the moment, only work for guides because it makes an assumption
that the BASE tag is used in the HEAD of the output.

I have seen a few spots in the Renderer and the Environment classes that
definitely need improvement and rework; but keep in mind that these are
used anywhere and everywhere and as such I cannot do big refactorings
and deliver these directives at the same time :)


jaapio (approved):
Apart from the failing ci, this looks good to me :+1:

PR #2842, opened 2021-03-02
Split transformer from template creation

jaapio:
To be able to add more transformers in an upcoming pr which will
allow us to support multiple versions, documentation sets etc. I removed
all template-related code from the transformer.

The Template Factory will now provide a collection of templates to render
equal to the existing template collection this is able to provide the required
transformations. The Transformer will only instruct the writers to execute transformations.

This commit contains a number of shortcuts to finish this faster. The configuration type
hints could be improved, and the destination should be removed from the transformer. Although
both are not really required at the moment. Things will work as before now.

### DIFF
--- a/src/phpDocumentor/Pipeline/Stage/Transform.php
+++ b/src/phpDocumentor/Pipeline/Stage/Transform.php
@@ -5,18 +5,19 @@ declare(strict_types=1);
 namespace phpDocumentor\Pipeline\Stage;
 
 use Exception;
+use League\Flysystem\FilesystemInterface;
 use phpDocumentor\Dsn;
 use phpDocumentor\Event\Dispatcher;
-use phpDocumentor\Reflection\DocBlock\ExampleFinder;
+use phpDocumentor\Parser\FlySystemFactory;
+use phpDocumentor\Transformer\Event\PostTransformEvent;
 use phpDocumentor\Transformer\Event\PreTransformationEvent;
 use phpDocumentor\Transformer\Event\PreTransformEvent;
 use phpDocumentor\Transformer\Event\WriterInitializationEvent;
+use phpDocumentor\Transformer\Template\Factory;
 use phpDocumentor\Transformer\Transformer;
 use phpDocumentor\Transformer\Writer\WriterAbstract;
 use Psr\Log\LoggerInterface;
 use Symfony\Component\Filesystem\Filesystem;
-use Symfony\Component\Stopwatch\Stopwatch;
-use function array_column;
 use function count;
 use function get_class;
 use function getcwd;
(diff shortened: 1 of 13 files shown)


src/phpDocumentor/Pipeline/Stage/Transform.php
```diff
@@ -42,20 +43,25 @@ class Transform
     /** @var LoggerInterface */
     private $logger;
 
-    /** @var ExampleFinder */
-    private $exampleFinder;
+    /** @var Factory */
+    private $factory;
```
mvriel, line 47: Can you make the property name more descriptive?

src/phpDocumentor/Pipeline/Stage/Transform.php
```diff
@@ -42,20 +43,25 @@ class Transform
...
 
     /**
      * Initializes the command with all necessary dependencies to construct human-suitable output from the AST.
      */
     public function __construct(
         Transformer $transformer,
+        FlySystemFactory $flySystemFactory,
         LoggerInterface $logger,
-        ExampleFinder $exampleFinder
+        Factory $factory
```
mvriel, line 59: Can you make the argument name more descriptive?

src/phpDocumentor/Pipeline/Stage/Transform.php
```diff
@@ -114,35 +139,20 @@ function (PreTransformationEvent $event) : void {
...
     {
         $target     = $dsn->getPath();
         $fileSystem = new Filesystem();
         if (!$fileSystem->isAbsolutePath((string) $target)) {
             $target = getcwd() . DIRECTORY_SEPARATOR . $target;
         }
 
+        $destination = $this->flySystemFactory->create(Dsn::createFromString((string) $target));
+
+        //TODO: the guides to need this, can we get rid of these lines?
```
mvriel, line 152: Perhaps if we move the creation of the flysystem to the mapping of the config to a spec; and pass that spec onto the guide renderer. But at the moment the transformer is that which binds the writers to the config

src/phpDocumentor/Transformer/Event/PostTransformEvent.php
```diff
@@ -53,4 +57,16 @@ public function setProject(ProjectDescriptor $project) : self
 
         return $this;
     }
+
+    /** @param Transformation[] $transformations */
+    public function setTransformations(array $transformations) : void
+    {
+        $this->tranformations = $transformations;
```
mvriel, line 64: Should be `transformations`

src/phpDocumentor/Transformer/Event/PreTransformEvent.php
```diff
@@ -24,6 +25,9 @@ final class PreTransformEvent extends EventAbstract
     /** @var ProjectDescriptor|null */
     private $project;
 
+    /** @var Transformation[] */
+    private $tranformations;
```
mvriel, line 29: should be `transformations`

src/phpDocumentor/Transformer/Template/Factory.php
```diff
@@ -173,6 +205,20 @@ private function resolve(Transformer $transformer, string $nameOrPath) : Filesys
         throw new TemplateNotFound($nameOrPath);
     }
 
+    private function getTemplatesDirectory() : Filesystem
+    {
+        $dsnString = $this->getTemplatesPath();
+        try {
+            $filesystem = $this->flySystemFactory->create(Dsn::createFromString($dsnString));
```
mvriel, line 212: Not for now, but it might be better to instantiate the filesystems in the config loading phase and inject them here. This will remove the need for passing DSNs around

mvriel (approved):

PR #2843, opened 2021-03-05
Guides: Refactor renderers into Flyweights

mvriel:
As part of a larger effort to make the "Nodes" of the Guides library
into true value objects and transform these into Descriptors for
phpDocumentor, I need to refactor the services out of these classes.

The NodeRenderer is currently injected into the Nodes and the first step
to refactor this out of these nodes is to make sure that the rendering
of the nodes is moved out of there, and that is only possible if the
renderers are not co-dependent on the nodes.

To remove this co-dependence, I am now transforming the NodeRenderers
into a series of flyweights so they can function independent and can be
created in the dependency injection container.

TODO:

- [x] Refactor the callable node renderer factory away so that they really are flyweights
- [x] Review the renderers and check their locations and functioning if further refactoring is needed


jaapio:
Looks good to me :+1:

PR #2892, opened 2021-05-12
Adjust the nightly releases documentation in README

NicolasCARPi:
* use plural form for the nightly releases
* remove non-existant phar link
* update the location of the downloadable phar archive


jaapio:
Sweet, thanks

PR #2894, opened 2021-05-13
Fix constant value desplaying as empty string when zero

### LINKED ISSUE #2893: Class constant value is displaying as an empty string when the value is zero
I have three constants with values `0`,`1`,`2`. The last constants is appearing without any issue. But the first constant is displaying as an empty string.

## Expected behavior
constant signature:-
```
 public mixed ANTENNA_LOCATION_CODE_OUTDOOR = 0
```

## Actual behavior
constant signature:-
```
 public mixed ANTENNA_LOCATION_CODE_OUTDOOR = ""
```

My class:-

```php
class Antenna {
     public const LOCATION_OUTDOOR=0;
     public const LOCATION_INDOOR=1;
     public const LOCATION_FREQUENTLY_CHANGING=3;
}
```

whizsid:
Fixes #2893 .


jaapio:
Thanks for this fix!

Issue #2937, opened 2021-06-23
Encoding error in reStructuredText multiline table cells


daiwai:
If you create a table cell with reStructuredText in the guide that goes over more than one line and in the second or one of the following lines a non-ASCII Unicode character is used, such as the German umlaut "ö". This character will be encoded incorrectly during rendering.

## Expected behavior

non-ASCII UTF-8 encoded characters should be rendered correctly in reST tables

## Actual behavior

non-ASCII UTF-8 encoed characters, that appear in the second or any subsequent line of a multi line table cell are not encoded correctly during rendering.

## Steps to reproduce the problem

Here is a fairly minimal example of the reST markup that triggers the problem

    +---+---+
    | A | B |
    |   | ö |
    +---+---+
This is rendered as

<table>
    <tbody>
        <tr>
            <td>A</td>
            <td>B Ã¶</td>
        </tr>
    </tbody>
</table>

    <table>
        <tbody>
            <tr>
                <td>A</td>
                <td>B Ã¶</td>
            </tr>
        </tbody>
    </table>

instead of

<table>
    <tbody>
        <tr>
            <td>A</td>
            <td>B ö</td>
        </tr>
    </tbody>
</table>

The following examples, on the other hand, _do not_ have that problem and are both rendered correctly:

    +---+---+
    | A | ö |
    +---+---+

    +---+---+
    | A |   |
    |   | ö |
    +---+---+

## Your environment

* Version used: a570609
* Install method: phar
* PHP version: 8.0.7
* Operating system and version: Ubuntu 20.04.2 LTS

daiwai:
This problem is caused by the fact that the `phpDocumentor\Guides\Nodes\Table\TableColumn::addContent` method encodes the passed string with `utf8_encode`.

Since the content of `TableColumn` is already encoded when it is instantiated and all calls to TableColumn::addContent pass the return value of TableColumn::getContent, the content is encoded twice.

So the quickest fix would be to simply remove the call to utf8_encode in `TableColumn::addContent`.

[...]

PR #2938, opened 2021-06-24
Fully support UTF-8 in reSructuredText tables

daiwai:
Removes the intermediate conversion from UTF-8 to ISO-8859-1 and vice
versa when compiling tables.

This fixes the double encoding that caused #2937 and similar issues. It
also adds support for characters, that are not covered by ISO-8859-1 and
would lead to data loss when compiling tables containing such
characters.


jaapio:
Thanks for reporting this issue and providing a fix!

PR #2959, opened 2021-08-01
[Guides] Span parser&lexer

jaapio:
This patch provides a new lexer to handle the span contents. It allows us a more flexible approach to support all features of RST links. This step is just a first step in order to support all link features. Like described in https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#hyperlink-references

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -36,6 +36,7 @@
         "ext-simplexml": "*",
         "ext-xml": "*",
         "doctrine/event-manager": "^1.1",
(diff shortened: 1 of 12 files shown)


stof:
@jaapio AFAICT, your RST-parser has been created by vendoring https://packagist.org/packages/doctrine/rst-parser at some point. Why not using the original library, in which *lots* of issues have been fixed in the past few months (thanks to symfony.com switching to using it) ?

PR #2972, opened 2021-08-04
Tests: fix an incorrect namespace

jrfnl:
Fixes the following issue when running `composer install`:
```
Class phpDocumentor\Guides\Twig\AssetsExtensionTest located in ./tests/unit/Guides/RestructuredText/Twig/AssetsExtensionTest.php does not comply with psr-4 autoloading standard. Skipping.
```


jaapio (approved):

jaapio:
Thanks!

PR #2976, opened 2021-08-06
PHP 8.1: fix various deprecation warnings

jrfnl:
A run on PHP 8.1 currently shows various notices along the lines of:
```
Deprecated: Return type of [CLASSNAME]::[METHODNAME]() should either be compatible with [INTERFACENAME]::[METHODNAME](): [Type], or the #[ReturnTypeWillChange] attribute should be used to temporarily suppress the notice
```

These deprecation notices relate to the [Return types for internal methods RFC](https://wiki.php.net/rfc/internal_method_return_types) in PHP 8.1.

Basically, as of PHP 8.1, these methods in classes which implement PHP native interfaces are expected to have a return type declared.
The return type can be the same as used in PHP itself or a more specific type. This complies with the Liskov principle of covariance, which allows the return type of a child overloaded method to be more specific than that of the parent.

While this package has a minimum PHP requirement of PHP 7.2 based on the `composer.json` file, several of these return types expect for the PHP 8.0 `mixed` type to be used, so for consistency across the codebase, I've used the attribute for now to silence the warning.

While attributes are a PHP 8.0 feature only, due to the syntax choice for `#[]`, they will ignored in PHP < 8 and can be safely added.

### DIFF
--- a/src/phpDocumentor/Configuration/LegacyArrayAccess.php
+++ b/src/phpDocumentor/Configuration/LegacyArrayAccess.php
@@ -14,6 +14,7 @@ declare(strict_types=1);
 namespace phpDocumentor\Configuration;
 
 use InvalidArgumentException;
+use ReturnTypeWillChange;
 
 use function lcfirst;
 use function property_exists;
@@ -23,6 +24,7 @@ use function ucwords;
 trait LegacyArrayAccess
 {
     /** @param string $offset */
+    #[ReturnTypeWillChange]
     public function offsetExists($offset): bool
     {
         $property = $this->normalizePropertyName($offset);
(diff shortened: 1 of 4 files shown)


jrfnl:
Note: the CS workflow will fail now due to various sniffs in PHPCS not handling attributes correctly yet. This has already been fixed upstream and the fixes will be included in the next PHPCS release (3.6.1).

Though if I'm honest, your coding standards repo can use some updates anyway as it looks like it is very much out of date.

jaapio:
@jrfnl I think this one can be merged after a rebase?

jrfnl:
Rebased, though I expect the PHPCS workflow to still fail - see my previous remark about this.

jaapio:
Well, since you fixed our code style in earlier pr's I think this is good to go :-)

jrfnl:
Looks like some other build runs failed though - want to restart those ?

jaapio:
Mac os on github actions is unstable. :-(  I didn't have time yet to figure that out.

PR #2977, opened 2021-08-08
CS update after upstream changes

jrfnl:
###  PHPCS ruleset: update ruleset for upstream changes

* Set the minimum PHP version for the PHPCompatibility standard.
* Prevent false positives for PHP features polyfilled via the Symfony PHP polyfills.
* Don't require property type declarations.
* Don't require "modern class name references".
* Remove reference to sniff which no longer exists.

### CS: docblock tag grouping

Apparently Doctrine wants the related `@psalm` and `@phpstan` annotations to be grouped with the `@param` or `@return` tags.

### CS: miscellaneous other whitespace fixes 

👉🏻 After these changes there is just one issue still remaining. IMO the "fix' Slevomat wants there is pretty opinionated, so I leave it up to you whether you want to apply it or ignore the issue.


jaapio:
I will address the last remaining issue :-)

PR #2978, opened 2021-08-08
GH Actions: don't include dev dependencies in PHAR

jrfnl:
As they shouldn't be needed in the production PHAR and will allow for the PHAR file to be smaller.


jrfnl:
Strangely enough, I see no difference in the Phar size.. 😕

jaapio:
Box automatically excludes the dev dependencies ;-)

jrfnl:
In that case, let's close this.

PR #2987, opened 2021-08-17
Poc/rst parser upgrade

jaapio:
First attempt to merge back changes in doctrine/rst-parser back to phpDocumentor.


mvriel:
I have reviewed these changes and it would seem that most of these have become less or even irrelevant with the latest parser changes. The List Item bullet list and method seems to deviate from what we have; that is worth looking into

mvriel:
I just need to extract the changes for the definition list; and then this PR can be closed because all updates are in one form or another incorporated into phpDocumentor

PR #3004, opened 2021-09-15
3002: Render externally written UML Diagrams

### LINKED ISSUE #3002: Render linked PlantUML diagrams with Guides
# Feature request

In our Guide rendering (using RST) we now have the `uml` directive with which you can write a diagram inline; but for larger diagrams it may be more interesting to write the diagram in a separate file and refer to it from the RST. Upon referring to it; phpDocumentor could generate a diagram for it and include the diagram that location.

The UML directive for Sphinx already supports such a thing: https://pypi.org/project/sphinxcontrib-plantuml/

mvriel:
Fixes #3002 

This PR contains support for including UML diagrams from external files into the documentation. The changes here will also solve a bug where you could not use Directives without Blocks (single line directives). Prior to this change a single line Directive would also cause the subsequent paragraph or list to disappear because it would be 'parsed' as part of the directive and discarded

### DIFF
--- a/docs/internals/flow.rst
+++ b/docs/internals/flow.rst
@@ -16,8 +16,6 @@ Activity Diagram.
    :Transform AST into artifacts;
    stop
 
-.. uml:: my-external-file.puml
-
 This three step process enables phpDocumentor to break down a project into its most basic components, called Structural
 Elements, and depending on which template was selected generate various types of output.
 
--- a/docs/internals/flow.rst
+++ b/docs/internals/flow.rst
@@ -16,6 +16,8 @@ Activity Diagram.
    :Transform AST into artifacts;
    stop
 
+.. uml:: my-external-file.puml
+
 This three step process enables phpDocumentor to break down a project into its most basic components, called Structural
 Elements, and depending on which template was selected generate various types of output.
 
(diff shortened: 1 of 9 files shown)


docs/internals/flow.rst
```diff
@@ -16,6 +16,8 @@ Activity Diagram.
    :Transform AST into artifacts;
    stop
 
+.. uml:: my-external-file.puml
```
mvriel, line 19: Note to self: remove this

Mint-6263:
:blush:

Mint-6263:
~~#6263~~

src/Guides/RestructuredText/HTML/Directives/Uml.php
```diff
@@ -56,4 +73,25 @@ public function wantCode(): bool
     {
         return true;
     }
+
+    private function loadExternalUmlFile(Environment $environment, string $path): ?string
```
jaapio, line 77: if I get this method correctly, It will load the contents of the uml as it was inline? So the rendering will pick this up?

Why was the absolute path needed for this? I would expect that flysystem can still load the file? This won't work when a file is outside the root of the documentation dir. I think this would be a valid constraint.

mvriel: > if I get this method correctly, It will load the contents of the uml as it was inline? So the rendering will pick this up?

This is correct; our PlantUML renderer adds a bit of styling and it will thus follow the process like any other UML diagram.

> Why was the absolute path needed for this? I would expect that flysystem can still load the file? 

It is the absolute path on the Origin filesystem and not on the physical filesystem; the origin filesystem's root is at the project folder, but the filename in the environment is relative to the Guide's root. Thus a translation is needed from the Guide root to the Project root

jaapio (approved):

PR #3010, opened 2021-09-16
Fix the width of the source view window for source-files with LOOOOON…

rotdrop:
…G lines.

The commit restricts the maximum size of the source-window to 100vw and set box-sizing to border-box. Otherwise the source-view window can overflow up to the point where not even the close button is accessible.


jaapio:
Thanks!

PR #3011, opened 2021-09-17
Restore functionality of the 'replace' directive

### LINKED ISSUE #3006: Environment variables are empty at "renderVariables".
https://github.com/phpDocumentor/phpDocumentor/blob/aae2e648cc1c9dad55b8ec7cd8488ec34dbdd05a/src/Guides/NodeRenderers/SpanNodeRenderer.php#L117

My replace directives for ResturcturedText files are working with v3.1.0 of phpDocumentor but doesn't work anymore if I update to v3.1.2

I have debugged and the replace directive is processing for all replace directives but at the call of `$spanNodeRenderer->renderVariables(...)` the `$environment->variables` is an empty array.

test.rst
```
.. |test_var| replace:: my test var

Title
=====

Test the var: |test_var|
```

Output:
> Title
>
> Test the var:

mvriel:
In the process of splitting the parsing from the rendering, we did not
take into account that variables were stored on the Environment class.
In the original implementation this Environment class shared state
between just about everything in the whole app.

By pushing the variables into the DocumentDescriptor and restoring it
when rendering, the replace functionality now works again.

Additionally, I found a slight bug in the DocumentParser; when a
Directive is followed by something other than a CodeNode we want to
ignore it and not pass it as a SpanNode to the directive's process
function. In that scenario, the deeper business logic will convert that
SpanNode into a piece of text on the page. Which you do not want.

Fixes #3006

### DIFF
--- a/src/Guides/RestructuredText/Parser/DocumentParser.php
+++ b/src/Guides/RestructuredText/Parser/DocumentParser.php
@@ -505,15 +505,9 @@ class DocumentParser
 
             if ($currentDirective !== null) {
                 try {
-                    // if a Directive is not followed by a CodeNode; this means it totally relies on
-                    // its data and we should not process the $node as part of handling this directive
-                    $directiveNode = $node instanceof CodeNode
-                            ? $node
-                            : new SpanNode($this->environment, $this->directive->getData());
-
                     $currentDirective->process(
                         $this->parser,
-                        $directiveNode,
+                        $node instanceof CodeNode ? $node : null,
                         $this->directive->getVariable(),
                         $this->directive->getData(),
                         $this->directive->getOptions()
--- a/docs/getting-started/index.rst
+++ b/docs/getting-started/index.rst
@@ -1,7 +1,7 @@
 Getting started
 ===============
 
-Is this the first time you work with phpDocumentor? You have come to the right section in the documentation! In the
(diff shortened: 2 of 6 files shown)


src/Guides/RestructuredText/Parser/DocumentParser.php
```diff
@@ -505,15 +505,9 @@ private function flush(): void
 
             if ($currentDirective !== null) {
                 try {
-                    // if a Directive is not followed by a CodeNode; this means it totally relies on
-                    // its data and we should not process the $node as part of handling this directive
-                    $directiveNode = $node instanceof CodeNode
-                            ? $node
-                            : new SpanNode($this->environment, $this->directive->getData());
```
jaapio, line 512: this has been removed? I remember you introduced this recently? when introducing the inline uml?

mvriel: > Additionally, I found a slight bug in the DocumentParser; when a
Directive is followed by something other than a CodeNode we want to
ignore it and not pass it as a SpanNode to the directive's process
function. In that scenario, the deeper business logic will convert that
SpanNode into a piece of text on the page. Which you do not want.

mvriel: Yeah, I found out it causes a weird situation with single-line directives

jaapio (approved):

PR #3018, opened 2021-09-22
False and Null are missing their first letter in a constant's value

### LINKED ISSUE #2866: First character missing from values of global bool constants ('rue', 'alse', 'ull')
1. Create a simple PHP file:
```
<?php
  /**
   * This is description.
   *
   * @var boolean
   */
  const MY_TEST = false;
?>
```
2. Generate API docs with phpDocumentor.
3. Open the docs in the browser. They will contain **MY_TEST = alse** instead of **false** in multiple places. See [this screenshot](https://i.imgur.com/CeuPbET.png).
4. The same thing happens for values **true** and **null**.

## Your environment
phpDocumentor v3.0.0 with PHP 7.4.15 under Windows, default settings.

### LINKED ISSUE #3006: Environment variables are empty at "renderVariables".

mvriel:
We attempted to compensate for an earlier report (issue 2421) by undoing
the adding of slashes by PHP-Parser's PrettyPrinter. However, this had
the unintended side-effect of stripping the first slash when a boolean
or null was provided and subsequently that was interpreted as a unicode
character.

I have removed the un-escaping of the slashes by PHP-Parsers
prettyprinter, reintroducing double slashes in some instances, as this
will ensure that values are presented correctly.

While it may seem odd that a single slash in a single quoted string is
now converted into a double-slash; this notation is technically more
correct as the php documentation states that even in a single quotes
string, a double slash is intended to represent a literal backslash.

See
https://www.php.net/manual/en/language.types.string.php#language.types.string.syntax.single

Although, in code, a single backslash is also used literally; this means
the output by phpDocumentor is not incorrect; albeit verbose perhaps.

This change fixes #2866


jaapio (approved):

src/phpDocumentor/Descriptor/Builder/Reflector/AssemblerAbstract.php
```diff
@@ -17,7 +17,6 @@
 use phpDocumentor\Descriptor\Builder\AssemblerReducer;
 use phpDocumentor\Descriptor\Builder\Reflector\Docblock\DescriptionAssemblerReducer;
 use phpDocumentor\Descriptor\Collection;
-use phpDocumentor\Descriptor\Descriptor;
```
jaapio, line 20: Please re-add this line to fix the phpstan issues.

PR #3020, opened 2021-09-24
Extract directive parsing into subparser

mvriel:


src/Guides/RestructuredText/Parser/DocumentIterator.php
```diff
@@ -0,0 +1,113 @@
+<?php
+
+declare(strict_types=1);
+
```
jaapio, line 4: File license is missing here.

src/Guides/RestructuredText/Parser/DocumentParser.php
```diff
@@ -373,7 +381,11 @@ private function parseLine(string $line): bool
                     }
 
                     $this->flush();
-                    $this->initDirective($line);
+                    $this->subparser = new Subparsers\DirectiveParser($this->environment, $this->lineChecker, $this->lineDataParser, $this->directives);
+                    $directive = $this->subparser->init($line);
+                    if ($directive instanceof DirectiveHandler) {
+                        $this->directive = $directive;
+                    }
```
jaapio: Is here a return missing? when `$directive instanceof DirectiveHandler === false`? Feels like we need a reset if the line didn't match a DirectiveHandler?

src/Guides/RestructuredText/Parser/DocumentIterator.php
mvriel: Well-spotted; will fix in all files

src/Guides/RestructuredText/Parser/DocumentParser.php
```diff
@@ -200,12 +200,17 @@ private function parseLines(string $document): void
         $this->documentIterator->load($this->environment, $document);
         $this->setState(State::BEGIN, '');
 
-        foreach ($this->documentIterator as $line) {
+        // We explicitly do not use foreach, but rather the cursors of the DocumentIterator
+        // this is done because we are transitioning to a method where a Substate can take the current
+        // cursor as starting point and loop through the cursor
+        while ($this->documentIterator->valid()) {
```
jaapio, line 153: I like this :+1:  it makes sense to me that each Production has control over when it stops parsing.

src/Guides/RestructuredText/Parser/DocumentIterator.php
```diff
@@ -49,6 +49,16 @@ public function key(): int
         return $this->position;
     }
 
+    /**
+     * @deprecated Work around for Production's eating one line too many
+     * @todo Revisit The Loop in {@see DocumentParser::parseLines()} and see if the Look Ahead timing should be done
+     *       differently
+     */
+    public function prev(): void
```
jaapio, line 76: The doctrine lexer has a lookAhead method with a special pointer which can be increased by each look ahead. Would that fit here as well? 
I think you had this intention with `getNextLine`

jaapio (commented):
I didn't look for bugs but just at the bigger picture. And I think I can follow what you are doing here. 
Having a production for each state which is able to move the iterator forward makes a lot of sense to me. A minor issue would be that a production rule is able to step back. I would not allow this in the iterator at all. Each line must be peeked by a look ahead or consumed. If we would allow rules to move before they were triggered we could get some odd bugs.

src/Guides/RestructuredText/Parser/DocumentIterator.php
mvriel: I want to fix this in a subsequent PR; the prev is evil and should be gone but I think this is big enough as is :D

src/Guides/RestructuredText/Parser/DocumentParser.php
mvriel: 🎉

src/Guides/RestructuredText/Parser/DocumentIterator.php
```diff
@@ -0,0 +1,99 @@
...
+    public function mergeIncludedFiles(Environment $environment, string $document): string
+    {
+        return preg_replace_callback(
+            '/^\.\. include:: (.+)$/m',
+            function ($match) use ($environment) {
+                $path = $environment->absoluteRelativePath($match[1]);
+
+                $origin = $environment->getOrigin();
+                if (!$origin->has($path)) {
+                    throw new \RuntimeException(
```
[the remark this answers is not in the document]
mvriel: Sounds good to fix that!

src/Guides/RestructuredText/Parser/DocumentIterator.php
```diff
@@ -0,0 +1,99 @@
...
+
+namespace phpDocumentor\Guides\RestructuredText\Parser;
+
+use Iterator;
+use phpDocumentor\Guides\Environment;
+
+/**
+ * @implements Iterator<string>
+ */
+class DocumentIterator implements Iterator
```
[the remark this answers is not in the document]
mvriel: Fixed in next commit

jaapio (approved):

wouterj:
Interesting! I'll have a closer look at these changes this week.
This is getting closer to the state machine parser used by docutils, which should allow following the specs easier.

PR #3041, opened 2021-10-11
POC: extensions 

jaapio:
This PR contains a proof of concept to make phpDocumentor extendable using extensions. 

The code works as is, however we need a number of tests and more stabelized code to merge this. 

- [ ] create 100% code coverage for extension namespace
- [x] add validator for extensions based on manifest data
- [x] test with extensions containing composer dependencies
- [ ] implement phar based extensions.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -44,6 +44,7 @@
         "league/tactician-bundle": "^1.2",
         "league/uri": "~6.3.0",
         "league/uri-interfaces": "^2.0",
+        "phar-io/manifest": "^2.0",
         "phpdocumentor/flyfinder": "^1.0",
         "phpdocumentor/graphviz": "^2.0",
         "phpdocumentor/guides": "*@dev",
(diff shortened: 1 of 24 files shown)


jaapio:
I came across an issue in this implementation which blocks me from making any progress on this. We do ship a phar, which is immutable. In this PR an option was added to restart the kernel and rebuild the cache. In case of a phar install this can be done outside the phar without issues. However this means that we need to include the config. 

The current service configuration contains some issue because of the incubator libraries we have in development right now, these relative paths are not working when in a phar container. This blocks the release of these extensions. But I'm wondering if we are on the right track with this approach.

bshaffer:
Any update on this?

jaapio:
I recreated this pr in #3567, this should provide basic support for extensions. More features will be added and documentation will be updated to provide information developers need to write their own extensions.

PR #3058, opened 2021-10-26
Feature/php81 enums

jaapio:
This pr adds basic support for php 8.1 enums. 

Covered:
- [X] enum in default template
- [X] case rendering 
- [X] methods support in enums
- [X] trait support as far as we do in phpdocumentor 
image


jaapio:
The main issue with this approach are the bundles, we are using phar deployements which are immutable. And do not contain config. Only the bare minimum. This doesn't allow us to recompile the symfony container cache. 
This needs more expiriments to get things working correctly.

PR #3062, opened 2021-10-30
POC: Webpack the dependencies and introduce tailwind

mvriel:


mvriel:
Too invasive. It works for now and I want to fix reported issues

PR #3067, opened 2021-11-13
[POC] resolving hooks in workpress code

jaapio:
This is a POC using the new metadata concept in phpDocumentor to add
extra information in your docs. It is far from usefull as many changes are
needed in the core code to support this case.
This code can be used as reference code to see what we need to change in phpDocumentor
and it reveals missing extension points in our application.


incubator/wordpress/src/Reflection/Php/Factory/Action.php
```diff
@@ -0,0 +1,60 @@
...
+use phpDocumentor\Reflection\Php\Factory\ContextStack;
+use phpDocumentor\Reflection\Php\StrategyContainer;
+use phpDocumentor\Wordpress\Reflection\Php\Action as ActionElement;
+use phpDocumentor\Wordpress\Reflection\Php\Hooks;
+use PhpParser\Node\Expr\FuncCall;
+use PhpParser\PrettyPrinter\Standard as PrettyPrinter;
+
+final class Action extends AbstractFactory
+{
+    private const ACTION_NAMES = [
```
Chrico, line 20: A hook (action or filter) can be in WordPress following:

```
'apply_filters',
'apply_filters_ref_array',
'apply_filters_deprecated',
'do_action',
'do_action_ref_array',
'do_action_deprecated',
```

PR #3178, opened 2022-03-17
Add missing return type from method tags

villfa:
Hi,

When a method is documented with the method tag, its return type is ignored.

Given this sample:
```php
<?php
/**
 * @method string test()
 */
class Foo
{
    public function __call() {
        return 'OK';
    }
}
```

### BEFORE

before

### AFTER

after

I could add some tests if needed, but I'll need some guidance.


jaapio:
Thanks a lot for this fix!. 

I think no tests are required at this moment.

PR #3182, opened 2022-03-21
Missing <base> definition in header

ILL-DIE-TRYING:
There is no <base> definition in the output headers in the "clean" template. I believe this will at least partially solve the class link issues. I have no way to test this so it needs to be tested.


jaapio:
Closing this as https://github.com/phpDocumentor/phpDocumentor/pull/3189 will solve the issue.

PR #3187, opened 2022-03-24
Update function signature to show return by reference

### LINKED ISSUE #3186: Returning a reference is not documented
## Expected behavior
Given this code:
```php
<?php
class Foo
{
    public $bar = 'bar';
    public function &bar() {
        return $this->bar;
    }
}
```

I would expect the generated documentation indicates that the method returns a reference.

## Actual behavior
Here the actual result:
returning-reference

1. Write the code sample from above in `Foo.php`
2. Run `phpdoc -f Foo.php -t doc/`
3. Check the documentation in doc/

## Your environment

I think the environment is not relevant for this issue, but here some details anyway:
* Version used: master
* Install method: Composer
* PHP version: PHP 8.0

villfa:
Fixes #3186

The signature for functions returning a reference now looks like this:
returning-reference-fixed


jaapio:
Thanks for helping us! I like the completeness of this PR. You updated all templates!

PR #3225, opened 2022-05-05
Bump flex to ^1.17, 1.3.1

em411:
https://symfony.com/blog/the-old-flex-infrastructure-is-shutting-down

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -71,7 +71,7 @@
         "symfony/expression-language": "^5.0",
         "symfony/filesystem": "^5.0",
         "symfony/finder": "5.4.*",
-        "symfony/flex": "^1.3.1",
+        "symfony/flex": "^1.17",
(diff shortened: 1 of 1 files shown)


mvriel:
Thank you for your contribution; unfortunately there is a composer.lock update missing. I have tried to merge your PR locally and update the lock file myself so to be able to include your contribution but I ran into a spot of trouble there. 

Unrelated to your PR, there is a version conflict regarding league/url. I will need to fix that first; including the breaking behaviour that the new version of league/url would bring. Please hold on, I will merge your PR as soon as that is resolved

mvriel:
I have manually merged your contribution (see https://github.com/phpDocumentor/phpDocumentor/commit/69ec4a02e17b867be78e58449f155c51a25450dd) and updated the lock file afterwards. Thank you for your help!

PR #3244, opened 2022-06-28
fix: ensure uses/see/link descriptors render to string as expected

### LINKED ISSUE #3243: Inline @see tags are being stripped in long descriptions
When using the current version `v3.3.1` and parsing using the `xml` template, inline `@see` tags are being stripped for long descriptions. According to [the docs](https://docs.phpdoc.org/3.0/guide/references/phpdoc/tags/see.html), the @see tag is valid as inline tags in long descriptions.

## Expected behavior
The contents of the `@see` tags are not stripped, and their contents are preserved in `structure.xml`

## Actual behavior
The contents are stripped

1. Create a file with inline `@see` tags in the long description
```php
class Test
{
    /**
     * An inline see tag in the description works. {@see Test::foo()}.
     *
     * An inline see tag in the long description does not work. {@see Test::bar()}.
     */
    public function foo()
    {
    }
}
```
2. Run `phpdoc --template xml`
3. Notice that the second `@see` reference has been stripped in `structure.xml` and replaced with `{@see}`:
```xml
<method final="false" abstract="false" static="false" namespace="" line="10" visibility="public" returnByReference="false">
    <name>foo</name>
    <full_name>\Test::foo()</full_name>
    <value></value>
    <docblock line="10">
        <description>An inline see tag in the description works. {@see Test::foo()}.</description>
        <long-description>An inline see tag in the long description does not work. {@see}.</long-description>
    </docblock>
</method>
```

## Your environment

* Version used: v3.3.1
* Install method: phar
* PHP version:  7.4.29
* Operating system and version (e.g. Ubuntu 20.04, Windows 10): Mac OSX 12.4

bshaffer:
fixes #3243 

When these descriptors are rendered as part of a `description` in the templates, they are empty

### DIFF
--- a/src/phpDocumentor/Descriptor/Tag/LinkDescriptor.php
+++ b/src/phpDocumentor/Descriptor/Tag/LinkDescriptor.php
@@ -41,4 +41,9 @@ class LinkDescriptor extends TagDescriptor
     {
         return $this->link;
     }
+
+    public function __toString(): string
+    {
+        return (string) $this->link;
+    }
 }
--- a/src/phpDocumentor/Descriptor/Tag/SeeDescriptor.php
+++ b/src/phpDocumentor/Descriptor/Tag/SeeDescriptor.php
@@ -34,4 +34,9 @@ class SeeDescriptor extends TagDescriptor
     {
         return $this->reference;
     }
+
+    public function __toString(): string
+    {
+        return (string) $this->reference;
+    }
 }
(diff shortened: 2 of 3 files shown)


jaapio:
Thanks for creating this patch, I think this was a regression when we refactored the way descriptions are handled. 

Tests that should pass are passing. So this is ready to be merged.

Issue #3184, opened 2022-03-22
space character in filename is URL-encoded


neil-pearce:
## Steps to reproduce the problem

1.
2.
3.

## Your environment

* Version used: 3.0.0-alpha1
* Install method
* PHP version
* Operating system and version (e.g. Ubuntu 20.04, Windows 10):
* Link to your project:
* Attach code that is involved
* ...

web-cooking-factory:
Hi ! Nobody ?
Exact same thing here ...

mvriel:
Thanks for reporting this issue. You are right and this does seem to be an issue with using league/uri's path value object. This is a similar issue to https://github.com/phpDocumentor/phpDocumentor/issues/3247 and https://github.com/phpDocumentor/phpDocumentor/issues/3239; for some reason our test cases did not pick up on this. 

As mentioned in 3239, we'd need a bit of time to fix this because we resolve paths across operating systems but also across virtual filesystems. And since the problem may be in the nature of League URI's path object itself, I have a gut feeling that we need to rip that out and replace it with something that works better.

mvriel:
A Draft PR is opened at https://github.com/phpDocumentor/phpDocumentor/pull/3253 containing at least a couple of test cases illustrating the issue. I still need to discern what more locations there may be, and what scenario's I am still missing

mvriel:
> The issue is being caused because the function createFromWindowsPath in League\Uri is performing a rawurlencode on the elements of the filename:

Thank you for your research, I'm not sure I would have found this easily. My testcases also seem to indicate there is a similar issue in a non-windows environment. Which makes this even more interesting

neil-pearce:
I was afraid that it wouldn't be a simple - I basically don't have a grasp of all the complexities. Thanks for picking it up

mvriel:
The first few tries to fix it haven't worked so far; I need to refactor quite a bit because of assumption errors. It also doesn't help that our Windows unit tests won't run at the moment..

A2020GK:
My problem is quite large: directory name, that's in Russian!
I've got a lot of problems with OneDrive.

michalb314:
Was this fixed? I have the same problem, trying to generate docs for a site hosted in Local, folder is inside "Local Sites", docs get generated in "Local%20Sites". Annoying to say the least.

PR #3253, opened 2022-07-09
Paths containing special characters paths in the configuration are wrongly resolved

mvriel:
See https://github.com/phpDocumentor/phpDocumentor/issues/3184 and related tickets for the original report.

When using special characters, these are URL encoded and phpDocumentor is unable to process this. What needs to be done is that phpDocumentor is able to support physical and virtual filesystems (through the DSN) and know when to URL encode and when not. Some virtual file systems may require this.

Part of the solution in this PR requires investigating when and if we should or should not URL encode


phpdoc.dist.xml
```diff
@@ -8,6 +8,7 @@
     <title>phpDocumentor</title>
     <paths>
         <output>build/website/docs</output>
+        <cache>build/ex$a#mple cache/</cache>
```
mvriel, line 11: Don't forget to remove this before submitting as a definitive PR :D

Issue #3160, opened 2022-03-02
Custom Template Location not able to be read / can't be found


ajcampbell1981:
I am trying to implement a custom template in my phpdoc.dist.xml file.  However, when I run `php phpDocumentor.phar`, it results in an error that states `The given template ... could not be found or is not readable`.  

## Expected behavior
I would expect that either a) the folder would be able to be found; or b) that it would give a little bit more information as to "where it's looking"  ie: 'Could not find theme in /ROOT_DIR/FOLDER_LOCATION_SEARCHED'.

## Actual behavior
As described above, it is just giving an error that the template could not be found.

## Steps to reproduce the problem

1.  Create a new template in PROJECT_ROOT/phpdocs/templates folder/new_template_name
2.  Add the following line to the phpdoc.dist.xml: `<template name="new_template_name" location="phpdocs/templates/">`
3.  Run the command above `php phpDocumentor.phar`

## Your environment

* Version used: 3.0.0
* PHP version 7
* Operating system and version (e.g. Ubuntu 20.04, Windows 10): Ubuntu 20.04

alanef:
I'm struggling with the same issue. Just seem to able to get it to find a custom template.  Did you resolve?

ajcampbell1981:
> I'm struggling with the same issue. Just seem to able to get it to find a custom template. Did you resolve?

I did not.  Still trying to figure it out.

alanef:
I some how got it working,  but not sure how / what made a difference

I created a folder in my project root  called data/templates   and put the template folder  in there

and set up in my phpdoc.dist.xml in the project root
`<template name="data/templates/fullworks"/>`

any attempt to load from vendor/xyz/etc  didn't seem to work

then I had a fun day deconstructing the 'default' template to strip out and simplify and remove CSS to just what I need - the documentation for the public API methods  and finally wrapping it up in a WordPress shortcode

Still WIP  by this is where I am https://staging.fullworks.net/2022/03/wfea-api-docs/

smileBeda:
Thanks @alanef for that - clearly the DOC is wrong.
It says to pass a _name_ and a _location_ but that won't work
What works is indeed `<template name="relative/path/to/where/your/config/file/is"/>`

@jaapio / @mvriel not sure entirely who is responsible for DOC issues, but this should be addressed, and does not need too much work to fix and avoid people getting frustrated.

[...]

PR #3273, opened 2022-07-17
Support Location attribute on templates

mvriel:
In this commit, I have added handling, tests and documentation for the
'location' attribute in the template configuration entry. This should
now work and attempt to load the template from the given directory on
the host's filesystem.

Because I kept running into issues with how phpstan and psalm interpret
the configuration array shape; I have also cleaned that part up and used
psalm-type definitions to provide a reusable shape. I chose not to use
phpstan-type annotations because phpstan understands the psalm
annotations but not vice versa.

Unfortunately, I discovered that we are using an antiquated version of
phpstan that does not like array-shape type aliases. This is fixed in a
newer version but I need @jaapio his input whether we version locked for
a reason.

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -163,7 +163,7 @@ jobs:
           composer-options: --optimize-autoloader
 
       - name: PHPStan
-        uses: phpDocumentor/phpstan-ga@master
+        uses: phpDocumentor/phpstan-ga@1.8
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         with:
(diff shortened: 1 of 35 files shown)


mvriel:
Once complete, this should fix #3160.

jaapio (approved):

jaapio:
Changes are looking good, but you have some failing tests, and phpstan is complaining :-)

mvriel:
Odd. Locally everything was green. I'll look at it once the temperature drops

mvriel:
Should be fixed now

jaapio:
Thanks @mvriel!

Issue #3147, opened 2022-01-31
phpdoc/phpdoc 3.3 fails with "Constant::isFinal(): Return value must be of type bool, null returned"


caugner:
## Expected behavior

The `phpdoc/phpdoc:3.3` image should run successfully.

## Actual behavior

```
Using docker image sha256:3caa4af29d3d561cea4d315b31d02cf4bcd5d21e971772df15ecbb877253a5c2 for gitlab.***.***:443/***/dependency_proxy/containers/phpdoc/phpdoc:3.3 with digest gitlab.***.***:443/***/dependency_proxy/containers/phpdoc/phpdoc@sha256:19a88bc569afcd8f4920015b24f6566cc45eab71b2ddb76ca811d448a6bf15ef ...
$ php -v
PHP 8.0.15 (cli) (built: Jan 20 2022 20:08:16) ( NTS )
Copyright (c) The PHP Group
Zend Engine v4.0.15, Copyright (c) Zend Technologies
$ phpdoc run -d src -t docs
phpDocumentor vdev-master@4c31809
Parsing files
    1/1127 [>---------------------------]   0%
  113/1127 [==>-------------------------]  10%
  339/1127 [========>-------------------]  30%
  451/1127 [===========>----------------]  40%
  564/1127 [==============>-------------]  50%
  677/1127 [================>-----------]  60%
  902/1127 [======================>-----]  80%
 1015/1127 [=========================>--]  90%
 1127/1127 [============================] 100%12:03:58 CRITICAL  [console] Error thrown while running command "run -d src -t docs". Message: "phpDocumentor\Reflection\Php\Constant::isFinal(): Return value must be of type bool, null returned" ["exception" => TypeError { …},"command" => "run -d src -t docs","message" => "phpDocumentor\Reflection\Php\Constant::isFinal(): Return value must be of type bool, null returned"]
In Constant.php line 120:
                                                                               
  phpDocumentor\Reflection\Php\Constant::isFinal(): Return value must be of t  
  ype bool, null returned                                                      
                                                                               
[...]

PR #3289, opened 2022-07-23
Purge caches upon version change

mvriel:
In issue https://github.com/phpDocumentor/phpDocumentor/issues/3147, it was noted that there might very well be caching issues at play after a version upgrade. Since in the past we had a mechanism to force a cache purge after updating phpDocumentor, I dived into the caching mechanism to see if that was broken.

I found that the caching mechanism indeed no longer purges after a version change, this PR will reintroduce that behaviour and add extra documentation on how caching works.

TODO

- [x] Make version number available to purging mechanism
- [x] Add version number to cache namespace, if the version changes the cache is 'empty'
- [x] Write cache docs

### DIFF
--- a/config/packages/framework.yaml
+++ b/config/packages/framework.yaml
@@ -15,10 +15,8 @@ framework:
         pools:
             files:
                 adapter: app.projectcache.provider
-                default_lifetime: 2629743 # 1 month
             descriptors:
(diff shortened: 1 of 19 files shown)


jaapio (approved):

jaapio:
Thanks, I like the new documentation part!

PR #3294, opened 2022-07-23
Line numbers show 0 with InvalidTag issues

### LINKED ISSUE #3057: No line number for errors
## Expected behavior
Errors are shown with the line number where the problem was found in the report

## Actual behavior
Line number is always 0.
e.g. https://docs.phpdoc.org/3.0/reports/errors.html
image

Run PHPDoc on any source code that contains errors

## Your environment

* Version used: 3.1.2
* Install method PHAR
* PHP version 8.0.11
* Operating system and version Ubuntu 20.04.3 LTS
* Link to your project: problem exists with your own documentation https://docs.phpdoc.org/3.0/reports/errors.html

mvriel:
The InvalidTag does not know its line number or location and as such the
errors overview in the generated documentation shows a line number of 0.
This is not quite helpful.

To improve the situation, I have introduced that any line number of 0
will be overridden with the containing element's line number when you
retrieve the error listing.

Fixes #3057

### DIFF
--- a/src/phpDocumentor/Descriptor/DescriptorAbstract.php
+++ b/src/phpDocumentor/Descriptor/DescriptorAbstract.php
@@ -18,6 +18,7 @@ use phpDocumentor\Descriptor\Tag\AuthorDescriptor;
 use phpDocumentor\Descriptor\Validation\Error;
 use phpDocumentor\Reflection\Fqsen;
 use phpDocumentor\Reflection\Location;
+use Webmozart\Assert\Assert;
 
 use function lcfirst;
 use function strpos;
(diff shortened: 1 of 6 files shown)


src/phpDocumentor/Descriptor/DescriptorAbstract.php
```diff
@@ -430,6 +431,20 @@ public function getErrors(): Collection
             }
         }
 
+        foreach ($errors as $error) {
```
jaapio, line 434: I'm wondering if we should remove these types of compiling / building steps from the descriptors.
Maybe not for now, but I would like to have less logic in those classes.

jaapio (approved):

src/phpDocumentor/Descriptor/DescriptorAbstract.php
mvriel: Agreed, I tried to add this in the InvalidTagAssembler; but that doesn't know enough about where it's at. :) and since this location merges all 'sub'-errors; I had to add it here :(

PR #3311, opened 2022-08-17
Extend timeout for UML generation to 20mins

Hackwar:
Hello, we have a larger codebase with ~800 files (according to the parse process of phpdocumentor) and while executing this in the docker image on our CI system, we regularly run into timeouts from the UML process. After 10 minutes it simply aborts and the rendering process never finishes. This change here is the attempt to fix that.


jaapio:
Thanks

PR #3312, opened 2022-08-17
Fix Code Block Options

amayer5125:
Currently `:linenos:` shows up in the code block. I checked all other pages for the same issue but didn't see anything.
image


jaapio:
Thanks!

PR #3324, opened 2022-09-24
Fix OutputHello Example Param Wording

CraicOverflow89:
Just a tiny one: fixed wording of `outputHello` example param description... it suggested that when `$quiet` was true it _would_ echo 'Hello World' when in actual fact, that's the opposite of what the example code does.


jaapio:
Thanks

PR #3362, opened 2022-11-19
GH#3307 Make readonly mutator for class available

mvriel:
When a class has the readonly keyword, we should be able to show that in the generated documentation. These changes will ensure that the twig templates can read whether a whole class is readonly.

image

This is now shown as a label with the text when present, if absent: it is omitted. This will now also show that label for abstract and final classes, and there are e2e tests for this behaviour.

I am not 100% convinced of the choice for this UI, but it is an improvement over what we have now and I am looking into a slight update of the styling to clean a few things up

### DIFF
--- a/data/templates/default/components/tags.html.twig
+++ b/data/templates/default/components/tags.html.twig
@@ -1,6 +1,6 @@
 {% set tags = node.tags|filter((v,k) => k not in ['var', 'param', 'property', 'property-read', 'property-write', 'method', 'return', 'package', 'api']) %}
 
-{% if tags|length > 0 %}
+{% if tags|length > 0 and tags|first|length > 0 %}
     <h5 class="phpdocumentor-tag-list__heading" id="tags">
         Tags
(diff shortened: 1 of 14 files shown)


jaapio:
I just merged another pr that required v5.3 of reflection. So please remove that part from this pr :-)

data/templates/default/components/tags.html.twig
```diff
@@ -1,6 +1,6 @@
 {% set tags = node.tags|filter((v,k) => k not in ['var', 'param', 'property', 'property-read', 'property-write', 'method', 'return', 'package', 'api']) %}
 
-{% if tags|length > 0 %}
+{% if tags|length > 0 and tags|first|length > 0 %}
```
mvriel, line 3: This will fix all the loose `tags` headings

mvriel:
The build of this PR currently seems to break due to an issue on master and not in this PR, it uses a method from ReflectionDocBlock that is not released yet

PR #3379, opened 2022-11-24
Introduce CanHaveAType and CanHaveADefaultValue traits for Descriptors

mvriel:
In this PR, I have revisited the ArgumentDescriptor, introduced a couple more traits and expanded on the test suite for this class. In a subsequent PR I want to move more elements into traits and add e2e tests (I noticed some are missing) but first it would be nice to have this reviewed instead of a monster PR :D

### DIFF
--- a/src/phpDocumentor/Descriptor/Traits/CanHaveAType.php
+++ b/src/phpDocumentor/Descriptor/Traits/CanHaveAType.php
@@ -13,10 +13,8 @@ declare(strict_types=1);
 
 namespace phpDocumentor\Descriptor\Traits;
 
-use phpDocumentor\Descriptor\Interfaces\ConstantInterface;
 use phpDocumentor\Descriptor\Interfaces\InheritsFromElement;
-use phpDocumentor\Descriptor\Interfaces\PropertyInterface;
-use phpDocumentor\Descriptor\Tag\BaseTypes\TypedAbstract;
+use phpDocumentor\Descriptor\IsTyped;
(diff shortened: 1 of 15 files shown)


src/phpDocumentor/Descriptor/Traits/CanHaveAType.php
```diff
@@ -0,0 +1,56 @@
...
+{
+    /** @var Type|null $type normalized type of this argument */
+    protected ?Type $type = null;
+
+    public function setType(?Type $type): void
+    {
+        $this->type = $type;
+    }
+
+    public function getType(): ?Type
```
jaapio, line 35: Should type be optional? 
- in case of an argument, when it is missing it can be mixed?
- in case of an function / method return type the result is void? (This comes from the PSR-19 spec AFAIK)

jaapio (approved):
This looks like a nice improvement.

src/phpDocumentor/Descriptor/Traits/CanHaveAType.php
mvriel: This is a really good question. There are more elements in the descriptors for which we can ask: should this be nullable. Given that I do not exactly what will break when I remove the nullability from this element I have chosen to postpone addressing this issue :)

Concerning this whole effort: I have mentally prepared myself that I/we need to go over each Descriptor and trait several times to improve it. This is intended to prevent too many monster pr and isolate potentially dangerous changes in a single PR :)

PR #3388, opened 2022-11-29
Restyle elements and add color support

mvriel:
To make adoption easier, I have added support for setting a color using
the command line using the settings functionality. The setting in
question is 'template.color'.

Because settings were a flat structure before and I wanted to add
multiple settings under the template umbrella I changed it to explode on
a '.' in the key.

In addition, the elements have been changed visually to be more legible
and use a different font.

Supported colors are the material design colors:

- red
- pink
- purple
- deep-purple
- indigo
- blue
- light-blue
- cyan
- teal
- green
- light-green
- lime
- yellow
- amber
- orange
- deep-orange
- brown

And: phpdocumentor-green, which is the default.

Example of the red theme:

image
image

TODO:

- [x] Check styling of all elements for inconsistencies
- [x] Fix issues in mobile view:
    - Signatures should not have the pull to left
    - Long names go out of the box, reduce font-size and/or break word
    - File indicators overlap with element names, on mobile they should not be on the same line
- [x] Change icon rendering to use new colors
- [x] Add unit tests for changes in settings handling
- [x] Document the changes


mvriel:
The currently failing build pipeline seems to have something to do with the setting for guides not being recognized and returning true. Spawning a subprocess for the UML tag fails (don't know why it would fail, but for now a good signaling feature)

PR #3390, opened 2022-12-03
Convert tabs to spaces when parsing ReST

mvriel:
RestructureText is a line-based indentation sensitive format. To support tabs and make this easier to parse, we need to convert tab indentation into spaces internally. This will allow for a universal bit of logic when dealing with indentation.

### DIFF
--- a/incubator/guides-restructured-text/src/RestructuredText/Parser/LinesIterator.php
+++ b/incubator/guides-restructured-text/src/RestructuredText/Parser/LinesIterator.php
@@ -18,8 +18,11 @@ use OutOfBoundsException;
 
 use function chr;
 use function explode;
+use function preg_replace_callback;
 use function sprintf;
+use function str_repeat;
 use function str_replace;
+use function strlen;
 use function trim;
 
 /**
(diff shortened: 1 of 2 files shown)


incubator/guides-restructured-text/src/RestructuredText/Parser/LinesIterator.php
```diff
@@ -28,18 +31,32 @@
...
+        foreach ($this->lines as $key => $line) {
+            $this->lines[$key] = $this->tabToSpaces($line);
+        }
+
         $this->rewind();
     }
 
+    private function tabToSpaces(string $line): string
+    {
+        $tabStop = 8;
```
jaapio, line 51: Should this not be a setting? As people can mix their content using spaces and tabs the tabstop will make sense to configure. A tabstop of 2 will give a different results than a stop of 8

incubator/guides-restructured-text/src/RestructuredText/Parser/LinesIterator.php
```diff
@@ -28,18 +31,32 @@
...
-    private $position = 0;
+    private array $lines = [];
+    private int $position = 0;
 
     public function load(string $document): void
     {
         $document = trim($this->prepareDocument($document));
         $this->lines = explode("\n", $document);
+
+        foreach ($this->lines as $key => $line) {
```
jaapio, line 42: Should this not be part of the prepare document function?

mvriel: The prepareDocument function treats the contents as a whole, and before I can remove tab indentation it needs to be split into lines. A search and replace on all tabs could be detrimental.. If I were to move this into the prepareDocument method, then its output would change to lines instead of text. Is that desired?

jaapio: It makes sense to do it by line. As we should only replace it at the start of the line. Leave it where it is right now 👍

incubator/guides-restructured-text/src/RestructuredText/Parser/LinesIterator.php
mvriel: As far as I was concerned, I wanted to keep it simple instead of adding just another setting :) Even if you mix content, it would not change the way the application works (unless you are inconsistent in your tab and space use, but I'm not sure a setting would help there ;) )

PR #3393, opened 2022-12-04
Fix aggregated types sometimes rendered wrong; closes 3323

Mai-Lapyst:
Fixes the rendering of aggregated types being rendered wrong when using them for example as value type in an array.

Needs a patch for phpdocumentor-typeresolver ([see here](https://github.com/phpDocumentor/TypeResolver/pull/179)) to add a method `getToken()` to actually get the token (seperator) from the aggregated type.


jaapio:
Thanks for taking the time to create a PR! we solved this issue in a more advanced way by introducing a new system to render collections. This new system allows us to do more advanced rendering of types as phpDocumentor will support callables in the future for example. 

This is how it will look in the next release:
image

PR #3475, opened 2023-03-14
Refactor LinkRenderer into a series of adapters and introduce references to guides

mvriel:
The LinkRenderer is rather complicated, and when you look closer at it,
this has 2 reasons:

1. it violates the SRP and actually does 3 things
2. it violates the SRP even more because the linkrendering itself is
actually a couple of adapters in a chain of responsibility all coded
into a single method.

This sets the stage for transforming the LinkRenderer into a chain of
responsibility with adapters per type of object that needs to be
transformed and to extract the presentation style to a decorator or
another type of adapter.

The end goal was to support a new type of adapter to parse the URL 
starting with a `doc://` virtual scheme in a different way, so that you 
can link to guides from phpDocumentor using, for example, the 
`@see` tag. This has been accomplished.

**Note**

Refactoring this part of the code raised a few questions that I am not 
going to address with this PR. This PR is getting big enough as is.

Answering these questions might discover bugs that we can fix, or 
improvements to the whole routing system.


jaapio:
I like this change very much. I have no comments about the code at this point. Should be ready to go when you think it's ready.

PR #3482, opened 2023-03-18
Remove file and index getters and setters from project

mvriel:
In this change, I have begun the effort of moving stuff out of the ProjectDescriptor that doesn't belong there into the DocumentationSetDescriptors. This also means Project references might be deleted from some services in favour of this one.

This doesn't yet replace the getFiles and getIndexes calls because these are used in a lot of files and warrants a separate PR

### DIFF
--- a/src/phpDocumentor/Pipeline/Stage/Transform.php
+++ b/src/phpDocumentor/Pipeline/Stage/Transform.php
@@ -6,6 +6,7 @@ namespace phpDocumentor\Pipeline\Stage;
 
 use Exception;
 use League\Flysystem\FilesystemInterface;
+use phpDocumentor\Descriptor\ApiSetDescriptor;
 use phpDocumentor\Dsn;
 use phpDocumentor\Event\Dispatcher;
 use phpDocumentor\Parser\FlySystemFactory;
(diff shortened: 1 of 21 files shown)


src/phpDocumentor/Pipeline/Stage/Transform.php
```diff
@@ -87,26 +88,32 @@ public function __invoke(Payload $payload): Payload
...
-        $preTransformEvent = PreTransformEvent::createInstance($this);
-        $preTransformEvent->setProject($project);
-        $preTransformEvent->setTransformations($transformations);
-        Dispatcher::getInstance()->dispatch(
-            $preTransformEvent,
-            Transformer::EVENT_PRE_TRANSFORM
-        );
+        foreach ($project->getVersions() as $version) {
+            $apiSets = $version->getDocumentationSets()->filter(ApiSetDescriptor::class);
+            foreach ($apiSets as $apiSet) {
```
mvriel, line 93: Just to clarify, using multiple documentation sets would still not work because it is not fully supported when it comes to linking and even writing to the correct folder; but this will ensure the documentation set is propagated throughout the chain, which allowed me to remove findElement from the Project

src/phpDocumentor/Transformer/Writer/Twig/LinkRenderer.php
```diff
@@ -62,6 +64,9 @@ public function __clone()
         $this->adapters = $this->createAdapters();
     }
 
+    /**
+     * @deprecated will be removed once getProject is removed
```
jaapio, line 68: I'm not sure if we should remove the full project from the rendering... You might need the project in some cases.
For example when you need a list of versions in a template.

mvriel: As far as I am concerned, we can reintroduce that at that moment. Partly because I am not sure whether this specific way of handling will be preferable or how linking will come together.

jaapio (approved):

PR #3485, opened 2023-03-20
Extract pipeline for documentation set

mvriel:
With this change, I have extracted the parsing of each API set into a
sub-pipeline. This means that we can now parse multiple sets of API
Documentation instead of just the first.

However, it doesn't work quite yet. As the files are still added onto
the ProjectDescriptor. This change merely changes the process, but
shouldn't affect the application as long as the guard that only 1 API
Set may be present is in place.

In a subsequent change, we will be modifying the parsing logic itself to      
work more with the Documentation Set descriptors instead of the
ProjectDescriptor.

With this, I updated the pipeline documentation to reflect the new 
situation and I was able to drop the Parser Payload as it had been 
replaced by the ApiSetPayload. I also found that Guides type-hinted 
the parser payload but didn't really use it, so I have simplified that 
part of the code and subsequently was able to remove the Parser 
Payload, including the pipeline stage that maps it, entirely.


mvriel:
I still need to update tests and what comes out of static analysis

PR #3486, opened 2023-03-21
Collect files and indexes in API Documentation Set

mvriel:
**Important: this PR is based on the work in https://github.com/phpDocumentor/phpDocumentor/pull/3485, and that should be reviewed and merged first**

In this change, I have changed the way the ParseFiles stage works by
having it directly push the parsed files into the ApiSetDescriptor
instead of the ProjectDescriptor.

Since the ProjectDescriptor's Files and Indexes collections are proxies
for the first ApiSet at the moment, nothing changes in the way
phpDocumentor works. This change is again moving forward with the plan
to remove the getFiles and getIndexes from the ProjectDescriptor. As
soon as that is gone, we have successfully decoupled documentation sets
from the main project and be able to support multiple of them.


src/phpDocumentor/Pipeline/Stage/Parser/ParseFiles.php
```diff
@@ -41,19 +33,17 @@ public function __construct(
...
-            return $payload;
-        }
+        $apiConfig = $payload->getApiSet()->getSettings();
 
         $builder = $payload->getBuilder();
-        $builder->setApiSpecification($apiConfig);
+        $builder->usingApiSpecification($apiConfig);
+        $builder->usingDefaultPackageName($apiConfig['default-package-name'] ?? '');
+
+        // TODO: The setVisibility call should purge the cache if it differs; but once we are here, cache has already
```
jaapio, line 44: I think when we move the filtering into the compiler... we do not need to invalidate the cache... filtering should just transform the output of the parser stage, it will not harm to have a pure parser output. As the filters are already running over the descriptors.

mvriel: IIRC, the visibility part was originally in the parser instead of the compiler so that 'unauthorized' code would not end up in the cache or in the ast dump file. This as a means of protecting IP even though things could be distributed. I am doubting whether it is a real thing, but that was the original reason :)

PR #3492, opened 2023-03-23
Refactor all compilers to work on a DocumentationSet

mvriel:
Since the compilers work on files and indexes on a DocumentationSet, the entirety of them should really interact with a DocumentationSet and not a project

### DIFF
--- a/src/phpDocumentor/Compiler/Pass/Debug.php
+++ b/src/phpDocumentor/Compiler/Pass/Debug.php
@@ -14,7 +14,7 @@ declare(strict_types=1);
 namespace phpDocumentor\Compiler\Pass;
 
 use phpDocumentor\Compiler\CompilerPassInterface;
-use phpDocumentor\Descriptor\Interfaces\ProjectInterface;
+use phpDocumentor\Descriptor\DocumentationSetDescriptor;
 use phpDocumentor\Descriptor\ProjectAnalyzer;
 use Psr\Log\LoggerInterface;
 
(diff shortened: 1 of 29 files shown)


src/phpDocumentor/Compiler/Pass/Debug.php
```diff
@@ -48,11 +48,11 @@ public function getDescription(): string
         return 'Analyze results and write report to log';
     }
 
-    public function __invoke(ProjectInterface $project): ProjectInterface
+    public function __invoke(DocumentationSetDescriptor $documentationSet): DocumentationSetDescriptor
     {
-        $this->analyzer->analyze($project);
+        $this->analyzer->analyze($documentationSet);
```
jaapio, line 53: This should only run on apiSet?

jaapio (commented):
I'm missing some checks on ApiSetDescriptor in some of the compiler passes. Did I overlook something?

mvriel:
> I'm missing some checks on ApiSetDescriptor in some of the compiler passes. Did I overlook something?

I have explicitly not added these and tried to keep it as generic as possible. At the moment, guides do not use files and indexes; but I would love to have this a generic mechanism across all Documentation Sets.

At this point, because only API Sets populate these indexes, I have limited the checks on API Sets only where it really is about API Sets. You can find a few of these in the compiler passes themselves

src/phpDocumentor/Compiler/Pass/Debug.php
mvriel: The analyzer itself does varying things depending on the type of documentation set

PR #3497, opened 2023-03-25
Introduce views to make sure templates keep working

mvriel:
The templates currently rely on the project descriptor, but we want to
get rid of the getFiles and getIndexes methods in the Project
Descriptor.

Instead of changing all the templates, I am looking at introducing a
View layer in phpDocumentor so that we can replace the direct use of
these top level descriptors with a set of views.

The idea behind this is that in the long run, views are aware of the
type of documentation set and can change their output depending on this.

But, give it a couple more commits to crystalize. I am also not too sold
on it being a static factory method at the moment..

### DIFF
--- a/src/phpDocumentor/Transformer/View/DefaultViewSet.php
+++ b/src/phpDocumentor/Transformer/View/DefaultViewSet.php
@@ -1,0 +1,78 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * This file is part of phpDocumentor.
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ *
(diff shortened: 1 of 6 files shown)


mvriel:
@jaapio I am not quite done with this change yet, but can I have your opinion on adding a view instead of changing the templates? I hope to keep it BC-compatible with people using custom templates through this

jaapio:
Closing this, as the rendering of guides has changed to much to make this approach work. We need to revisit this when we want to implement this. 

Neverless I like the idea of having a viewmodel that can be used to create a bc layer in the templates. So I will give this a try myself. To see if this helps to merge the rendering of guides and our own twig templates.

PR #3507, opened 2023-04-21
@unsafe tag

jrcii:


jaapio:
Thanks for your interest in phpDoc. We do appreciate the time you spend to document this new tag. However we do not want to add this to phpDocumentor as a specialized tag. Adding a tag should impact phpDocumentor in a certain way. This tag doesn't add any special behavior to our application. So there is no particular reason to document it as such. 

I might be wrong on this, but I think this case is something special for your project. If not please provide some examples of projects using this tag. So we can decide whether this should be part of the specification. 

---

Having that said, phpDocumentor supports any tag added to the docblocks as tag. You can add this new tag to your docblocks and it will show up in you documentation including the features we support on all descriptions. So for example you can do this:

```
/**
 * @unsafe this is an unsafe method, better use {@see safeMethod}
 */
```
This will render as:
image

There is no reporting available right now to list elements by tag. Which might be an interesting feature to have a look at. 

--- 

Thanks for your time.

PR #3514, opened 2023-05-22
feat: add class magic methods to xml template, fix magic method return tag

bshaffer:
## Adds class magic methods to XML template

Magic methods from class annotations (e.g. `@method`) are not being added to `structure.xml` (e.g. `--template=xml`),. This PR adds them

## Fixes class and trait magic method return tags

When setting the `return` tag for magic methods in traits and classes, the call to `filter` is breaking the reference chain. The `filter` method returns a _new_ collection, and so anything set with `add` will not be set on the original `Collection` as intended. The method `EnumDescriptor::getMagicMethods` does not have this issue.

### DIFF
--- a/data/templates/xml/structure.xml.twig
+++ b/data/templates/xml/structure.xml.twig
@@ -84,6 +84,9 @@
                     {% for method in class.inheritedMethods %}
                         {{ include('method.xml.twig', {method: method, inherited_from: method.parent}) }}
                     {% endfor %}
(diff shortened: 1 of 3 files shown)


jaapio:
Thanks, could you please have a look at the code style issues, after that this seems to be good to go.

jaapio:
Yeah, windows tests are currently broken nothing to do with your PR. 

Thanks for your contribution!

PR #3520, opened 2023-06-05
feat: add magic flag to xml template

bshaffer:
We added support for [magic methods added to the Class template](https://github.com/phpDocumentor/phpDocumentor/pull/3514) recently, but in the generated `structure.xml`, there's no way to tell if they've been inherited, or if they're magic methods. 

Summary of changes: 

 - sets `inherited_from` for magic methods
     - This logic is determined in the template when `method.class` does not match `class`. This may not be the best way, but it seemed like the approach with the least churn
     - this feature is (I think) important, since it exists for other methods, and so the behavior now is inconsistent
 - a `magic="true"` attribute to the `<method>` tag (e.g. `<method /* ... */ magic="true">`)
    - I think it makes sense to have some way to know if the methods are magic or not, but since the users could check the existence of the `@method` tag, that's already possible, so I'm 100% okay getting rid of this. 

I'd love some feedback here!!

**Note**: `magic` is probably the wrong label to use for methods added with `__call`, because according to the PHP Documentation, [PHP magic methods](https://www.php.net/manual/en/language.oop5.magic.php) are methods prefixed with `__`.  The methods added with `__call` are referred to as "[overloaded methods](https://www.php.net/manual/en/language.oop5.overloading.php#object.call)", which I also don't like, as that represents something different in other OO languages such as Java.


jaapio:
I think magic makes sense... as the methods do not exist... but are called in a magic way :-) 

I have no better name for them.

PR #3539, opened 2023-07-20
Feature/filesource tag

### LINKED ISSUE #3200: Problem including source code
# Template issue/request

I am having an issue including source code in the generated documentation. It may well be that I do not understand the functionality but there appears to be some sort of issue.  
  
If I use the **default** template there are not any files generated under the artifact name defined in _templates\default\template.xml_ (ie files/{{path}}.txt). I believe that this is because the source file _templates/default/source/file.source.txt.twig_ does not exists (I'm not sure why that twig file is defined under the sub-directory _source_ but that is a different matter).  
  
If I use the **clean** template then files are generated under the artifact name but they are all empty. That may be because the _template.xml_ for clean refers to **transformation writer="twig"** whereas the writer is **sourcecode** for the default template.  

As an aside, because of the PHAR it is difficult to make changes to test these theories but that may well be as a result of my ignorance!
  

- Version used: phpdocumentor-3.3.1
- Install method: PHIVE
- PHP version: PHP 8.1.4 (cli) (built: Mar 16 2022 09:33:31) (ZTS Visual C++ 2019 x64)
- Operating system and version: Windows 10
-

### LINKED ISSUE #3336: @filesource tag dont work?
- **phpDocumentor version:** 3.3.1 (phar)
- **template:** default (but i have tried others)
- **test php file:**
```
<?php
/**
 * @filesource
 */

/**
 * @author alice
 */
class Test {}
```
- **cli run command:** `php phpDocumentor.phar -d . -t result`
- **what i get:**
image
- **what i expect:** I'm not sure, maybe a demonstration of the full source code of the file on this page?

Hello. I'm trying to understand how and if the `@filesource` tag works. Its addition does not lead to any visible result.

jaapio:
This includes the new feature for filesource tags and updated the documentation regarding this feature. 

Fixes #3336
Fixes #3200

### DIFF
--- a/data/templates/default/components/element-found-in.html.twig
+++ b/data/templates/default/components/element-found-in.html.twig
@@ -4,6 +4,6 @@
     <span class="phpdocumentor-element-found-in__line">{{ node.line }}</span>
 
     {% if node.file.source %}
-        <a href="{{ target_path }}#source-view.{{ node.line }}" class="phpdocumentor-element-found-in__source" data-line="{{ node.line }}" data-modal="source-view" data-src="{{ path('files/' ~ node.path ~ '.txt')|raw }}"></a>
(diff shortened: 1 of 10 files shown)


jaapio:
This feature will be directly available in our docker image, and is part of the next phar release.

PR #3563, opened 2023-08-18
Enable Code Coverage when locally running PHPUnit

mvriel:
I occasionally want to add a unit test from the backlog of unwritten tests. To do this, I need the code coverage report but that was disabled a while ago.

To reintroduce the code coverage in a non-intrusive way I have introduced a new build stage in the Dockerfile that will install pcov, and I make sure the production builds stop at the prod target.

The added benefit from this is that we use the phpdoc container to run the tests in, meaning that if a test fails in the container it will surely fail in production as both environments are identical

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -29,17 +29,17 @@ RUN echo "memory_limit=-1" >> /usr/local/etc/php/conf.d/phpdoc.ini && phpdoc cac
 
 ENTRYPOINT ["/opt/phpdoc/bin/phpdoc"]
 
-FROM phpdoc_base as dev
+COPY --from=composer /usr/bin/composer /usr/bin/composer
+
+FROM phpdoc_base as prod
+
+FROM prod as dev
 
 RUN apt-get update \
     && apt-get install -yq git \
     && useradd -m -s /bin/bash phpdoc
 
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -38,3 +38,8 @@ RUN apt-get update \
 COPY --from=composer /usr/bin/composer /usr/bin/composer
 
 FROM phpdoc_base as prod
+
+FROM prod as dev-pcov
```
jaapio, line 42: Did you see we have a dev target? 
https://github.com/phpDocumentor/phpDocumentor/blob/b48c54995079a58888d9150ca534992582ca6710/Dockerfile#L32

mvriel: Yes, but that is included in prod already (build stages are built from top to bottom always) I thought: let's take it from prod. But I can also extend it from dev if you'd like.

It can't be included in the dev stage as it would slow down the build

jaapio: Maybe we should turn them around?  So dev is latest? Reason for me the set prod as latest and accept the delay during the build was to mitigate the possibility for a dev build in production by accident. Not sure that's a valid reason

mvriel: Switching them around makes sense; I'll do that

jaapio:
Thanks!

PR #3567, opened 2023-08-20
Feat: Add basic extension loading

jaapio:
Enable loading of phpdoc extensions. Extensions are based on symfony DI extensions. We do check the manifest.xml for compatibility to ensure we have everything needed.


jaapio:
This is a follow up PR of #3041. I removed all logic of symfony/framework-bundle that makes it less complex to load extensions as we do not have to take care of the container cache.

PR #3663, opened 2024-01-28
Fix wrong value for visibility option in XSD

sebastian-meyer:
The phpDocumentor command line helper says, that the parameter `visibility` allows the following values:

`api`, `public`, `protected`, `private`, `internal`.

But the configuration XSD enumerates the following values as valid options:

`api`, `public`, `protected`, `private`, `hidden`.

So `internal` is missing from the XSD and using `hidden` actually leads to an error.

This pull requests fixes this issue by replacing `hidden` with `internal` in the XSD.


sebastian-meyer:
What happened to this pull request? Since the `force-push` from the pull bot my changes are gone, but the master branch still has an invalid XSD... So although this is marked as merged, my changes clearly weren't merged into `master`.

PR #3666, opened 2024-02-06
Fix wrong value for visibility option in XSD

sebastian-meyer:
The phpDocumentor command line helper says, that the parameter `visibility` allows the following values:

`api`, `public`, `protected`, `private`, `internal`.

But the configuration XSD enumerates the following values as valid options:

`api`, `public`, `protected`, `private`, `hidden`.

So `internal` is missing from the XSD and using `hidden` actually leads to an error.

This pull requests fixes this issue by replacing `hidden` with `internal` in the XSD.

(I already opened a pull request for this issue [#3663], but that was marked as "merged" although is wasn't.)

### DIFF
--- a/data/xsd/phpdoc.xsd
+++ b/data/xsd/phpdoc.xsd
@@ -28,7 +28,7 @@
             <xs:enumeration value="public"/>
             <xs:enumeration value="protected"/>
             <xs:enumeration value="private"/>
-            <xs:enumeration value="hidden"/>
+            <xs:enumeration value="internal"/>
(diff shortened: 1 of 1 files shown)


sebastian-meyer:
The failing QA workflow reports an issue with a part of code this pull request doesn't change. Am I supposed to fix this anyway with this PR?

jaapio:
It's ok, I will take care of psalm. If you want to fix it feel free to open a pr. That would be very helpful. 

Thanks for addressing this!

PR #3671, opened 2024-02-15
Small improvements to guide rendering

sebastian-meyer:
This fixes a few issues with rendering, especially for guides:

- The default left padding for list items is `var(--spacing-md)` which unfortunately looks awful, because it gives all list bullet points a small negative indentation compared to text paragraphs. I've therefore increased the left padding to `var(--spacing-lg)` which looks a lot better in my opinion.
- By default the guide section in the sidebar is named "documentation" (with a lowercase first letter) while all other sections have an uppercase first letter. To unify the look and feel I've added twig's [`title`](https://twig.symfony.com/doc/2.x/filters/title.html) filter to the `sidebar.html.twig` template. Now all section headers start with uppercase letters.
- I also unified the "implements", "extends" and "uses" statements for classes, traits and enums, because all of them were lowercase except for "Uses". I've now changed that to lowercase as well.
- All sidebar sections got an additional unique class assigned to them. In default configuration this does nothing, but it allows for example easy custom re-ordering of the sections by simple CSS (using `display: flex;` and `flex-direction: column;` on the sidebar itself and `order: X;` on each section).

There is one more major annoyance in the rendering of guides which I couldn't fix because I am not an expert with how RST rendering works:

- For the API documentation the header hierarchy for content starts with `h2`, but for guides it always starts with `h1`, which makes it very hard to have the same look and feel for both. I didn't find any way to tell RST to make the highest hierarchy heading a `h2` instead of `h1`.

### DIFF
--- a/data/templates/default/components/class-title.html.twig
+++ b/data/templates/default/components/class-title.html.twig
@@ -29,7 +29,7 @@
 
     {% if node.usedTraits is not empty %}
         <span class="phpdocumentor-element__extends">
-            Uses
+            uses
             {% for trait in node.usedTraits %}
                 {{ trait|route('class:short') }}{% if not loop.last %}, {% endif %}
             {% endfor %}
--- a/data/templates/default/components/enum-title.html.twig
+++ b/data/templates/default/components/enum-title.html.twig
@@ -29,7 +29,7 @@
 
     {% if node.usedTraits is not empty %}
         <span class="phpdocumentor-element__extends">
-            Uses
+            uses
             {% for trait in node.usedTraits %}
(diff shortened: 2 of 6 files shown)


sebastian-meyer:
A demo of the changes (with some more custom modifications) can be seen here:
https://opencultureconsulting.github.io/php-basics/guides/changelog.html
https://opencultureconsulting.github.io/php-basics/classes/OCC-Basics-DataStructures-StrictList.html

jaapio:
Thanks this is very nice stuff you did!

jaapio:
Regarding the titles, I need to have a look how to resolve this as the rst header counting starts by default at 1. This is configurable via the internals but its a bit hidden

PR #3680, opened 2024-03-07
Add CSS background colour for body

### LINKED ISSUE #3679: Background colour is not defined, but text colour is
## Expected behavior
Visiting rendered docs should be readable, also if the default colours of the browser were adjusted.

## Actual behavior
The css defines a text colour, but no background colour. That will break depending on default background colour defined within the browser.

1. Change default background colour of your browser to something dark
2. Visit rendered documentation with default theme

## Your environment
The actual website at https://docs.phpdoc.org/guide/getting-started/installing.html#installation and any rendered documentation with downloaded phar file.

image

I'll prepare a PR.

DanielSiepmann:
As text is defined, but background was not.
Most developers assume every browser has the same default colors, e.g. white for background.
But actually users can alter those colors, and only changing one color will end up in a hard to read result.

Therefore the background color is now explicitly set to the expected white defaults.

Resolves #3679

### DIFF
--- a/data/templates/default/css/base.css.twig
+++ b/data/templates/default/css/base.css.twig
@@ -13,6 +13,7 @@
 /* Base Styles
 -------------------------------------------------- */
 body {
+    background-color: #fff;
     color: var(--text-color);
     font-family: var(--font-primary);
     font-size: var(--text-md);


jaapio:
Thanks Daniel!

PR #3686, opened 2024-03-19
Checking if the template location was provided before parsing it

vcampitelli:
Checking if the `location` was provided before normalizing the path, in order to fix the error below.

```shell
$ phpdoc --template=default 
PHP Warning:  Undefined array key "location" in src/phpDocumentor/Configuration/PathNormalizingMiddleware.php on line 90
PHP Stack trace:
PHP   1. {main}() bin/phpdoc:0
PHP   2. Symfony\Component\Console\Application->run($input = *uninitialized*, $output = *uninitialized*) bin/phpdoc:14
```

### DIFF
--- a/src/phpDocumentor/Configuration/PathNormalizingMiddleware.php
+++ b/src/phpDocumentor/Configuration/PathNormalizingMiddleware.php
@@ -87,7 +87,7 @@ final class PathNormalizingMiddleware implements MiddlewareInterface
 
         /** @var array{name: string, location?: ?Path, parameters?: array<string, mixed>} $template */
         foreach ($configuration['phpdocumentor']['templates'] as $key => $template) {
-            if (!isset($template['location'])) {
+            if (isset($template['location']) === false) {
                 continue;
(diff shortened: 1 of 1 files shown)


src/phpDocumentor/Configuration/PathNormalizingMiddleware.php
```diff
@@ -87,6 +87,10 @@ private function makeDsnRelativeToConfig(Configuration $configuration, UriInterf
 
         /** @var array{name: string, location?: ?Path, parameters?: array<string, mixed>} $template */
         foreach ($configuration['phpdocumentor']['templates'] as $key => $template) {
+            if (!isset($template['location'])) {
```
jaapio, line 90: I have a personal preference to read the if statements more verbose. the `!` is easy to overlook, not specifically to in this case but in other cases it is. So please change this to:

```suggestion
            if (isset($template['location']) === false) {
```

Apart from that, thanks for solving this issue!

PR #3693, opened 2024-04-10
Update Twitter icon to X

fulldecent:
This fixes only the project homepage.

Separately we can fix the template for this project's https://docs.phpdoc.org/


jaapio:
Thanks for updating this.

PR #3695, opened 2024-04-10
Document how to use with GitHub Actions

### LINKED ISSUE #2242: Document how to use phpDocumentor as github action
In phpDocumentor's master branch a `action.yml` has been added. This opens new doors for projects to use phpDocumentor in their github workflows. The action is a currently a minimal setup. But no docs are available. 

We should add a page in our cookbook to explain how to use phpDocumentor in github actions.

fulldecent:
Fixes #2242

### DIFF
--- a/docs/guides/continuous-integration.rst
+++ b/docs/guides/continuous-integration.rst
@@ -10,7 +10,93 @@ continuous integration services.
 GitHub Actions
 ==============
 
-<to be written>
+Here is a working starting point to deploy phpDocs to your GitHub Pages for a project. You may also want
+to update this to handle multiple versions of your project instead of just the latest.
+
+.. code-block:: yaml
+
+    name: Generate phpDoc
+    
+    # Allow GITHUB_TOKEN to deploy to GitHub Pages
+    permissions:
+      contents: read
+      pages: write
+      id-token: write
+    
+    on:
+      workflow_dispatch: # Allow to manually deploy
+      push:
+        branches: ["main"]
+    
+    jobs:
+      build:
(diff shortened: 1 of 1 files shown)


docs/guides/continuous-integration.rst
```diff
@@ -10,7 +10,93 @@ continuous integration services.
...
+            uses: actions/cache@v4
+            with:
+              path: phpdoc-cache
+              key: ${{ runner.os }}-phpdocumentor-${{ github.sha }}
+              restore-keys: ${{ runner.os }}-phpdocumentor-
+    
+          # Notice: -d xdebug.mode=off is required due to a bug/workaround
+          # Issue: https://github.com/phpDocumentor/phpDocumentor/issues/3642#issuecomment-1912354577
+          - name: Build with phpDocumentor
+            run: php -d xdebug.mode=off ${{ github.workspace }}/tools/phpDocumentor run -vv -d source --target docs --cache-folder phpdoc-cache --template default
```
jaapio, line 75: phpDocumentor provides a github action from this repository: https://github.com/phpDocumentor/phpDocumentor/blob/master/action.yml

```
      - name: "Build"
        uses: "phpDocumentor/phpDocumentor@master"
        with:
          target: "build/docs"
 ```
 
 I think this is a better way of running since we can pack every dependency we need in a docker image. So people do not have to care about installing php correctly.

jaapio (changes requested):
Thanks for creating this part of the docs. It's really nice that you are adding fixes for the docs.  It would be nice if you could check the github action, I think this helps to make the workflow easier.

PR #3701, opened 2024-04-15
Inherit method documentation through multiple generations

### LINKED ISSUE #2717: Inherit docblock method from interface thru abstract class
```php
// Interface, abstract and final class described in the same file.

interface SomeInterface
{
    /**
     * Description.
     */
    public function myMethod(int $arg): void;
}

abstract class SomeAbstraction implements SomeInterface
{
}

class SomeClass extends SomeAbstraction
{
    /**
     * {@inheritDoc}
     */
    public function myMethod(int $arg): void
    {
    }
}
```

## Expected behavior
SomeClass::myMethod() has same description as SomeInterface::myMethod().

## Actual behavior
SomeClass::myMethod() has description "{@inheritDoc}".

## Environment
* PHPDoc version used: 3.0.0
* Install method: phar
* PHP 7.4.10
* Windows 10

smaddock:
Fixes #2717

### DIFF
--- a/src/phpDocumentor/Descriptor/MethodDescriptor.php
+++ b/src/phpDocumentor/Descriptor/MethodDescriptor.php
@@ -210,47 +210,23 @@ class MethodDescriptor extends DescriptorAbstract implements
             return $this->inheritedElement;
         }
 
-        /** @var ClassInterface|InterfaceInterface|null $associatedClass */
-        $associatedClass = $this->getParent();
-        if (! $associatedClass instanceof ClassInterface && ! $associatedClass instanceof InterfaceInterface) {
-            return null;
-        }
-
-        $parentClass = $associatedClass->getParent();
-        if ($parentClass instanceof ClassInterface || $parentClass instanceof Collection) {
-            // the parent of a class is always a class, but the parent of an interface is a collection of interfaces.
-            $parents = $parentClass instanceof ClassInterface ? [$parentClass] :
-                $parentClass->filter(InterfaceInterface::class);
-            foreach ($parents as $parent) {
-                /** @var MethodInterface|null $parentMethod */
-                $parentMethod = $parent->getMethods()->fetch($this->getName());
-                if ($parentMethod instanceof self) {
-                    $this->inheritedElement = $parentMethod;
-
-                    return $this->inheritedElement;
-                }
+        /** @var ClassInterface|InterfaceInterface|null $methodParent */
+        $methodParent = $this->getParent();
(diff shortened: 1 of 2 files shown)


smaddock:
Just pasting here so it's easy to find.

`PHP Fatal error:  Uncaught Error: Call to undefined method phpDocumentor\Reflection\Fqsen::getMethods() in /home/runner/work/phpDocumentor/phpDocumentor/src/phpDocumentor/Descriptor/MethodDescriptor.php:265`

https://github.com/phpDocumentor/phpDocumentor/actions/runs/8739710047/job/23994028767?pr=3701#step:7:328

jaapio:
The error you get is because of the design of the descriptors. The workflow is as following. 

First we parse all files in a project one by one. If that's done we transform the output of the parsing to descriptors. The links between classes, namepaces ect. are not part of the descriptors at this moment. This allows us to cache the descriptors without to much complexity because the linked objects will have bi-directional relationships. 

The next step is to link all objects (replace the FQSEN objects) if this cannot be done (because the actual descriptor doesn't exist) the FQSEN remains part of the list. 

So when iterating over the interfaces, you need to check if the returned object is an instance of a `InterfaceDescriptorInterface`

smaddock:
Oh, the error is in my unit test. Fixing...

smaddock:
Okay, I think this will pass all the checks now, but it won't solve issue #2717 until issue #3706 is also fixed.

jaapio:
Thanks a lot for your work, I checked your code and it seems all good too me. 

I will wrap up my current work and investigate what is going on with the other issue regading the inheritance.

jaapio:
@smaddock do you have time to check the issues you were facing are now solved? And check if your work does now work as expected? 

If not, please leave a message, I will check this pr later this week if it needs more changes.

jaapio:
did you close this by accident?

smaddock:
Yes, was just trying to FF my branch... did not mean to close.

jaapio:
This is very nice, I do not see any issues in the code. All tests are passing to I expect no issues are introduced. Thanks a lot for your hard work. 

If you have some time left it would be very nice to get some extra cypress tests that cover the new usecase of inheritance. We can do that in a new PR, as setting up the cypress test might not be very easy. The setup isn't very reliable as I experienced yesterday myself. In theorie `make e2e-test` should do the trick. But for me it didn't directly out of the box. 

Please let me know if you can do that. If not, thanks for your efforts. I hope to see you around when you are facing another issue or would like to see an improvement.

PR #3705, opened 2024-04-23
snorkell_ai/config_2ac2f

suchat932:
master


jaapio:
Sorry, I'm not accepting this pr without any explanation about what it does and why we would need this.

PR #3732, opened 2024-06-24
Fix settings:list and --list-settings not working

Mai-Lapyst:
Currently, the `settings:list` command is not being loaded by the CLI application, which also impacts the `--list-settings` flag to be not useable. This PR fixes this.


jaapio:
Thanks for your contribution, I do appreciate the time you took to solve this issue.

Issue #3737, opened 2024-06-27
Document the `@final` annotation


nicolas-grekas:
In Symfony, we use this annotation quite extensively - and Doctrine does also.

There are two use cases :
- when used without any comment, this is done to express that the class is final, but this final is intentionally not enforced so that technical use cases can still extend the class (think lazy proxies, like EntityManager's class)
- when used with a comment it reads as `@final since Symfony 6.4`, and obviously this means "don't extend me, I'm final, but for BC you can still extend me until the next major"

It'd be great to document these practice outside of the Symfony ecosystem. I think this use of `@final` is addressing a generic problem.

jaapio:
@nicolas-grekas I added a proposal for the description of this tag. I would like to invite you to leave any comments in the pr if you want to improve the text to clarify somethings.

jaapio:
@ashnazg do you think we should add a pr for this on PSR-19?

PR #3739, opened 2024-06-27
[DOCS] add @final tag to reference

jaapio:
We do see the @final tag being used in may projects to make it more clear what the intention of the tag is this has been added to the tag reference.

The actual behavior will be implemented in phpdocumentor in a later stadium.

refs #3737

### DIFF
--- a/docs/references/phpdoc/tags/final.rst
+++ b/docs/references/phpdoc/tags/final.rst
@@ -1,0 +1,67 @@
+@final
+======
+
+The `@final` decorator is used to denote that the associated *Structural Element* is final,
+are not allowed to extend or override the *Structural Element* in a child element.
+
+This tag can only be used on the following *Structural Elements*:
+
+- ``method``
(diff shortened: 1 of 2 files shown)


docs/references/phpdoc/tags/final.rst
```diff
@@ -0,0 +1,67 @@
...
+*Structural Elements* that are marked as ``@final`` will be displayed
+as such in the generated documentation. The same way as the ``final`` keyword.
+
+Example
+-------
+
+.. code-block:: php
+
+    /**
+     * @final should not be extended by consuming code.
```
nicolas-grekas, line 46:
```suggestion
     * @final since version X.y.
```

nicolas-grekas (commented):
Thanks for the PR. Just one suggestion where the current example just duplicates the semantics of the tag while there are more interesting use cases (at least the one in the example :))

PR #3764, opened 2024-09-09
GH Actions: fix up action runner version

jrfnl:
Looks like dependabot doesn't take subdirectories of the `.github/workflows` directory into account when creating updates for the GH action runners.

Not sure how to fix that, but this PR at least fixes the outdated versions (before things break due to Node 16 no longer being supported).


jaapio:
Thanks!

PR #3793, opened 2024-10-10
feat: open rendered SVGs in separate tab/window onclick

dgoosens:
added a small bit of JS & CSS:  

- CSS to change the cursor to `zoom-in` on hover
- JS to open the SVG in a new window/tab onclick


jaapio:
Thanks a lot for this improvement!

PR #3840, opened 2024-12-19
trait.html.twig: Fix deprecated constant(s)

adriendupuis:
The only difference (ignore white spaces) between trait.html.twig and class.html.twig is this test.

To test `constants.deprecated` instead of `constant.deprecated` is highly suspicious.

`constants.deprecated` seems to be always `NULL`.

With `constant.deprecated` the CSS class `-deprecated` is applied when needed on "On this page" items.


jaapio (approved):
This was definitely a mistake on our side. Thanks for adding this fix!

PR #3851, opened 2025-01-03
Docs: fix link in generating-documentation.rst

ThatsJustCheesy:
It was linking to https://docs.phpdoc.org/guide/features/extensions/index.html#installing (which is irrelevant), instead of https://docs.phpdoc.org/guide/getting-started/installing.html#installation.


jaapio:
Thanks for this change!

Issue #3943, opened 2025-07-14
Clean template fails to show @uses if no description is provided in DocBlock


NabeelKhanYYC:
#When using the **clean**  template _@uses_ references are not displayed unless a description is provided.

**Steps to Replicate:**
1. Create function and add a DocBlock with the @uses tag like:
```
/**
 * Method to show bug with uses tag in phpDocumentor
 * 
 * @access public
 * @uses SomeNamespace\SomeClass::SomeMethod()
 * @return void
 */ 
```
2. Generate documentation using the clean template
3. Navigate to the method in generated documentation and note in the sidebar an empty uses
4. Update the DocBlock as follows
```
/**
 * Method to show bug with uses tag in phpDocumentor
 * 
 * @access public
 * @uses SomeNamespace\SomeClass::SomeMethod() Now you can see me!
 * @return void
 */ 
```
5. Generate documentation using the clean template.
6. Navigate to the method in generated documentation and note in the sidebar the uses is populated with "_Now you can see me!_" and clicking will navigate to **SomeNamespace\SomeClass::SomeMethod()**
7. Update the DocBlock to the one in step 1 above.
8. Generate documentation using the default template
9. Navigate to the method in generated documentation and note proper reference of @uses

**Suggested Fix:**
It appears that the issue is due to the template using the @uses description rather than the tag itself when rendering. Fixing the template to use the tag itself should resolve the issue.

I anticipate the code site in question is here: https://github.com/phpDocumentor/phpDocumentor/blob/7b7ae4320462d5d60ef9762820a7ac2c71e95170/data/templates/clean/parts/elements/property.html.twig#L42-L49

Version: ```phpDocumentor v3.8.0```

**Search Tags:**
* phpdocumentor @uses 
* phpdocumentor uses not working
* phpdocumentor uses not rendering
* phpdocumentor uses display bug
* phpdocumentor uses template bug

jaapio:
I think you are right, also the fix seems to be the correct way. Would you be able to create a pr with a proposal to solve this?

NabeelKhanYYC:
Yes, I'll do my best to get one done shortly @jaapio.

## Code Sites
Upon investigating this further it looks like I was incorrect on the code sites.

The side bar is actually controlled by: https://github.com/phpDocumentor/phpDocumentor/blob/7b7ae4320462d5d60ef9762820a7ac2c71e95170/data/templates/clean/parts/elements/method.html.twig#L93-L100

[...]

PR #3944, opened 2025-07-17
Bugfix #3943 - Clean template fails to show @uses if no description is provided in DocBlock

NabeelKhanYYC:
See: https://github.com/phpDocumentor/phpDocumentor/issues/3943#issuecomment-3082246102

### DIFF
--- a/data/templates/clean/class.html.twig
+++ b/data/templates/clean/class.html.twig
@@ -305,9 +305,10 @@
 								{% endif %}
 								{% for tag in tags %}
 									<dd>
-										<a href="{{ tag.reference|route('url') }}">
-											<span class="namespace-wrapper">{{ tag.description ?: tag.reference }}</span>
-										</a>
+										<span class="namespace-wrapper">
+											<a href="{{ tag.reference|route('url') }}">{{ tag.reference }}</a> 
+											{{ tag.description }}
+										</span>
 									</dd>
 								{% endfor %}
 							{% endif %}
--- a/data/templates/clean/parts/elements/constant.html.twig
+++ b/data/templates/clean/parts/elements/constant.html.twig
@@ -43,7 +43,12 @@
                 {% if uses_tags is not empty %}
                     <dt>Uses</dt>
                     {% for tag in uses_tags %}
-                        <dd><a href="{{ tag.reference|route('url') }}"><span class="namespace-wrapper">{{ tag.description ?: tag.reference }}</span></a></dd>
+                        <dd>
+                            <span class="namespace-wrapper">
+                                <a href="{{ tag.reference|route('url') }}">{{ tag.reference }}</a> 
+                                {{ tag.description }}
+                            </span>
+                        </dd>
                     {% endfor %}
                 {% endif %}
             </dl>
--- a/data/templates/clean/parts/elements/method.html.twig
+++ b/data/templates/clean/parts/elements/method.html.twig
@@ -95,7 +95,12 @@
                 {% if uses_tags is not empty %}
                     <dt>Uses</dt>
                     {% for tag in uses_tags %}
-                        <dd><a href="{{ tag.reference|route('url') }}"><span class="namespace-wrapper">{{ tag.description ?: tag.reference }}</span></a></dd>
+                        <dd>
+                            <span class="namespace-wrapper">
+                                <a href="{{ tag.reference|route('url') }}">{{ tag.reference }}</a> 
+                                {{ tag.description }}
+                            </span>
+                        </dd>
                     {% endfor %}
                 {% endif %}
             </dl>
--- a/data/templates/clean/parts/elements/property.html.twig
+++ b/data/templates/clean/parts/elements/property.html.twig
@@ -44,7 +44,12 @@
                 {% if uses_tags is not empty %}
                     <dt>Uses</dt>
                     {% for tag in uses_tags %}
-                        <dd><a href="{{ tag.reference|route('url') }}"><span class="namespace-wrapper">{{ tag.description ?: tag.reference }}</span></a></dd>
+                        <dd>
+                            <span class="namespace-wrapper">
+                                <a href="{{ tag.reference|route('url') }}">{{ tag.reference }}</a> 
+                                {{ tag.description }}
+                            </span>
+                        </dd>
                     {% endfor %}
                 {% endif %}
             </dl>


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM php:8.1 as base
+FROM php:8.1 AS base
```
NabeelKhanYYC, line 1: To resolve errors: `FromAsCasing: 'as' and 'FROM' keywords' casing do not match (line 37)`

Dockerfile
```diff
@@ -31,15 +34,25 @@ ENTRYPOINT ["/opt/phpdoc/bin/phpdoc"]
 
 COPY --from=composer /usr/bin/composer /usr/bin/composer
 
-FROM phpdoc_base as prod
+FROM phpdoc_base AS prod
```
NabeelKhanYYC, line 37: To resolve errors: `FromAsCasing: 'as' and 'FROM' keywords' casing do not match (line 37)`

Dockerfile
```diff
@@ -12,11 +12,14 @@ RUN apt-get -yq install openjdk-17-jre-headless \
 
 FROM composer:2 AS build
 
-COPY . /opt/phpdoc
+# This should speed up the build process faster when developing
+RUN mkdir -p /opt/phpdoc
```
NabeelKhanYYC, line 16: Because of the change on the next line this folder doesn't exist

Dockerfile
```diff
@@ -12,11 +12,14 @@ RUN apt-get -yq install openjdk-17-jre-headless \
...
+COPY ./composer* /opt/phpdoc
```
NabeelKhanYYC, line 17: Since we're installing composer dependencies at this time we only need the composer files not the source

Dockerfile
```diff
@@ -12,11 +12,14 @@ RUN apt-get -yq install openjdk-17-jre-headless \
...
+
 WORKDIR /opt/phpdoc
 RUN /usr/bin/composer install --prefer-dist -o --no-interaction --no-dev
 
-FROM base as phpdoc_base
+FROM base AS phpdoc_base
```
NabeelKhanYYC, line 22: To resolve errors: `FromAsCasing: 'as' and 'FROM' keywords' casing do not match (line 19)`

Dockerfile
```diff
@@ -31,15 +34,25 @@ ENTRYPOINT ["/opt/phpdoc/bin/phpdoc"]
...
 COPY --from=composer /usr/bin/composer /usr/bin/composer
 
-FROM phpdoc_base as prod
+FROM phpdoc_base AS prod
 
-FROM prod as dev
+# We parent dev from phpdoc_base so that code updates don't triger a full rebuild (also prod is blank)
+# 
+# This will need to be changed in the future if/when more commands are added but for now this vastyl speeds up the devleopment process
+FROM phpdoc_base AS dev
```
NabeelKhanYYC, line 42: To resolve errors: `FromAsCasing: 'as' and 'FROM' keywords' casing do not match (line 42)`

Also we want to minimize unnecessary composer installs and apt installs when the code is changing and we are doing development, this will need to change in the future but for now lets us build much faster per iteration

Dockerfile
```diff
@@ -31,15 +34,25 @@ ENTRYPOINT ["/opt/phpdoc/bin/phpdoc"]
...
 
 RUN apt-get update \
     && apt-get install -yq git \
     && useradd -m -s /bin/bash phpdoc
 
-FROM dev as dev-pcov
+# Move the full copy last so modifications don't triger a re-install
+COPY . /opt/phpdoc
```
NabeelKhanYYC, line 49: Copy the code over LAST so that we can use the docker cache as much as possible for faster builds

Dockerfile
```diff
@@ -31,15 +34,25 @@ ENTRYPOINT ["/opt/phpdoc/bin/phpdoc"]
...
+    
+# We do this for the same reasons as for dev (above)
+FROM phpdoc_base AS dev-pcov
 
 RUN pecl install pcov \
 	&& docker-php-ext-enable pcov
+
+# Move the full copy last so modifications don't triger a re-install
+COPY . /opt/phpdoc
```
NabeelKhanYYC, line 58: Copy the code over LAST so that we can use the docker cache as much as possible for faster builds

data/templates/clean/class.html.twig
NabeelKhanYYC, line 1: These changes are untested.

data/templates/clean/parts/elements/constant.html.twig
NabeelKhanYYC, line 1: These changes are untested.

data/templates/clean/parts/elements/method.html.twig
```diff
@@ -95,7 +95,12 @@
                 {% if uses_tags is not empty %}
                     <dt>Uses</dt>
                     {% for tag in uses_tags %}
-                        <dd><a href="{{ tag.reference|route('url') }}"><span class="namespace-wrapper">{{ tag.description ?: tag.reference }}</span></a></dd>
+                        <dd>
+                            <span class="namespace-wrapper">
+                                <a href="{{ tag.reference|route('url') }}">{{ tag.reference }}</a>
```
NabeelKhanYYC, line 100: Changed the layout since it's logical for the reference to be clickable and the comment to not be.

data/templates/clean/parts/elements/property.html.twig
NabeelKhanYYC, line 1: These changes are untested.

docker-compose.yml
```diff
@@ -1,18 +1,22 @@
-version: '3.8'
```
NabeelKhanYYC, line 1: To fix `the attribute `version` is obsolete, it will be ignored, please remove it to avoid potential confusion` warning message

docker-compose.yml
```diff
@@ -1,18 +1,22 @@
...
     build:
       context: .
       target: dev
       dockerfile: Dockerfile
-    user: ${CURRENT_UID}
-    volumes: [".:/opt/phpdoc"]
+    volumes:
+      - ./:/opt/phpdoc
+      # Use the container vendor folder not the one from ./
+      - /opt/phpdoc/vendor
```
NabeelKhanYYC, line 10: Without this we are mapping the current directory to the container and so the vendor folder is missed, with this line docker now knows to ignore that container when mapping and use the existing folder contents from our early build stage.

docker-compose.yml
```diff
@@ -1,18 +1,22 @@
...
     working_dir: "/opt/phpdoc"
+    entrypoint: "/opt/phpdoc/bin/phpdoc --template clean --force"
```
NabeelKhanYYC, line 12: This should not be here, will be fixed in a future commit.

docker-compose.yml
```diff
@@ -1,18 +1,22 @@
-version: '3.8'
 services:
   phpdoc:
     build:
       context: .
       target: dev
       dockerfile: Dockerfile
-    user: ${CURRENT_UID}
```
NabeelKhanYYC, line 8: Not used anywhere.

docker-compose.yml
NabeelKhanYYC: This is now fixed in [b262ab5ca](https://github.com/NabeelKhanYYC/phpDocumentor/commit/b262ab5cac3129196acfad34e7d029a2d620b6f2)

Dockerfile
jaapio: This could be tricky, there are a number of composer plugins active that need more files. So changing this could lead to unexpected behavior in the future during development.

docker-compose.yml
jaapio: This is being used in the makefile. Which I use for development. so please keep this

Dockerfile
jaapio: Not sure we have this, but the vendor directory should be in the dockerignore. Otherwise a possible local vendor dir would be built in.

jaapio (commented):
Minor comment

jaapio (approved):

Issue #3228, opened 2022-05-09
XML Template missing enum


Wulfheart:
Hello,

the XML template does not output any enums in a project. I am using v3.3.1. Would you accept such a PR? If so, can you offer any guidance on how to test this properly?

jaapio:
Hi

Thanks for reaching out to us. Yes I would love to get a pr to fix this issue. 

You can tests this by creating an directory called `.phpdoc/template` in your project and modify the sources that you can find here: 

https://github.com/phpDocumentor/phpDocumentor/tree/master/data/templates/xml

Each file can be overwritten separately. Allowing you to test with phpDocumentor versions without changing the way you run phpDocumentor. 

If you need more help feel free to ping us at any time.

NabeelKhanYYC:
@jaapio I re-discovered this bug today and will be issuing a PR shortly. Please keep in mind I'm still new to this codebase so the implementation might be incomplete. When reviewing the PR if you see anything missing please let me know.

PR #3947, opened 2025-07-20
Bugfix #3228 - Added XML template support for Enums

NabeelKhanYYC:
Fixed bug #3228 - Added XML Template support for Enums


jaapio (approved):
Thanks a lot for helping out!

PR #3954, opened 2025-07-28
GH Actions: remove GH Token set via `env`

jrfnl:
As of the release of `shivammathur/setup-php` v`2.35.0`, this should no longer be needed as the `setup-php` action runner will automatically use the `secrets.GITHUB_TOKEN` token.

Ref:
* https://github.com/shivammathur/setup-php/releases/tag/2.35.0
* https://github.com/shivammathur/setup-php/commit/55463ffe4fef24b158d1d7f861dce8b7d3811c13


jaapio:
Thanks for the update

Issue #3960, opened 2025-08-14
Build documentation github action is failing


jaymeh:
When trying to run the build documentation Docker container for my project, the action is failing.

## Expected behavior
I'd expect that the action would install dependencies and run 

## Actual behavior
The command is failing with the attached error messages:

```
  #13 [base 2/3] RUN mkdir -p /usr/share/man/man1     && apt-get update     && apt-get install --no-install-recommends -yq libicu-dev libicu72 zlib1g-dev ca-certificates-java gpg
  #13 0.166 Hit:1 http://deb.debian.org/debian trixie InRelease
  #13 0.167 Get:2 http://deb.debian.org/debian trixie-updates InRelease [47.1 kB]
  #13 0.168 Get:3 http://deb.debian.org/debian-security trixie-security InRelease [43.4 kB]
  #13 0.184 Get:4 http://deb.debian.org/debian trixie/main amd64 Packages [9668 kB]
  #13 0.241 Get:5 http://deb.debian.org/debian trixie-updates/main amd64 Packages [2432 B]
  #13 0.241 Get:6 http://deb.debian.org/debian-security trixie-security/main amd64 Packages [8780 B]
  #13 0.873 Fetched 9769 kB in 1s (13.5 MB/s)
  #13 0.873 Reading package lists...
  #13 1.366 Reading package lists...
  #13 1.841 Building dependency tree...
  #13 1.964 Reading state information...
  #13 1.979 E: Unable to locate package libicu72
  #13 ERROR: process "/bin/sh -c mkdir -p /usr/share/man/man1     && apt-get update     && apt-get install --no-install-recommends -yq libicu-dev libicu72 zlib1g-dev ca-certificates-java gpg" did not complete successfully: exit code: 100
  ------
   > [base 2/3] RUN mkdir -p /usr/share/man/man1     && apt-get update     && apt-get install --no-install-recommends -yq libicu-dev libicu72 zlib1g-dev ca-certificates-java gpg:
  0.166 Hit:1 http://deb.debian.org/debian trixie InRelease
  0.167 Get:2 http://deb.debian.org/debian trixie-updates InRelease [47.1 kB]
  0.168 Get:3 http://deb.debian.org/debian-security trixie-security InRelease [43.4 kB]
  0.184 Get:4 http://deb.debian.org/debian trixie/main amd64 Packages [9668 kB]
[...]

PR #3961, opened 2025-08-15
Restrict dependency of php:8.1 to 8.1-bookworm. This is due to missin…

jaymeh:
…g packages in Trixie causing builds to fail.

#3960


jaapio:
Thanks for helping out!

PR #3996, opened 2025-10-30
Fix the reference doc about visibilities

stof:
3 changes here:
- fixing the sentence for protected and private visibilities (instead of copying the sentence of the public visibility)
- using literals to show the actual values to use
- using a list to make the output more readable (the renderer does not preserve the line breaks of the source code of the existing source, making the rendered doc a lot less readable than the source)


jaapio (approved):

jaapio:
Thanks for these improvements

PR #4085, opened 2026-04-09
Add support for PHP 8.3 typed class constants

### LINKED ISSUE #3721: Add support for typed constants
# Feature request

Add support for native typed php constants.

lacatoire:
## Summary

Wire the native type from PHP 8.3 typed class constants into the `ConstantDescriptor`, so it is available for rendering in templates.

- Pass the type from the reflection `Constant` to the descriptor in `ConstantAssembler`, following the same pattern used by `PropertyAssembler`
- The HTML template already displays `node.type` for constants, so typed constants will render their native type automatically

When a native type is present, it takes precedence over `@var` docblock annotations (consistent with how PHP itself treats type declarations).

Depends on phpDocumentor/Reflection#736.

Fixes #3721

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -62,7 +62,7 @@
         "phpdocumentor/guides-markdown": "^1.7",
         "phpdocumentor/guides-restructured-text": "^1.7",
         "phpdocumentor/json-path": "^0.2.1",
-        "phpdocumentor/reflection": "^6.4.4",
+        "phpdocumentor/reflection": "^6.5.0",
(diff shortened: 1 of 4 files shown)


jaapio:
Thanks, I released phpdocumentor/reflection 6.5.0 you can update this version here to make the pipeline  pass.

jaapio (approved):

jaapio:
Very nice, thanks for your help. I will try to prepare a new release soon.

PR #4087, opened 2026-04-10
Add support for PHP 8.4 interface properties

lacatoire:
## Summary

PHP 8.4 allows declaring properties in interfaces using property hooks:

```php
interface HasId
{
    public int $id { get; }
}
```

This was not supported, causing a crash: `Expected an instance of any of Class_, Trait_. Got: Interface_`.

- Accept `InterfaceInterface` in `PropertyDescriptor::setParent()`
- Add property collection to `InterfaceInterface` / `InterfaceDescriptor` with inheritance support
- Assemble interface properties in `InterfaceAssembler`
- Display properties in the default interface template

Fixes #4061


src/phpDocumentor/Descriptor/InterfaceDescriptor.php
```diff
@@ -87,6 +92,34 @@ public function getInheritedConstants(): Collection
         return $inheritedConstants;
     }
 
+    public function setProperties(Collection $properties): void
```
jaapio: There is a trait we can use for this, HasProperties.
This reduces the code needed in this class.  You can have a look at the ClassDescriptor how to use that.

jaapio (changes requested):

jaapio:
I merged your changes in reflection a new versio (6.6.0) contains the fixes.

jaapio:
Thanks for your help again!

PR #4088, opened 2026-04-10
Adopt reflection 6.6.0 for source line info in parse errors

### LINKED ISSUE #3537: No line number shown when having a parse error
This is a duplicate of the following issue with the same title: https://github.com/phpDocumentor/phpDocumentor/issues/3531

I am re-opening this because the issue was not solved in general, only for the individual poster.

Please consider adding the line number OR surrounding code so there is a way to find where the error happened.

If there are any other ways to find the line number, such as with a differing lint command, please share.

Regards.

lacatoire:
Fixes #3537. Follow-up to @jaapio's review — the line/file context now lives in `phpDocumentor/Reflection 6.6.0` (via `NodesFactory`), so this PR is reduced to:

- bump `phpdocumentor/reflection` to `^6.6.0`
- drop the middleware's walking logic (superseded by the wrapped exception message)
- drop the NOTICE log that used to point at phpDocumentor internals rather than the offending source file

Thanks a lot for the guidance on the right place to host this fix!

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -62,7 +62,7 @@
         "phpdocumentor/guides-markdown": "^1.7",
         "phpdocumentor/guides-restructured-text": "^1.7",
         "phpdocumentor/json-path": "^0.2.1",
-        "phpdocumentor/reflection": "^6.5.0",
+        "phpdocumentor/reflection": "^6.6.0",
         "phpdocumentor/reflection-common": "^2.0",
         "phpdocumentor/reflection-docblock": "^5.4",
         "phpdocumentor/type-resolver": "^1.8",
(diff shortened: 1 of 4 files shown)


jaapio:
While this fix works I think it's better to keep this out of this project. If we could change phpDocumentor/reflection to give a better error that contains the file and line number we keep the parser more isolated in that library. It would also serve other consumers of the project better. 

We could add this logic in the NodesFactory in reflection. That's the location doing the php parsing.

jaapio:
I merged your changes in reflection a new versio (6.6.0) contains the fixes.

jaapio:
Thanks this is a very nice improvement

PR #4093, opened 2026-04-12
Group repeated tag entries in the default template

### LINKED ISSUE #2230: Show `@since` statements as a changelog.
In the `default` template, we could best represent one or more `@since` as a changelog. Especially if these are on a file of class DocBlock.

lacatoire:
In the default template, multiple occurrences of the same tag (for instance two `@since` entries or two `@link` entries on the same element) were rendered as a staircase of duplicated `<dt>name</dt><dd>value</dd>` pairs. jrfnl's mockup in issue #2230 shows a cleaner grouping under a single label per tag name; that also matches the canonical `<dl>` pattern for term-with-multiple-definitions.

- One `<dt>` per tag name, one `<dd>` per occurrence
- Pre-filter empty tag series so upstream compiler passes that keep a placeholder empty entry no longer leak an orphan `<dt>` (the previous `tags|first|length > 0` guard also hid the whole section whenever the first series was empty; that is now consistent for every series)
- Fold the `{% if tag.link %}` / `{% if not tag.link %}` pair into a single `{% if … %}{% else %}{% endif %}` while the block was being touched anyway

Adds a cypress assertion on `Marios-Pizzeria.html` confirming the grouping for both `@link` and `@since` with the exact number of entries and the first occurrence contained in the next `<dd>` sibling.

Fixes #2230

### DIFF
--- a/cypress/integration/default/classes.spec.js
+++ b/cypress/integration/default/classes.spec.js
@@ -82,6 +82,25 @@ describe('Classes', function() {
                     .parent()
                     .should('have.attr', 'href', 'https://docs.phpdoc.org')
             });
+
+            it('Groups repeated tags under a single entry with one definition per occurrence', function () {
+                cy.get('.phpdocumentor-element.-class > .phpdocumentor-tag-list > .phpdocumentor-tag-list__entry')
+                    .as('entries')
+                    .should('have.length', 2);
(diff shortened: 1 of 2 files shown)


jaapio (approved):
Thanks this make sense. 👍

PR #4104, opened 2026-04-14
Clarify $this keyword usage in PHPDoc types

### LINKED ISSUE #1896: Outdated docs page for Anatomy of a Type
http://docs.phpdoc.org/guides/types.html lists `$this` as a valid Type keyword.

But that page links to the "definition of a Type" page (http://docs.phpdoc.org/references/phpdoc/types.html) which says that `$this` is not a valid type.

lacatoire:
Fixes #1896.

Removes the `$this` entry from the Supported Types guide (it is not a formal PHPDoc type) and moves the explanation to the `@return` tag reference, where `$this` is actually meaningful for fluent interfaces.

Direction agreed with @jaapio in the issue thread.

### DIFF
--- a/docs/guides/types.rst
+++ b/docs/guides/types.rst
@@ -124,8 +124,11 @@ static
     An object of the class where this value was consumed, if inherited it will represent the child class. (see late
     static binding in the PHP manual).
 
-$this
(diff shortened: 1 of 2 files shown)


jaapio (approved):
Thanks for this improvement

PR #4106, opened 2026-04-14
docs: clarify @package separator characters

### LINKED ISSUE #3668: Documentation issue: Separator characters and the @package tag
This bug report doesn't relate to the application; instead, the report relates to the documentation for the `@package` tag and more specifically the separator character used.

[Here](https://docs.phpdoc.org/3.0/guide/references/phpdoc/tags/package.html) the documentation notes the underscore character can be used as a separator in the same way as the backslash:

> Aside from the backslash (\), phpDocumentor also allows the underscore (_) and dot (.) as separators for compatibility with existing projects. Despite this the backslash is RECOMMENDED as separator.

To me this indicates `_` and `\` are equivalent.

[Here](https://manual.phpdoc.org/HTMLSmartyConverter/HandS/phpDocumentor/tutorial_tags.package.pkg.html) though, the documentation has an example which uses the underscore character in the package name:

```
/**
 * class bluh
 * class-level DocBlock example.
 * @package applies_to_bluh
 */
```

To me this indicates `_` isn't quite the same as `\` because in this case `_` isn't being used as a level "separator"; instead, `_` is part of the top level name.

lacatoire:
The current reference lists only `_` and `.` as alternative separators, and the legacy manual example (`applies_to_bluh`) suggests `_` can be part of a level name. In practice `PackageTreeBuilder::normalizePackageName` converts `.`, `_`, `-`, `[` and `]` into backslashes unconditionally, so `applies_to_bluh` is three levels.

Extend the note with the full list and an explicit example so the reference matches the implementation.

Fixes #3668

### DIFF
--- a/docs/references/phpdoc/tags/package.rst
+++ b/docs/references/phpdoc/tags/package.rst
@@ -53,9 +53,13 @@ organized in their own sidebar section.
 
 .. note::
 
-    Aside from the backslash (``\``), phpDocumentor also allows the
-    underscore (``_``) and dot (``.``) as separators for compatibility
-    with existing projects. Despite this the backslash is RECOMMENDED
+    Aside from the backslash (``\``), phpDocumentor also accepts the
+    underscore (``_``), dot (``.``), hyphen (``-``) and square brackets
+    (``[``, ``]``) as separators for compatibility with existing projects.
+    They are treated as equivalent to the backslash and cannot be part
+    of a level name: ``@package PSR_Documentation_API`` is interpreted
+    as the three levels ``PSR``, ``Documentation`` and ``API``, not as
+    a single top-level name. Despite this the backslash is RECOMMENDED
     as separator.
 
 Examples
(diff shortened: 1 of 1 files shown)


docs/references/phpdoc/tags/package.rst
```diff
@@ -53,10 +53,14 @@ organized in their own sidebar section.
...
 
-    Aside from the backslash (``\``), phpDocumentor also allows the
-    underscore (``_``) and dot (``.``) as separators for compatibility
-    with existing projects. Despite this the backslash is RECOMMENDED
-    as separator.
+    Aside from the backslash (``\``), phpDocumentor also accepts the
+    underscore (``_``), dot (``.``), hyphen (``-``) and square brackets
+    (``[``, ``]``) as separators for compatibility with existing projects.
+    They are treated as equivalent to the backslash and cannot be part
+    of a level name: ``@package applies_to_bluh`` is interpreted as the
```
jaapio: I would like to see the example below being used rather than something random. This gives it a more uniform style.

jaapio (approved):

PR #4108, opened 2026-04-15
Show constant type alongside value in TOC and truncate long values

### LINKED ISSUE #2443: 3.0.0: Suggestion: show constant type in TOC instead of value
**Given the following piece of code in file `src/issue-2443.php`:**
``` php
<?php

class Issue2443
{
    /**
     * Summary.
     *
     * @var string
     */
    const DOCBLOCK_LONG_STRING_VALUE = '
    <div>
        <p>This is a long string, spanning multiple lines.</p>
        <p>%1$s</p>
        <p>%2$s</p>
    </div>';

    /**
     * Summary.
     *
     * @var string
     */
    const DOCBLOCK_LONG_NOWDOC_VALUE = <<<'EOD'
This is a really long string, spanning multiple lines.

Lorem ipsum dolor sit amet, consectetur adipiscing elit. Mauris finibus sed sem vel pulvinar.
Maecenas sit amet sapien sed quam porttitor condimentum. Nulla ac nisi et lectus facilisis
sollicitudin. Suspendisse potenti. In ac sollicitudin urna, sed molestie augue. Nulla posuere
risus vitae lacinia tristique. Donec lacinia, ipsum non bibendum vehicula, neque mi sodales
lacus, at facilisis nisl ex a elit. Suspendisse egestas purus dui, sed egestas lorem aliquam
a. Fusce varius nulla vel ex egestas rhoncus. Quisque vulputate, diam quis iaculis rutrum,
quam magna lacinia urna, sit amet euismod lacus velit quis turpis.
EOD;

    /**
     * Summary.
     *
     * @var string[]
     */
    const DOCBLOCK_ARRAY_VALUE = [
        'a' => 'foo',
        'b' => 'bar',
        'c' => 'baz',
        'd' => 'fool',
        'e' => 'bal',
        'f' => 'food',
        'g' => 'bat',
        'h' => 'for',
        'i' => 'bah',
        'j' => 'foz',
        'k' => 'yoh!',
    ];
}
```

<details>
  <summary><b>with the following config:</b></summary>

```xml
<?xml version="1.0" encoding="UTF-8" ?>
<phpdocumentor
    configVersion="3"
    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
    xmlns="https://www.phpdoc.org"
    xsi:noNamespaceSchemaLocation="https://docs.phpdoc.org/latest/phpdoc.xsd"
>
    <title>Demo</title>
    <version number="latest">
        <api format="php">
            <source dsn=".">
                <path>src</path>
            </source>
        </api>
    </version>
    <template name="default"/>
</phpdocumentor>
```
</details>

## Expected behavior
I'd expect the TOC to show the type of the constants, not the value.
Same as is done for properties and method (return type).

Mind: when the type is unknown, `mixed` should be shown. See #2442.

## Actual behavior
The value is shown and is barely readable for long values as it suffers from the same problem as described in #2228.

issue-2443

1. Issue can be reproduced with the above code sample and config.
2. Run `php -f /path/to/phpDocumentor/bin/phpdoc`

## Your environment

* Version used: phpDocumentor v3.0.x-dev@d15f2d6
* Install method: git clone
* PHP version: 7.4.6
* Operating system and version: Windows 7

lacatoire:
Fixes #2443

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -86,6 +86,7 @@
         "symfony/stopwatch": "^6.0",
         "symfony/string": "^6.4",
         "symfony/yaml": "^6.0",
+        "twig/string-extra": "^3.17",
         "twig/twig": "^3.17",
         "webmozart/assert": "^1.3"
     },
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "258bebe7677041d8b6993320a152b7c3",
+    "content-hash": "d86c414752dc6e5fde60dfc5f09a4a56",
     "packages": [
         {
             "name": "dflydev/dot-access-data",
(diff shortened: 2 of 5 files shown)


jaapio (changes requested):
I think it's better to move the truncation of this value into a filter in the twig extension. This allows other templates to reuse the same logic. 

The current line is hard to interpret, a filter with the option to set the max length is easier to understand for users.

jaapio:
I didn't know this exists when I added my first comment. But we can use this twig filter: https://twig.symfony.com/doc/3.x/filters/u.html

jaapio (approved):

jaapio:
Thanks for adjusting this.

Issue #3016, opened 2021-09-22
Default template doesn't show the current section


speller:
I just tried PHPDocumentor and generated test documentation. And the issue is the same as in the official documentation page: it's impossible to quickly learn which section is currently displayed. For example https://docs.phpdoc.org/3.0/namespaces/phpdocumentor-configuration.html - there are FOUR pages called `Configuration` and if you get to one of these pages by a link you have no chance to understand what of these four pages the current one is except you start reading the URL or the main content to get the context and then compare to the TOC. It's extremely inconvenient.
Can it be solved by default?

Another question from a newbie - where are templates? The default green doesn't look fancy and it doesn't seem to be a good idea to develop my own from scratch.

mvriel:
> I just tried PHPDocumentor and generated test documentation. And the issue is the same as in the official documentation page: it's impossible to quickly learn which section is currently displayed. For example https://docs.phpdoc.org/3.0/namespaces/phpdocumentor-configuration.html - there are FOUR pages called Configuration and if you get to one of these pages by a link you have no chance to understand what of these four pages the current one is except you start reading the URL or the main content to get the context and then compare to the TOC. It's extremely inconvenient.
Can it be solved by default?

The headers of each page contains a breadcrumb, showing where you come from. But am I correct in assuming that what you are missing is that, for example, the currently active section is shown in bold in the sidebar? 

For example, like this:

![image](https://user-images.githubusercontent.com/193704/134301369-4727ff55-7016-4233-9e8e-ffdd78374696.png)

(this is a mock, not real)

> Another question from a newbie - where are templates? The default green doesn't look fancy and it doesn't seem to be a good idea to develop my own from scratch.

The easiest way is to create a folder `.phpdoc/template` in your project and add overrides for individual Twig template files. You can see an example here: https://github.com/phpDocumentor/phpDocumentor/tree/master/.phpdoc/template

If you were to add a file called `.phpdoc/template/css/custom.css.twig` you can even define CSS overrides to customize colors or you could even override the CSS variables that we use to customize colors and layout parameters without having to override the CSS itself. You can see a list of all CSS variables in use here: https://github.com/phpDocumentor/phpDocumentor/blob/master/data/templates/default/css/variables.css.twig

And yes, this should be documented properly (which is ironic, I know 😉). Thank you for asking this question and I hope that the answer helps you

speller:
I also tried the `clean` template but it produced broken styles and broken links....

[...]

PR #4109, opened 2026-04-15
Include element FQN in HTML page title

lacatoire:
Pages for classes, interfaces, traits, enums, namespaces, packages, files and the file index now set a descriptive `<title>`, so browser tabs, bookmarks and history can distinguish pages that share the same short name (for example several nested `Configuration` namespaces).

Refs #3016


jaapio (approved):

jaapio:
Thank you!

